PR #24, opened 2024-12-26
minor improvements

del000:
1. add webhook for remote image send
2. add config option "telegram_thread_id" - send to specified topic of super group


flyrouter (commented):
Thank you !

PR #25, opened 2024-12-27
Backup & restory functionality

del000:
backuper

what its do:
1. script auto create backuper conf, if its not exist
2. sampe config:
```
#=== OpenIPC webui backuper config,
#=== <- this is comment line
#=== FILES you need to backup, one per line, full path, line begins from #
#/etc/webui/
#/etc/majestic.yaml
#/etc/fstab
#/usr/sbin/motion.sh

#=== COMMANDS, one per liine, to be executed AFTER restoring from backup
cli -s .audio.enabled true
cli -s .audio.speakerPin 64
cli -s .audio.codec opus
cli -s .audio.srate 48000
```
it can create backup copies of files/directories from a list from conf and execute commands after recovery. The conf file is essentially a script, which is why the syntax is so strange.

3. backup creation works on webui and via url  `wget --content-disposition http://root:12345@192.168.70.32/cgi-bin/ext-backuper.cgi?backup=create`

4. currently, works only manual restore via ssh, until file upload is fixed by majestic devs

### DIFF
--- a/www/cgi-bin/ext-backuper.cgi
+++ b/www/cgi-bin/ext-backuper.cgi
@@ -1,0 +1,32 @@
+#!/usr/bin/haserl --upload-limit=200 --upload-dir=/tmp
+<%in p/common.cgi %>
+<%
+config_file=/etc/webui/backup.conf
+
+backup_create() { # backup_create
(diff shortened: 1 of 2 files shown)


flyrouter (approved):
Thank you !
Waiting for the next PR to disable the temporarily non-working button.

PR #36, opened 2025-04-24
convert config-utils.sh to haserl fpv-common.cgi

mikecarr:
The original script relied on config-utils.sh to be in /usr/lib/openipc, we decided to make this part of the cgi/p/fpv-common.cgi. We need both p/common.cgi and p/fpv_common.cgi.

Also need to chmod 755 www/cgi-bin/fpv-wfb.cgi


flyrouter:
Thank you

@cronyx Please check this PR & permission of file and add to package, thanks

mikecarr:
Thank you

Issue #44, opened 2025-07-06
Lost connection with GitHub afte firmware update via WEBUI .


PaddyP90:
After updating firmware via WEBUI and rebooting it looses connection to GitHub unless I manually sync time..

bebach8:
I can confirm that it looses connection to Github even if I make manual time sync and there is no check for firmware updates

PR #122, opened 2026-07-21
fw-update: check updates via sysupgrade, not a verifying HTTPS HEAD (#121)

### LINKED ISSUE #121: [BUG] WebUI firmware update shows "No access to GitHub" although GitHub access works
Hello,

Note: I used ChatGPT to help diagnose the issue and prepare this report. All tests and results were collected from my OpenIPC camera.

I encountered an issue while trying to update my OpenIPC camera firmware from the WebUI.

The WebUI reports:  "No access to GitHub"

However, the camera has full access to GitHub and OpenIPC servers.

**Camera information:**
SoC: gk7205v200
Model: gk7205v200_lite
OpenIPC System Updater: OpenIPC System Updater v1.0.53

**Firmware before update:**
Kernel: 01:51:48 2026-07-09
RootFS: master+0992750, 2026-07-08

I checked the network connectivity from the camera: 
`ping github.com`
works correctly.

HTTPS access also works:
`curl -I [https://openipc.github.io/](https://openipc.github.io/?utm_source=chatgpt.com)`
returns `HTTP/1.1 200 OK`

Access to the firmware manifest works: `curl -I https://openipc.github.io/firmware/manifest.flat`
returns `HTTP/1.1 200 OK`

The updater can also retrieve available builds: `sysupgrade --list-builds`

Output:
```
Available builds for gk7205v200_lite:
nightly-20260716-0992750
nightly-20260708-0992750
```

The update failed from the WebUI with the message "No access to GitHub".

However, the update succeeds manually from SSH:
`sysupgrade --build=nightly-20260716-0992750 --no_reboot`

The firmware was downloaded and flashed successfully:
```
Firmware:
Resolved nightly-20260716-0992750 for gk7205v200_lite
Kernel updated successfully.
RootFS updated successfully.
```

After reboot, the camera is running:
```
Kernel:
01:39:52 2026-07-17

RootFS:
master+0992750, 2026-07-16
```

It looks like the "No access to GitHub" message is misleading. GitHub/OpenIPC access is working correctly, and the sysupgrade tool can download and install the firmware.

Could you please check if the WebUI update check is reporting the wrong error message or failing at another step?

Thank you very much for all your work.

Piechmu

widgetii:
Closes #121

## Problem

`fw-update.cgi` shows **"— no access to GitHub —"** and disables the *Install update from GitHub* button, even when the camera has full GitHub access and `sysupgrade` updates fine from SSH. Reported on gk7205v200_lite / gk7205v210, reproduced on a hi3516ev300_lite.

## Root cause

The "Latest on GitHub" probe was a **separate, more fragile** code path than the actual updater:

- `github_ver()` did a **verifying** `curl -m5 -ILs …/releases/download/latest/openipc.<soc>-<flash>-<variant>.tgz | grep last-modified`.
- A fresh/rarely-rebooted flash boots with its **clock stuck at the build date** until NTP converges. Against that past clock GitHub's real TLS cert is "not yet valid", so curl fails instantly with `curl: (60) Cert verify failed: BADCERT_FUTURE` → empty result → "No access to GitHub" + disabled button. Same failure family as #44.
- The **real** updater (`fw-update.js` → `/ws/upgrade` → firmware `sysupgrade`) fetches OpenIPC's `manifest.flat` with `curl -k`, so it is immune to the stale clock — which is why `sysupgrade --list-builds` / `--build=…` work while the WebUI reports no access.

## Fix

Ask the same tool the updater uses, so the availability check and the updater share one source of truth:

```sh
latest_build() {
    [ -z "$network_gateway" ] && return
    command -v sysupgrade >/dev/null 2>&1 || return
    timeout 15 sysupgrade --list-builds 2>/dev/null \
        | grep -Eo '[A-Za-z0-9._]+-[0-9]{8}-[0-9a-f]+' | head -1
}
```

- Drops the bespoke `latest`-release-asset URL, the 5 s cap, the `ingenic` URL juggling, and the vestigial `fw_printenv -n upgrade` branch — `sysupgrade` already routes stock vs builder manifests by model.
- Keeps the `ver` (button gate) and `fw_date` (label) variables, so the rest of the page and `a/fw-update.js` are untouched. One file serves both the standard and FPV navs.
- "Latest on GitHub" now shows a clean `YYYY-MM-DD`.
- The `synctime` retry is no longer needed here (the check no longer depends on the clock); `synctime()` in `p/common.cgi` is left in place for a separate cleanup.

## Verification

Deployed to a hi3516ev300_lite whose clock was stuck at its build date and rendered over its real HTTP server:

| | "Latest on GitHub" | Install button |
|---|---|---|
| Before | `— no access to GitHub —` | disabled + no-access note |
| After | `2026-07-17` | **enabled** |

### DIFF
--- a/www/cgi-bin/fw-update.cgi
+++ b/www/cgi-bin/fw-update.cgi
@@ -3,25 +3,24 @@
 <%
 	page_title="Firmware Update"
 
-	github_ver() {
+	# Latest build available for THIS board, via sysupgrade — the same updater the
+	# Install button drives over /ws/upgrade. It reads OpenIPC's manifest.flat
+	# (firmware or builder repo, chosen by model) with `curl -k`, so unlike the old
+	# verifying HTTPS HEAD it is not defeated by a fresh flash's stale clock
+	# (issue #44/#121: BADCERT_FUTURE made curl fail while sysupgrade updated fine).
+	# `timeout` bounds it so a dead network cannot hang the page.
+	latest_build() {
 		[ -z "$network_gateway" ] && return
-		fw_soc=$soc
-		[ "$soc_vendor" = "ingenic" ] && fw_soc=$soc_family
-		builder=$(fw_printenv -n upgrade)
-		fw_url="${builder:-https://github.com/openipc/firmware/releases/download/latest/openipc.${fw_soc}-${flash_type}-${fw_variant}.tgz}"
-		curl -m5 -ILs "$fw_url" | grep -i last-modified | cut -d' ' -f2-
+		command -v sysupgrade >/dev/null 2>&1 || return
+		timeout 15 sysupgrade --list-builds 2>/dev/null \
+			| grep -Eo '[A-Za-z0-9._]+-[0-9]{8}-[0-9a-f]+' | head -1
 	}
 
-	ver=$(github_ver)
-	# Issue #44: a fresh flash boots with a stale clock so the HTTPS check
-	# fails; sync time once (NTP, else HTTP Date header) and retry.
-	if [ -z "$ver" ] && [ -n "$network_gateway" ]; then
-		synctime
-		ver=$(github_ver)
-	fi
-
+	ver=$(latest_build)
 	if [ -n "$ver" ]; then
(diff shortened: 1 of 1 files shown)


piechmu:
Thanks for the fix! Much appreciated.

PR #133, opened 2026-08-10
fw-update: confirm the version after the reboot, not that the camera answered

widgetii:
Fixes the part of #120 that survived the first pass. The reporter's 2026-07-29 build already had all three changes the issue was closed on — the `#` download bar in their screenshot proves firmware #2225 landed — so this is a remaining failure mode, not a regression.

### Reachability was never proof

`pollBack()` declared success as soon as the camera went unreachable and came back. That was meant to be evidence of a real flash (#123), reasoning that a failed upgrade leaves the camera up. But `sysupgrade`'s `die()` reboots on **every** failure path, so down-then-up is exactly what a failed upgrade looks like too — the page then redirected to `status.cgi` showing the version the user already had. That is the original **t31x** report verbatim: *"the page refreshes after several minutes, but the firmware is not updated"*.

Confirmed on hardware (lab hi3516av300, `busybox reboot` stubbed via `PATH`):

```
sysupgrade 1.0.56:  "File ... not found Aborting." -> "Unconditional reboot" -> REBOOTED
```

Now the version is compared instead. `fw-update.cgi` gains an `id` on the value it already renders; after the reboot the page is re-fetched from the camera, retrying while the CGI comes up, and falling back to "unknown" rather than "failed" when it genuinely cannot tell (a `--reset` run lands on the password page). `--force_ver` reflashes the same version on purpose, so it is exempt.

### The honest error message was dead code

`flashMarker` matched strings printed *before* the point of no return:

| marker | printed at | main flow |
|---|---|---|
| `Unmounting` | `check_sdcard` | line 685 |
| `Received and unpacked` | end of download | line 686 |
| `Protected: flashing` | — | line 693 ← actual point of no return |

`sawFlash` was therefore already true before the download started, so *"the camera never rebooted — the update may have failed"* could never fire; users always got *"Still flashing…"*. Now only markers at or after the first erase count, matched against a rolling window so one split across two frames is not missed.

### Two more

- **Report `Aborting.` immediately.** With OpenIPC/firmware#2249 a pre-flash failure no longer reboots, which leaves the socket open on a `tail -F` that will never emit another line. `pollBack` is only suppressed when nothing was flashed — a `die()` after the first erase still reboots and still needs following.
- **Stop the heartbeat during the upgrade.** The page kept polling `j/pulse.cgi` every 2 s through the download and flash; each tick spawns a shell that forks ~12 more processes and makes a loopback request to `/metrics/isp`, on a camera that is often single-core and busy unpacking firmware into tmpfs. The re-arm was also wrong — `.then(setTimeout(...))` invoked `setTimeout` immediately and passed `.then` a timer id, so ticks were scheduled 2 s apart regardless of how long the CGI took.

### Verification

Rendering checked on the lab hi3516av300 — `<dd id="fw-installed">2.6.08.09-lite</dd>`, exactly once. The reboot-reporting path needs sysupgrade ≥ 1.0.57 (OpenIPC/firmware#2249) to be exercised end to end; that has not been run against a real flash yet.

Related: OpenIPC/firmware#2249, and a majestic-side change guarding `/metrics` and `/cgi-bin` during upgrade mode.

### DIFF
--- a/www/a/fw-update.js
+++ b/www/a/fw-update.js
@@ -1,33 +1,81 @@
 // Firmware update over the robust /ws/upgrade WebSocket.
 // majestic stops video (frees RAM), streams download+verify here, then is
-// killed at the flash step; the page switches to "rebooting" and polls until
-// the camera returns. A local .tgz is first POSTed to /upload, then flashed via
-// sysupgrade --archive. Vanilla JS; `$` is the querySelector helper from main.js.
+// killed at the flash step; the page switches to "rebooting", polls until the
+// camera returns, and then checks the version it came back on — sysupgrade
+// reboots on failure too, so coming back is not by itself evidence of anything.
+// A local .tgz is first POSTed to /upload, then flashed via sysupgrade
+// --archive. Vanilla JS; `$` is the querySelector helper from main.js.
 (function () {
 	const out = $('#fw-output');
 	const ansi = /[][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g;
 	const dec = new TextDecoder('utf-8');
 
-	// sysupgrade prints one of these only after the (silent) download, once it is
-	// into the point-of-no-return flash. Seeing one lets us tell a real flash
-	// apart from an idle-timeout disconnect during the quiet download/time-sync.
-	const flashMarker = /Received and unpacked|Kernel updated|RootFS updated|Unmounting|Unconditional reboot|Protected: flashing/i;
+	// sysupgrade prints "Protected: flashing continues…" at its point of no
+	// return, immediately before the first erase, and the "… updated" lines once a
+	// partition has been written. Seeing one tells a real flash apart from a
+	// socket that dropped during the quiet download.
+	//
+	// Nothing printed EARLIER belongs here. "Unmounting SD card" comes from
+	// check_sdcard and "Received and unpacked" from the end of the download —
+	// both run before the flash, so listing them meant sawFlash was already true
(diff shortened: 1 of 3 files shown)


widgetii:
Findings 2 and 3 were correct and are fixed in `e87808b`.

**Heartbeat can stall.** Real, and introduced by this PR. Re-arming from `.finally()` fixed the pile-up but made liveness depend on the fetch settling, and it had no timeout — a request left hanging by a busy or half-rebooted camera would stop the heartbeat for the rest of the page's life. Bounded with an `AbortController`.

**Heartbeat never resumes.** Also real. `stopHeartbeat()` was one-way, so every outcome that leaves the user on the page with an idle camera froze the header widgets until a reload. Added `startHeartbeat()` and called it on all five: aborted before any write, rebooted but unchanged version, never rebooted, never came back, and an upload that failed before the upgrade started. Deliberately **not** on "Still flashing — give it a few minutes": there the camera is mid-write and not polling it is the entire point.

**Finding 1 (unescaped `fw_version`)** — not taken here. The `<%= "${fw_version}-${fw_variant}" %>` interpolation is pre-existing; this PR only adds an `id` to the surrounding `<dd>`. The values come from `/etc/os-release`, so injecting them requires root on the camera, and the same raw pattern is used for `$soc`, `$flash_type` and others on this page and elsewhere. It is a real deviation from the escaping rule in CLAUDE.md and worth a pass of its own — `ex`/`pre` both wrap output in block markup, so an inline escape helper would need adding first — but folding that into a fix for #120 would mix two unrelated changes.

Worth noting the new version check is unaffected either way: it reads `textContent` via `DOMParser`, and the before/after values are extracted the same way, so they compare equal regardless of what the raw string contains.

PR #136, opened 2026-08-12
status: estimate MemAvailable on kernels that don't report it

### LINKED ISSUE #116: WebUI incorrectly reports critical camera status on kernels < 3.14 due to missing MemAvailable
On kernels older than 3.14, `/proc/meminfo` [does not contain](https://lists.busybox.net/pipermail/busybox-cvs/2018-October/038415.html) the `MemAvailable` so `BusyBox` `free` shows 0 in the `available` column. As a result, the `WebUI` on `Ingenic T31x` reports a critical camera status on the main page.

I suggest a workaround: on kernels older than 3.14, use the `free` column instead of the `available` column from the `free` command output (or `MemFree` instead of `MemAvailable` from `/proc/meminfo`) on the main page.

Image

```bash
root@openipc-t31:~# cat /proc/meminfo | grep "MemAvailable"
root@openipc-t31:~# free
              total        used        free      shared  buff/cache   available
Mem:          59152       18568       20496           0       20088           0
-/+ buffers/cache:        18568       40584
Swap:             0           0           0
```

Compare to `SigmaStar SSC378DE`:
```bash
root@openipc-ssc378de:~# cat /proc/meminfo | grep "MemAvailable"
MemAvailable:      32552 kB
root@openipc-ssc378de:~# free
              total        used        free      shared  buff/cache   available
Mem:          53848       14300       25856           0       13692       32620
Swap:             0           0           0
```

widgetii:
Fixes #116.

## What's wrong

`/proc/meminfo` gained `MemAvailable` in kernel 3.14. The Ingenic T31 runs 3.10, and majestic's `/metrics` mirrors `/proc/meminfo` verbatim — exactly as node_exporter does, which omits the series rather than synthesising one — so `node_memory_MemAvailable_bytes` is **absent**, not zero.

`www/a/status.js` only assigns `m.memAvail` on an exact key match and initialised it to `0`, so `ram` computed to 100%, which trips `ram >= 97` and pins the badge to **"Critical"** forever. That's the screenshot in the issue.

`www/cgi-bin/j/pulse.cgi` already fell back to bare `MemFree` (added in c375cd6, #110). But measured on an idle T31 that reads **92%**, which trips the amber branch of `main.js:setProgressBar` — so the nav header bar has been showing a bogus warning as well. The workaround suggested in the issue therefore only trades a false Critical for a false Warning.

## Why not fix it in majestic

Synthesising the metric in the exporter would *break* the node_exporter compatibility that motivates the interface. Upstream omits it too — `collector/meminfo_linux.go` guards `if meminfo.MemAvailableBytes != nil`, and `prometheus/procfs` keeps it as a nil-able pointer with no fallback computation. The ecosystem convention is a consumer-side fallback via PromQL `or`; GitLab's node dashboards do exactly this, in a commit by a node_exporter maintainer. This patch is the JS/awk transliteration of that `or`. majestic already emits every input needed — no daemon change.

## The fallback

The shape of the kernel's own `si_mem_available()`:

```
MemFree + Active(file) + Inactive(file) + SReclaimable
```

Measured against a kernel that *does* report `MemAvailable`:

| candidate | error | reads on an idle T31 |
|---|---|---|
| `MemFree` (as suggested in the issue) | −98.3% | 89.9% used → false Warning |
| `MemFree + Buffers + Cached` | −22.9% | 39.9% used |
| **this patch** | **+1.7%** | **36.4% used** |

`wmark_low` lives in `/proc/zoneinfo` and isn't exposed via `/metrics` at all, so the watermark terms are dropped and the estimate runs ~2% optimistic.

## Result on the lab T31 (kernel 3.10.14, lite build)

| | badge | Memory | readout | bar |
|---|---|---|---|---|
| before | **Critical** | 100% | 32 / 32 MB | red |
| after | **All systems OK** | 36% | 11 / 32 MB | normal |

`pulse.cgi` goes 92% → 37%, matching busybox `free`'s own `-/+ buffers/cache` line (12120 of 33552 = 36.1%). Header bar and dashboard agree again — the parity goal of #110.

## Notes

- **Kernels that report `MemAvailable` are unaffected** — the metric is trusted whenever present, including a literal `0`. Hence `memAvail` initialising to `null`: absence has to be distinguishable from a genuine zero.
- `pulse.cgi` now does one `awk` pass instead of three forks; it runs every 2s.
- The `ram` computation is clamped — the estimate carries no watermark discount, so a large reclaimable slab could in principle push it past `memTotal`.
- Thresholds are untouched. They were only ever tripped by the bad input.

## Testing

- Real headless-browser render against the live camera, swapping only `status.js` between runs, for the before/after above. The "before" render reproduces the reporter's screenshot exactly.
- A node harness drives the real `status.js` through a DOM/`fetch` shim: 15 assertions over 5 fixtures (no MemAvailable / present / literal 0 / empty body / endpoint down). Run against the pre-patch file first to confirm it reproduces the bug. Passes on the patched file both before and after `terser --compress --mangle`.
- The awk was validated on the camera's busybox awk 1.36.1 (`Active\(file\)` needs the parens escaped — they're ERE grouping otherwise).

### DIFF
--- a/www/a/status.js
+++ b/www/a/status.js
@@ -7,7 +7,11 @@
 	function parseMetrics(text) {
 		const m = {
 			cpuIdle: 0, cpuTotal: 0, rx: 0, tx: 0,
-			memTotal: 0, memAvail: 0, temp: null, load1: null,
+			memTotal: 0, memAvail: null, temp: null, load1: null,
+			// Addends for the pre-3.14 MemAvailable estimate below. They init to
+			// 0, not null, so a field the kernel doesn't report contributes
+			// nothing instead of turning the whole sum into NaN.
+			memFree: 0, memActFile: 0, memInactFile: 0, memSReclaim: 0,
 			hls: 0, night: 0, ircut: 0, nodeTime: 0, nodeBoot: 0,
 		};
 		const lines = text.split('\n');
@@ -28,6 +32,10 @@
 					if (key.indexOf('device="lo"') < 0) m.tx += val;
 				} else if (key === 'node_memory_MemTotal_bytes') m.memTotal = val;
 				else if (key === 'node_memory_MemAvailable_bytes') m.memAvail = val;
+				else if (key === 'node_memory_MemFree_bytes') m.memFree = val;
+				else if (key === 'node_memory_Active_file_bytes') m.memActFile = val;
+				else if (key === 'node_memory_Inactive_file_bytes') m.memInactFile = val;
+				else if (key === 'node_memory_SReclaimable_bytes') m.memSReclaim = val;
 				else if (key === 'node_hwmon_temp_celsius') m.temp = val;
 				else if (key === 'node_load1') m.load1 = val;
 				else if (key === 'node_time_seconds') m.nodeTime = val;
@@ -39,6 +47,15 @@
 			} else if (key === 'hls_clients_total') m.hls = val;
 			else if (key === 'ircut_enabled') m.ircut = val;
 		}
+		// Kernels older than 3.14 omit MemAvailable, and /metrics mirrors
+		// /proc/meminfo verbatim (as node_exporter does), so the metric is absent
+		// rather than zero — leaving memAvail at 0 read as 100% used and pinned
+		// the badge to "Critical" (issue #116). Rebuild the kernel's own
+		// si_mem_available() estimate from the parts that are present. wmark_low
+		// is not in /proc/meminfo, so this runs ~2% optimistic. Keep in sync with
+		// the awk in cgi-bin/j/pulse.cgi.
+		if (m.memAvail === null)
+			m.memAvail = m.memFree + m.memActFile + m.memInactFile + m.memSReclaim;
 		return m;
 	}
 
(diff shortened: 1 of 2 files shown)


widgetii:
Thanks — the negative-MB finding was real, and it was a gap in my own reasoning: I anticipated the estimate overshooting `MemTotal` but clamped only the percentage, leaving `#st-ram-mb` to render a raw `memTotal - memAvail`.

Reproduced it with a fixture that forces the overshoot (an absurd `SReclaimable`), and the old build rendered exactly the described inconsistency:

```
FAIL  #st-ram-mb not negative
        got  "-940 / 32 MB"
        want "0 / 32 MB"
```

Fixed in 3797b38, though at the source rather than at the consumer as suggested: clamping `memAvail` itself makes `memAvail <= memTotal` an invariant every reader can rely on, keeps the percentage and the MB readout from ever disagreeing, and lets the percentage revert to its original unclamped expression — so the net diff is smaller than before.

Re-verified: 18/18 assertions on both the source and the `terser --compress --mangle` output, and the status page still renders "All systems OK" / 36% / "11 / 32 MB" on the lab T31.

PR #138, opened 2026-08-12
ci: syntax check the haserl templates and guard <%= escaping

widgetii:
Follows #137. Nothing validated the CGI templates: a syntax error in one only surfaces as a 500 on a camera, and the escaping rule CLAUDE.md lays down — never render `POST_`/`GET_` data with `<%=` — was enforced by review alone.

## Why haserl's own codegen, not a regex

Shell constructs legally span `<% %>` blocks in these files — `status.cgi` opens `if ...; then` in one block and closes it with `<% fi %>` several lines of HTML later (also `preview.cgi:29`, `mj-settings.cgi:36`, `ext-wireguard.cgi:39`). Anything linting blocks in isolation sees an unterminated `if` in every one of them.

`haserl -d` prints the script it *would* execute without running it, and resolves `<%in p/common.cgi %>` includes. That's the same idiom as ERB's `erb -x file.erb | ruby -c`, PHP's `php -l`, or Hadolint running ShellCheck over Dockerfile `RUN` blocks. Feed the output to `sh -n`.

## Two things found the hard way

**haserl and the shell fail independently, and both statuses matter.** A bad tag makes haserl exit non-zero while printing `Missing %> near line N` **on stdout**. So the obvious `haserl -d f | sh -n` throws that status away *and* hands the message to the shell, which finds it a perfectly valid command — a broken template gets reported clean. The codegen is captured before checking. My first attempt had exactly this bug.

**`<%in %>` resolves relative to the current working directory**, not to the including file. The manpage doesn't say so. Invoked from the repo root, every single page fails to find its include. The script runs from `www/cgi-bin`, which is the cwd majestic uses anyway.

## Mutation tested, not assumed

A gate that never fires is worse than none, so each fault class was injected and confirmed caught:

| injected fault | caught by |
|---|---|
| missing `<% fi %>` (spans blocks) | `sh -n` |
| unterminated `<%` tag | haserl exit status |
| bad `<%in %>` include path | haserl exit status |
| `<%= $POST_evil %>` | escaping grep |
| `<%= ${GET_foo} %>` | escaping grep |

30 templates check in ~0.8s. There are currently **zero** escaping violations across 111 `<%=` uses, so that half is purely a regression guard.

## Known limit

`sh -n` is syntax-only: `<% if [ ; then %>` passes, because `[` is just a command name. ShellCheck catches that class and could be layered on later — it'll need noise-tuning against generated code, so it felt like its own change.

## Follow-up

`sh -n` over the plain `#!/bin/sh` scripts (`sbin/*`, `j/*.cgi`) was originally left out of this PR. That was the wrong call — it left the device-side scripts unchecked, which is the half where a syntax error actually costs something. Added in #139, stacked on this branch, along with the one thing that was blocking it (`j/locale.cgi` is a data file wearing a script's shebang).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -34,3 +34,11 @@ jobs:
           node-version: '20'
       - run: npm ci
       - run: npm run build:dist
+
+  templates:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # universe; same 0.9.36 the cameras ship, just not the musl/mips build
+      - run: sudo apt-get update && sudo apt-get install -y haserl
+      - run: ./tools/lint-templates.sh
--- a/tools/lint-templates.sh
+++ b/tools/lint-templates.sh
@@ -14,9 +14,15 @@
 # WORKING DIRECTORY, not to the including file — undocumented in the manpage,
 # and the reason every page 404s its include if you invoke haserl from the repo
 # root. majestic execs the CGIs with that same cwd, so this matches production.
-cd "$(CDPATH= cd -- "$(dirname -- "$0")/../www/cgi-bin" && pwd)"
+#
+# Both of these are checked rather than assumed: this is a gate, and a gate that
+# cannot run has to say so instead of reaching the "templates ok" line. An
+# unguarded cd would silently lint the wrong directory (or none), and an
+# unguarded mktemp would drop every finding on the floor.
+DIR=$(CDPATH= cd -- "$(dirname -- "$0")/../www/cgi-bin" && pwd) || exit 1
+cd "$DIR" || exit 1
 
-FAILS=$(mktemp)
+FAILS=$(mktemp) || exit 1
 trap 'rm -f "$FAILS"' EXIT
 
 # --- 1. syntax --------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


widgetii:
Both findings were real and are fixed in 5787553 — thanks.

They were the same failure mode in three places: **the gate could report success without checking anything.** That's the exact thing I'd argued elsewhere in this PR makes a gate worse than useless, and I'd built three of them.

Demonstrated rather than assumed. Old version, run from a directory with no `www/cgi-bin`:

```
cd: .../www/cgi-bin: No such file or directory
cd: null directory
templates ok          <-- rc=0, having checked precisely nothing
```

Now each case exits non-zero naming the guard that failed:
- `grep` exits 0/1/2+ for match/clean/error and `if hits=$(grep ...)` conflated 1 with 2 — a broken grep looked identical to "no violations". Now a `case` on the status, with the >=2 branch recording a failure.
- `cd` and `mktemp` are checked.
- The GNU-only `grep -r --include` is gone in favour of a find-built file list passed to one grep — which is also what makes the exit code meaningful, so the portability point and the reliability point had the same fix.

Verified by injecting each failure (fake `grep`/`mktemp` on PATH returning non-zero, and running from a tree with no `www/cgi-bin`): all three now exit 1, the clean tree still passes, and the eight syntax/escaping mutations are still caught.

PR #139, opened 2026-08-12
ci: lint the plain shell scripts too, and stop the locale files lying

widgetii:
Stacked on #138 — it edits `tools/lint-templates.sh`, which #138 introduces. GitHub will retarget this to `master` once #138 merges.

This is the follow-up to the "deliberately not in this PR" section of #138. On reflection that exclusion was the wrong call: it left the *device-side* scripts unchecked, which is the half where a syntax error actually costs something.

```
sbin/setnetwork    writes /etc/network/interfaces.d
sbin/updatewebui   the deploy path
sbin/telegram, sbin/openwall
www/cgi-bin/j/*.cgi   12 JSON endpoints, incl. pulse.cgi
```

Nothing else tests any of them.

## The blocker, and why an exclusion list was the wrong fix

`j/locale.cgi` carries `#!/bin/sh` but is not a script — it's a data file parsed with `sed` by `mj-settings.cgi:13`, never sourced, never executed. `mj_cloud=Cloud (WebRTC)` isn't valid shell, so it fails `sh -n` at line 22.

Rather than special-case it, **drop the shebang** from it and `locale_fpv.cgi` so shebang-based selection skips them for what they are. Verified inert: the sed output is **byte-identical** before and after, since the shebang line never matched `^mj_` anyway.

Quoting the value instead would *not* work — the sed captures `\(.*\)` straight into JSON, so `"Cloud (WebRTC)"` would put literal quote characters inside the label.

## Correction to something I said in #138

I claimed dropping the shebang would stop `/cgi-bin/j/locale.cgi` returning 500. **It doesn't** — I tested it on the lab camera. The file ships `0755` (`build-dist.sh:29` forces that on every `*.cgi`) and majestic still execs it; it now fails with `ENOEXEC` instead of a shell syntax error. Same 500.

Nothing requests it, so it's cosmetic, but getting the file out of the CGI tree is a genuinely separate change and I haven't made it.

## Mutation tested

Both halves, since a gate that never fires is worse than none:

| injected fault | result |
|---|---|
| stray `fi` in `sbin/setnetwork` | caught |
| unterminated `case` in `j/pulse.cgi` | caught |
| unterminated `while` in `sbin/updatewebui` | caught |
| unbalanced quote in `sbin/telegram` | caught |
| `locale.cgi` re-acquiring a shebang | caught (correctly — then it *claims* to be a script) |
| all four haserl cases from #138 | still caught after the refactor |

15 shell scripts + 30 templates. Same `sh -n` limitation applies: `if [ -z "$x" ; then` passes, because `[` is just a command name — one of my first mutation attempts, and it was a bad test rather than a linter gap.

## Unrelated finding, not touched here

`locale_fpv.cgi` has **no reader anywhere in the tree** — the only references are CLAUDE.md:155 and structure.md:49. `mj-settings.cgi:13` hardcodes `j/locale.cgi` and isn't `fw_variant`-gated, so FPV builds appear to get the standard labels. Either the docs drifted or FPV lost its labels; worth a look independently of this PR.

### DIFF
--- a/tools/lint-templates.sh
+++ b/tools/lint-templates.sh
@@ -12,14 +12,17 @@
 # artifact rather than a haserl bug, but it does mean on-device linting is out.
 #
 # The cd and the mktemp below are checked rather than assumed: this is a gate,
-# and a gate that cannot run has to say so instead of reaching the "templates
-# ok" line. An unguarded cd would silently lint the wrong tree, or none at all,
+# and a gate that cannot run has to say so instead of reaching the success
+# line. An unguarded cd would silently lint the wrong tree, or none at all,
 # and an unguarded mktemp would drop every finding on the floor.
 ROOT=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd) || exit 1
 cd "$ROOT" || exit 1
 
 FAILS=$(mktemp) || exit 1
-trap 'rm -f "$FAILS"' EXIT
+# What actually got inspected, so the summary can report real counts. A gate
+# that checked nothing should not be indistinguishable from a gate that passed.
+SEEN=$(mktemp) || exit 1
+trap 'rm -f "$FAILS" "$SEEN"' EXIT
 
 # --- 1. syntax --------------------------------------------------------------
 # Shell constructs legally span <% %> blocks here: status.cgi opens `if ...;
@@ -48,6 +51,7 @@ if ! (
 	cd www/cgi-bin || exit 1
 	find . -name '*.cgi' | sed 's|^\./||' | sort | while IFS= read -r f; do
 		head -1 "$f" | grep -q haserl || continue
+		echo "t" >> "$SEEN"
 		if ! out=$(haserl -d "$f" < /dev/null 2>&1); then
 			printf 'www/cgi-bin/%s: haserl: %s\n' "$f" "$(printf '%s' "$out" | tail -1)" >> "$FAILS"
 			continue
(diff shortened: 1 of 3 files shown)


widgetii:
Thanks — one of the four is valid and is fixed in 98e07a9; the other three are stale.

**Valid: misleading success message.** The script printed `templates ok` after it started linting shell scripts too. It now reports what it actually covered:

```
ok: 30 haserl templates, 15 shell scripts
```

Counting the work also closes the last silent-pass route: if the `find` pattern or the shebang test ever stopped matching, the old code would have printed a confident success over zero files. Zero of either kind is now a failure naming the broken selector — verified against an empty tree.

**Stale: the `--include` / GNU-grep and "lint can silently skip" findings.** All three cite `606837f`, which was the branch tip ~90 seconds before I force-pushed the rebase (`062f7ca`). They were fixed in 5787553 on #138's branch, which this branch now sits on: `--include` and `-r` are gone, replaced by a find-built list into a single grep with a `case` on the exit status. Current tip is 98e07a9 if you want to re-run the review.

widgetii:
Superseded by #140 — same change rebased onto master, with the review fixes.

GitHub auto-closed this when #138 merged with `--delete-branch`: deleting a base branch closes the PRs targeting it, and it can't be reopened once the head has been rebased (I restored the base ref to try; `reopenPullRequest` still refuses after a force-push). My mistake for stacking onto a branch I then deleted on merge — the base should have been retargeted to master *before* merging #138.

PR #140, opened 2026-08-12
ci: lint the plain shell scripts too, and stop the locale files lying

widgetii:
Replaces #139, which GitHub auto-closed when #138 merged with `--delete-branch` (deleting a base branch closes the PRs targeting it, and it can't be reopened once the head has been rebased). Same change, rebased onto master, plus the review fixes. My mistake on the stacking — noted below.

Follow-up to #138. That PR deliberately skipped plain-shell linting because one file failed it; that was the wrong call, since the skipped half is the **device-side** half:

```
sbin/setnetwork    writes /etc/network/interfaces.d
sbin/updatewebui   the deploy path
sbin/telegram, sbin/openwall
www/cgi-bin/j/*.cgi   12 JSON endpoints, incl. pulse.cgi
```

Nothing else tests any of them.

## The blocker, and why an exclusion list was the wrong fix

`j/locale.cgi` carries `#!/bin/sh` but is not a script — it's a data file parsed with `sed` by `mj-settings.cgi:13`, never sourced, never executed. `mj_cloud=Cloud (WebRTC)` isn't valid shell, so it fails `sh -n` at line 22.

Rather than special-case it, **drop the shebang** from it and `locale_fpv.cgi`, so shebang-based selection skips them for what they are. Verified inert: sed label output is **byte-identical** before and after, since the shebang line never matched `^mj_`.

Quoting the value would *not* work — the sed captures `\(.*\)` straight into JSON, so `"Cloud (WebRTC)"` would put literal quotes inside the label.

## Review fixes carried in

The success line now reports what was actually covered, instead of claiming `templates ok` after also linting shell scripts:

```
ok: 30 haserl templates, 15 shell scripts
```

Counting the work also closes the last silent-pass route: if the `find` pattern or shebang test ever stopped matching, the old code printed a confident success over zero files. Zero of either kind now fails and names the broken selector.

## Correction to something I claimed in #138

I said dropping the shebang would stop `/cgi-bin/j/locale.cgi` returning 500. **It doesn't** — tested on the lab camera. The file ships `0755` (`build-dist.sh:29` forces that on every `*.cgi`) and majestic still execs it; it now fails `ENOEXEC` instead of a shell syntax error. Same 500. Nothing requests it, so it's cosmetic — but getting the data file out of the CGI tree is a separate change I have not made.

## Mutation tested

| injected fault | result |
|---|---|
| stray `fi` in `sbin/setnetwork` | caught |
| unterminated `case` in `j/pulse.cgi` | caught |
| unbalanced quote in `sbin/telegram` | caught |
| `locale.cgi` re-acquiring a shebang | caught (correctly — it would then *claim* to be a script) |
| all four template/escaping cases from #138 | still caught |
| `grep`, `mktemp`, `cd` failing | each exits non-zero |
| empty tree (selector broken) | caught |

Same `sh -n` limit as #138: `if [ -z "$x" ; then` passes, because `[` is just a command name — that was one of my first mutation attempts, and it was a bad test rather than a linter gap.

## Unrelated finding, not touched

`locale_fpv.cgi` has **no reader anywhere in the tree** — only CLAUDE.md:155 and structure.md:49 mention it. `mj-settings.cgi:13` hardcodes `j/locale.cgi` and isn't `fw_variant`-gated, so FPV builds appear to get the standard labels. Either the docs drifted or FPV lost its labels; worth a look independently.

### DIFF
--- a/tools/lint-templates.sh
+++ b/tools/lint-templates.sh
@@ -15,8 +15,19 @@
 # and a gate that cannot run has to say so instead of reaching the success
 # line. An unguarded cd would silently lint the wrong tree, or none at all,
 # and an unguarded mktemp would drop every finding on the floor.
-ROOT=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd) || exit 1
+ROOT=$(CDPATH= cd "$(dirname "$0")/.." && pwd) || exit 1
 cd "$ROOT" || exit 1
+
+# POSIX `read` rather than `head -1 | grep`: head's -1 form is obsolescent (POSIX
+# spells it -n 1) and the pipeline forked twice per file across the whole tree.
+# Returning non-zero for an unreadable file matters — the old `|| continue` let
+# one drop silently out of the run, which is the same class of hole as the
+# guards above.
+first_line() {
+	[ -r "$1" ] || return 1
+	IFS= read -r first_line < "$1" || first_line=''
+	return 0
+}
 
 FAILS=$(mktemp) || exit 1
 # What actually got inspected, so the summary can report real counts. A gate
@@ -50,7 +61,11 @@ trap 'rm -f "$FAILS" "$SEEN"' EXIT
 if ! (
 	cd www/cgi-bin || exit 1
 	find . -name '*.cgi' | sed 's|^\./||' | sort | while IFS= read -r f; do
-		head -1 "$f" | grep -q haserl || continue
+		if ! first_line "$f"; then
+			printf 'www/cgi-bin/%s: cannot read\n' "$f" >> "$FAILS"
+			continue
+		fi
+		case $first_line in *haserl*) ;; *) continue ;; esac
 		echo "t" >> "$SEEN"
 		if ! out=$(haserl -d "$f" < /dev/null 2>&1); then
 			printf 'www/cgi-bin/%s: haserl: %s\n' "$f" "$(printf '%s' "$out" | tail -1)" >> "$FAILS"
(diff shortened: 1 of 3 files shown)


widgetii:
All four fixed in eb6a402 — thanks, they collapsed into a single change.

`head -1` is the obsolescent spelling (POSIX is `head -n 1`), `dirname --`/`cd --` aren't dependable outside GNU coreutils and bought nothing here since `$0` is always a path, the `head | grep` pair forked **146 processes to read 73 first lines**, and — the one that actually matters — `|| continue` treated *"cannot read this file"* identically to *"not a shell script"*, so an unreadable script silently dropped out of the run. That's the same hole as the `cd`/`mktemp`/`grep` ones earlier in this series; good catch.

A `first_line()` helper built on the POSIX `read` builtin covers all four: no forks, no non-POSIX flags, and a distinct non-zero return for unreadable files that callers now report.

Verified:
- selection unchanged — still `ok: 30 haserl templates, 15 shell scripts`
- `chmod 000` on a script now fails the run (rc=1, `sbin/telegram: cannot read`) instead of skipping it
- all 10 fault mutations still caught, 0 missed; all 4 cannot-run guards still fire
- ~14% quicker as a side effect, 0.92s → 0.79s

widgetii:
Declining finding 3 ("shebang scan reads whole files") — the premise doesn't hold here, and the remedy would cost real coverage. Measurements:

**The cited file isn't one long line.** `www/a/bootstrap.min.css` has 4 lines and its *first* line is 21 bytes; `bootstrap.bundle.min.js` has 6 lines with a 4-byte first line. Both carry a comment header before the minified payload, so nothing reads the file to decide it isn't a script:

```
read first_line < www/a/bootstrap.min.css   ->  0.00 s
```

**Even the imagined worst case is immaterial.** A synthetic 1 MB file with zero newlines:

```
read builtin  ->  0.02 s
head -n 1     ->  0.00 s
```

against a job that takes ~0.8s end to end, dominated by 30 haserl invocations.

**And the suggested narrowing has a downside.** Restricting to `www/cgi-bin sbin` walks 50 files instead of 73 and provably changes nothing today (0 shell scripts live outside `www/cgi-bin`) — but it means a script added anywhere else under `www/` would silently never be linted. Given the rest of this series has been about closing exactly that kind of silent gap, trading a coverage property for a 0.02s worst case that doesn't currently occur seems like the wrong way round.

Happy to revisit if a genuinely single-line asset ever lands. The other three findings are fixed in eb6a402.

PR #142, opened 2026-08-12
logs: render timestamps in the viewer's timezone, and stop the header lying

widgetii:
`info-logs.cgi` shows log lines in GMT while the header reads `8/12/2026, 6:52:37 PM Etc/GMT`. Users read that as the logs being in the wrong timezone.

## The logs were right; the header was lying

`www/a/main.js:129-130` rendered the device epoch with `toLocaleString()` and no `timeZone` option — so the **browser's** zone — then appended `json.timezone`, the **camera's** zone name from `/etc/timezone`. Two different clocks concatenated into one string. `www/a/fw-time.js:41` had the identical bug, on the card explicitly labelled the device clock.

## Why nothing downstream could just re-zone the log lines

busybox syslogd destroys the epoch at ingest — `sysklogd/syslogd.c:837`:

```c
timestamp = ctime(&now) + 4;
timestamp[15] = '\0';
```

That string is `sprintf`'d into `printbuf` and `log_to_shmem()` is handed only the string; the `time_t` reaches `log_locally()` alone, for rotation. So every reader — `logread`, `/ws/logs`, SSH — gets `Mmm dd hh:mm:ss` with no year and no zone.

Alternatives considered and rejected:

- **Read the shm ring directly** (`logread.c:118-129`) — drops the fork, but not the string, and takes on busybox's private IPC ABI (the `"GENA"` key, `shbuf_ds` layout, semaphore protocol) as a permanent compatibility surface.
- **`/dev/kmsg`** — the readable/seekable form landed in Linux 3.5; 17 of 125 defconfigs are older (3.0.8 hi3516cv100/hi3520dv200/fh8852v100, 3.0.101 xm510, 3.3.0 gm813x, 3.4.43 gk710x).
- **`syslogd -R` forwarding into majestic** — inverts the roles; the system's log collector would depend on the camera app being alive.

So the conversion happens in the browser, and no daemon is touched. This also means it reaches every camera already in the field via `sbin/updatewebui`, with no firmware flash and no majestic rebuild.

## Changes

| File | |
|---|---|
| `www/cgi-bin/j/logmeta.cgi` | **new** — the UTC offset syslogd is actually running under, plus the camera epoch |
| `www/a/logs.js` | rows convert to the viewer's zone at render time |
| `www/cgi-bin/j/pulse.cgi` | `utc_offset` from the same `date '+%s %z'` as the epoch — no extra fork on the 2 s heartbeat |
| `www/a/main.js` | `parseTzOffsetMs`/`fmtDeviceTime` globals; header on device time + drift badge |
| `www/a/fw-time.js` | device clock card via the same helper |
| `www/cgi-bin/info-logs.cgi` | one line noting the two clocks now differ on purpose |

Two details worth review attention:

`logmeta.cgi` reads syslogd's **own** `/proc/<pid>/environ`, not `/etc/TZ` and not plain `date +%z`. `rcS:4` exports `TZ` once at boot so syslogd keeps the boot zone; `info-logs.cgi:12` restarts it with the CGI's inherited environment; and a restart from a non-login shell leaves it with no `TZ` at all. That last case is why the fallback *unsets* `TZ` — with `TZ=MSK-3` inherited, `date +%z` reports `+0300` while a `TZ`-less syslogd actually stamps `+0000`.

`logs.js` keeps **raw** lines in its buffer and converts only at render time, so the source selector, filter and `sevClass()` all keep working on the untouched wire format. Severity especially must come off the raw line: `sevClass` counts whitespace-separated tokens, and `8/12/2026, 6:52:37 PM` has a different shape, so classing the formatted string would silently match nothing.

The header stays on device time deliberately — it is the one place a camera whose clock is genuinely wrong must look wrong — so real drift past a minute is flagged inline. Skew is timezone-independent, so a correct camera reads zero whatever the two zones are. `/etc/timezone` is never passed to `Intl`: `fw-time.js:13` writes it de-underscored (`America/New York`) and `Intl.DateTimeFormat` throws `RangeError` on that, which is exactly why the numeric offset is shipped separately.

## Verification

On the lab T31 (busybox 1.36.1, musl, no `/etc/localtime`):

- syslogd under `TZ=MSK-3` stamped `Aug 12 20:00:22` for epoch `1786554022` = `17:00:22 UTC` — exactly the `+0300` `logmeta.cgi` reported, while `pulse.cgi` still read `+0000` from httpd's environment. That divergence is the case the endpoint exists for.
- restarted `TZ`-less, it stamped `17:04:16` for `17:04:17 UTC` and `logmeta.cgi` correctly fell back to `+0000`.
- the conversion functions are exercised against those measured instants under four browser zones, including a half-hour one (`Australia/Adelaide`) and a DST one (`America/New_York`), plus the December-in-January rollover.

`tools/lint-templates.sh` passes (29 templates, 16 shell scripts — the new CGI is in scope) and `npm run build:dist` passes.

## Known limit

Lines already in the ring across a DST transition were stamped at the previous offset and read an hour out. The 64 KiB default buffer is typically minutes of logs, and today none of it is convertible at all.

Out of scope: `www/a/files.js:47` still renders file mtimes in browser time, unlabelled — mtimes are used for ordering, and fixing it would need the files API to carry an offset.

### DIFF
--- a/www/a/fw-time.js
+++ b/www/a/fw-time.js
@@ -88,7 +88,7 @@
 			.then(j => {
 				skew = (Number(j.time_now) * 1000) - Date.now();
 				zone = j.timezone || '';
-				offsetMs = parseTzOffsetMs(j.utc_offset);
+				offsetMs = parseTzOffsetMs(j.utc_offset) || 0;
 			})
 			.catch(() => {})
 			.finally(() => { tick(); setInterval(tick, 1000); });
--- a/www/a/fw-time.js
+++ b/www/a/fw-time.js
@@ -35,10 +35,13 @@
 
 	// Live device clock: grab the device epoch once, keep the offset to the
 	// browser clock, then tick locally (no second 2s poller vs the header).
-	let skew = 0, zone = '';
+	let skew = 0, zone = '', offsetMs = 0;
 	function tick() {
 		const el = $('#tz-now');
-		if (el) el.textContent = new Date(Date.now() + skew).toLocaleString() + (zone ? ' ' + zone : '');
+		// Device wall clock, so it has to go through the camera's UTC offset:
+		// toLocaleString() alone renders the *browser's* zone under the
+		// camera's label. fmtDeviceTime/parseTzOffsetMs come from main.js.
+		if (el) el.textContent = fmtDeviceTime(Date.now() + skew, offsetMs) + (zone ? ' ' + zone : '');
 	}
 
 	function init() {
@@ -79,7 +82,14 @@
 
 		fetch('/cgi-bin/j/pulse.cgi')
 			.then(r => r.json())
-			.then(j => { skew = (Number(j.time_now) * 1000) - Date.now(); zone = j.timezone || ''; })
+			// skew = device epoch minus browser epoch, so `Date.now() + skew` is
+			// the camera's clock. Same quantity and sign the header bar reports
+			// as drift; here it is used to tick, not to warn.
+			.then(j => {
+				skew = (Number(j.time_now) * 1000) - Date.now();
+				zone = j.timezone || '';
+				offsetMs = parseTzOffsetMs(j.utc_offset);
+			})
 			.catch(() => {})
 			.finally(() => { tick(); setInterval(tick, 1000); });
 	}
--- a/www/a/logs.js
+++ b/www/a/logs.js
@@ -2,6 +2,10 @@
 // Streams majestic's /ws/logs WebSocket (a single `logread -f` over which klogd
 // has already merged kernel + application logs) and renders it with a source
 // selector, free-text filter, pause, clear and download. Vanilla JS, no deps.
+//
+// The buffer holds raw syslog lines; timestamps are converted to the viewer's
+// timezone only at render time (see fmtLine), so filtering and severity classing
+// keep working on the untouched wire format.
 (function () {
 	const MAX_LINES = 2000;
 	const el = $('#log');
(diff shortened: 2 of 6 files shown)


widgetii:
Both findings were real; fixed in 8db88e7.

**2. Invalid offset treated as UTC** — correct, and the more serious of the two. `parseTzOffsetMs` collapsing bad input to `0` defeated the guard it feeds: `logs.js` holds `tzOffsetMs` at `null` to mean *zone not known yet, keep lines raw*, so a malformed `tz_offset` switched conversion on and rendered every line as though syslogd ran on UTC. On a camera at +0300 that is a silent three-hour shift — worse than the untranslated stamp it replaced. Now returns `null` for anything that is not ±HHMM (minutes bounded too), and `logs.js` leaves `tzOffsetMs` alone when it gets one. A genuine `+0000` is still `0` and still converts, so the two cases stay distinguishable.

The header and the time-settings clock coalesce to `0` instead — neither has a raw form to fall back on, so UTC beats a blank bar there.

**1. Download ignores active filters** — the unfiltered dump is pre-existing (`master` already had `lines.join('\\n')`); what was new was my comment claiming it was "what is on screen". Of the two, the code is the part worth changing: the button sits next to the filter, and handing back lines the user deliberately filtered out is a surprising thing to discover after sharing the file. It now applies the same `matches()` predicate `render()` does.

The conversion functions are covered by a harness that extracts them from the shipped sources and asserts against instants measured on a real T31, run under four browser zones including a half-hour one and a DST one; it gained cases for `null` vs `+0000`.

PR #144, opened 2026-08-12
files: show modified times in camera time, matching recording paths

widgetii:
Follow-up to #142.

## The bug I listed there doesn't exist

#142 flagged `www/a/files.js:47` as "renders file mtimes in browser time, unlabelled", implying the same timezone defect the header had. That was wrong. `files.cgi:19` builds `mtime` from `stat -c '%Y'` — seconds since the Epoch, absolute and timezone-independent — so `new Date(mtime * 1000).toLocaleString()` was already converting it correctly. Nothing was being mis-shifted, and after #142 browser-zone rendering is what log rows do too, so it was *consistent* rather than inconsistent.

## The real problem is narrower

Everything else on this page is already stamped in the camera's zone. majestic writes recordings to `records.path` `/mnt/mmcblk0p1/%F` with `records.filename` `%H-%M` (`app_config.c:203-204`) — both `strftime` in device local time. So a clip lands at:

```
/mnt/mmcblk0p1/2026-08-12/20-00.mp4
```

…and the Modified column beside it read `5:00:00 PM` for a viewer on UTC with the camera on +0300. Far enough east or west it disagreed on the **date** too — from Auckland that file reads as the 13th. And nothing on screen said which clock the column meant, so there was no way to tell the two readings apart.

## Change

The column now follows the file names rather than the log viewer:

- **Modified** renders in camera time, so it agrees with the `%F` directory and `%H-%M` name the user is looking at.
- The `<th>` tooltip says which clock it is; each row's `title` carries the viewer's own equivalent, so correlation works both directions.
- Offset comes from `pulse.cgi`, the same source the header bar uses. Until it answers, the column renders browser time exactly as today; an unparseable offset leaves it there rather than switching to a wrong one.
- `mtime` 0 renders blank instead of `1/1/1970`, and the zone label is escaped before reaching the `title` attribute, being device-derived.

The log viewer deliberately keeps browser-zone rendering — nothing on that page encodes device-local strings, so the reasoning that applies here doesn't apply there.

## Verification

`fmtTime`/`titleTime` extracted from the shipped source and asserted against a real recording's epoch under viewer zones either side of the camera, including `Pacific/Auckland` (+12) where the naive rendering rolls to the next day. Covers the camera-time rendering, the tooltip contents, attribute escaping of the zone label, the pre-`pulse` fallback, and `mtime` 0.

`tools/lint-templates.sh` and `npm run build:dist` pass.

**Not verified:** the browser-side DOM rendering itself hasn't been through a real browser — I have no browser automation here, and `curl` against a camera would only confirm the files copied, not that the JS runs. The logic is unit-tested; the DOM integration is a `title=` attribute on an existing `<td>`. Worth an eyeball on the File Manager page before merge.

### DIFF
--- a/www/a/files.js
+++ b/www/a/files.js
@@ -51,24 +51,42 @@
 	// local time. Showing the browser's clock next to a folder called 2026-08-12
 	// holding 20-00.mp4 can disagree on the hour and, far enough east or west, on
 	// the date. The viewer's own equivalent stays one hover away.
-	let devOffsetMs = null, devZone = '';
+	//
+	// Conversion prefers Intl with the camera's real IANA zone over the flat
+	// offset the header bar uses, because these timestamps are historical. A flat
+	// offset is the *current* one, so on a DST zone every file written the other
+	// side of a transition renders an hour off the %H-%M in its own name — and an
+	// SD card holds months where the log ring holds minutes. /etc/timezone is
+	// stored de-underscored (fw-time.js:13 writes "America/New York"), so putting
+	// the underscores back recovers the name Intl wants; the flat offset stays as
+	// the fallback for a label it will not accept.
+	let devOffsetMs = null, devZone = '', devIana = null;
+	function ianaOrNull(label) {
+		try {
+			const z = String(label).replace(/ /g, '_');
+			new Intl.DateTimeFormat('en', { timeZone: z }).format(0);
+			return z;
+		} catch (e) { return null; }
+	}
 	function fmtTime(s) {
 		const ms = s * 1000;
 		if (!ms) return '';
-		if (devOffsetMs === null) return new Date(ms).toLocaleString(); // pre-pulse
-		return fmtDeviceTime(ms, devOffsetMs); // main.js
+		if (devIana) return new Date(ms).toLocaleString(undefined, { timeZone: devIana });
+		if (devOffsetMs !== null) return fmtDeviceTime(ms, devOffsetMs); // main.js
+		return new Date(ms).toLocaleString(); // pre-pulse
 	}
 	function titleTime(s) {
 		const ms = s * 1000;
-		if (!ms || devOffsetMs === null) return '';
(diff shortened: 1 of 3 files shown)


widgetii:
Both addressed in 5467bcb.

**2. Device mtime breaks on DST** — correct, and it undercut the whole point of the PR. The offset from `pulse.cgi` is *today's*; mtimes are historical. A clip written 15 January (EST, −0500) and listed in August (EDT, −0400) rendered an hour off the `%H-%M` in its own path — the exact mismatch this change exists to remove, applied to months of recordings. The equivalent caveat I documented on the log viewer only covers the minutes an in-RAM ring holds; an SD card is a different scale entirely.

`/etc/timezone` was only unusable because `fw-time.js:13` stores it de-underscored. Putting the underscores back recovers a real IANA name — `America/New York` → `America/New_York`, `Africa/Dar es Salaam` → `Africa/Dar_es_Salaam` — and `Intl` then resolves each timestamp at *its own* historical offset. Validated once with a probe format, so a label `Intl` rejects falls back to the flat offset instead of throwing, and a camera with neither usable keeps browser rendering.

The header bar and time-settings clock deliberately keep the flat offset: both show the current instant, where today's offset is by definition correct, and that path still works for a zone string `Intl` has never heard of.

Regression test added — a January mtime must render EST while the camera currently reports −0400, and must *not* equal the naive current-offset result. My original harness had no DST case at all, which is why this got through.

**1. `fmtTime` not escaped** — no exploitable path today: `mtime` is an integer from `stat -c '%Y'` and `toLocaleString()` cannot emit `<`/`>`/`&`. But `f.mode` and `f.name` beside it are escaped, and the device-derived zone label already sits one concat away in the `title`, so it now goes through `esc()` for consistency and to stay safe if that output ever grows a free-text component.

16 assertions now, green under `UTC`, `Asia/Tokyo`, `America/New_York` and `Pacific/Auckland`.

PR #147, opened 2026-08-14
preview: add an unmute button to the live MSE player

### LINKED ISSUE #289: Feature Request: Add Audio Support to the MSE Web Preview Page
The web preview interface currently uses MSE for video streaming, but it is video-only. To check audio, I must use external RTSP players or separate HTTP endpoints like /audio.opus . This adds extra steps and reduces the usefulness of the web UI for quick monitoring.

Integrate audio into the existing MSE web preview. Since Majestic already serves audio streams (Opus, AAC, PCM, etc.) via HTTP endpoints , the web UI could fetch and play this audio alongside the video.

widgetii:
Closes OpenIPC/majestic#289.

The preview player was hardcoded muted with no way to hear the camera — you had to open an RTSP player or a separate `/audio.opus` tab. This adds a mute toggle and volume slider that unmute the live stream in place.

## How it works

Audio rides the same `/ws/video` stream, negotiated per connection. The player probes which codecs this browser can decode in MP4 (Opus for Chrome/Firefox, AAC for Safari) and, on unmute, reconnects with `&audio=<list>` — the same brief cut as switching Main/Sub. Muting reconnects without it, so the camera stops encoding audio when nobody is listening.

The control only appears when `audio.enabled` is true in the config **and** the browser can actually play one of the codecs the camera produces. If the camera reports no audio track (mic off or not producing), the button says so rather than sitting silent.

Also fixes `freshVideo()`, which forced `muted = true` on every reconnect and would re-silence the element the instant the user unmuted — the cloned `<video>` does not carry the mute state or volume, so both are now re-applied.

## Requires

A majestic build that muxes audio into `/ws/video?...&audio=` (companion daemon change). Against an older daemon the `&audio=` parameter is ignored and the stream stays video-only, so the button simply produces no audio — no breakage.

Verified end to end in headless Chromium against a live camera: unmute reveals the control, negotiates Opus, and the `<video>` element decodes audio.

### DIFF
--- a/www/a/preview-page.js
+++ b/www/a/preview-page.js
@@ -65,6 +65,8 @@
 		if (badge) badge.textContent = 'MJPEG';
 	}
 
+	const mute = $('#mj-mute'), muteLbl = $('#mj-mute-lbl'), vol = $('#mj-vol');
+
 	const player = MajesticVideo.attach(initial, {
 		stream: 0,
 		onState: (s, d) => {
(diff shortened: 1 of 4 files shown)


widgetii:
Addressed the review finding in 3fd70d8:

- **Mute state desync (High)** — `onInit` now clears `wantAudio` when the camera returns a video-only stream (`audioCodec` null), so the player state matches the connection and a later unmute flips `wantAudio` and reconnects instead of hitting the `setAudio` early-return. No extra reconnect — the stream is already what a mute would have produced.

Issue #149, opened 2026-08-15
Safari on macOS: same-origin fetch() fails with "access control checks" — heartbeat dead, firmware upgrade never redirects


widgetii:
Split out of #120, where it was found. **Needs someone with a Mac** — I have no Safari and have exhausted what can be established without one.

## Symptom

In Safari on macOS, `fetch()` from a WebUI page to the *same origin* fails. Chrome on the same camera and network is fine.

Reported by @usa- (#120), console output from simply opening **Firmware → Update**:

```text
Fetch API cannot load http://192.168.0.201/metrics due to access control checks.
    m — status.js:1:1599

Fetch API cannot load http://192.168.0.201/cgi-bin/j/pulse.cgi due to access control checks.
    heartbeat — main.js:1:2035
        setTimeout
            (anonymous function) — main.js:1:3138

Unhandled Promise Rejection: TypeError: Load failed
```

Also observed: clicking **Install update from GitHub** makes Safari **prompt for credentials again**, even though the page itself is already authenticated — i.e. the `/ws/upgrade` handshake is not carrying the session either.

Environment where it reproduces:

| | |
|---|---|
| macOS | Sequoia 15.7.9 (24G830), Intel |
| Safari | 26.6 (20624.4.5.18.1) |
| Camera | gk7205v300 / imx335, firmware `2.6.08.14-lite`, majestic `master+c5bf196c` |
| URL | `http://192.168.0.201/…` (plain HTTP, private IP) |
| Window | New Private Window (normal window **not yet tested** — see below) |

## Impact

This is not limited to the upgrade page. `heartbeat` lives in `main.js`, so it runs on **every** page:

- the top bar (SoC temp, memory, uptime, clock) never updates;
- the status page metrics card is dead;
- on the firmware page, `rebootedAlready()`, `ping()` and `installedNow()` are all `fetch()`, so the post-upgrade redirect can never happen in Safari regardless of the logic — which is what #120 kept chasing.

## Already ruled out (please do not re-test these)

**1. Missing `credentials: "same-origin"` — refuted.**
Both a call that has it and a call that lacks it fail:

| shipped file (2.6.08.14) | call | result |
|---|---|---|
| `status.js` | `fetch("/metrics",{credentials:"same-origin"})` — has it | fails |
| `main.js` | `fetch("/cgi-bin/j/pulse.cgi",{signal})` — lacks it | fails |

[...]

PR #151, opened 2026-08-16
auth: cookie/session login so browser WebSocket handshakes work in Safari

### LINKED ISSUE #150: Auth: cookie/token session so browser WebSocket handshakes (upgrade, preview, logs) work in Safari
## Summary

The WebUI authenticates with **HTTP Basic against `/etc/shadow`**, enforced by majestic on every non-whitelisted path. This breaks browser flows that Basic composes badly with — most visibly **WebSocket handshakes in Safari**, which is the root cause of the firmware-upgrade failure investigated in #149.

**Safari/WebKit does not attach cached HTTP Basic credentials to a WebSocket upgrade handshake** (a longstanding WebKit limitation). majestic therefore `401`s the handshake, Safari pops an interactive credential dialog, and if the user cancels it the upgrade never starts (`ws.onerror` → "Could not start the upgrade."). Chrome carries the credential into the handshake, so it works there — which is why this looked Safari-specific.

This affects **all three** WebSockets the WebUI opens, not just upgrade:

| Endpoint | Used by | No creds | With creds |
|---|---|---|---|
| `/ws/upgrade` | `www/a/fw-update.js` (firmware update) | `401` | `101` |
| `/ws/video`   | `www/a/preview.js` (live MSE player) | `401` | `101` |
| `/ws/logs`    | `www/a/logs.js` (log viewer) | `401` | `101` |

(Verified with raw handshake probes against a camera.)

## Proposed fix: cookie/token session auth (keep Basic for back-compat)

A **session cookie is automatically sent on same-origin WebSocket handshakes by every browser, Safari included.** Issuing one after authentication fixes all three WS flows cleanly, and additionally gives us logout and session expiry that Basic can't — at a small footprint, without ripping Basic out.

This is a **cross-repo change**: majestic (server-side auth) + a small WebUI follow-up.

### majestic side

1. **Mint a session token on successful authentication.** Simplest and lowest-friction: on *any* successful Basic-authenticated response, also emit
   ```
   Set-Cookie: mjsession=<token>; HttpOnly; SameSite=Strict; Path=/
   ```
   (add `Secure` when served over TLS). With this, the first Basic-authenticated page load hands the browser a cookie, and every subsequent request — including the WS handshakes — carries it automatically, so **the WebUI needs no change for the WS flows to start working**.
   - Optionally add explicit `POST /api/v1/login` (validate Basic or a form body against `/etc/shadow`) and `POST /api/v1/logout` (clear the cookie) to support a real login form later.
2. **Token format.** A stateless HMAC keeps it storage-free on a constrained device: `token = base64(user|expiry) . HMAC(boot_secret, user|expiry)`, where `boot_secret` is random, generated once per majestic start and kept in RAM. Reboot invalidates all sessions (acceptable — and consistent with the transitional state during upgrade). A small in-RAM table is also fine.
3. **Accept the cookie as an auth alternative on every protected surface** — `/api/v1/*`, the CGI paths, and crucially `/ws/upgrade`, `/ws/video`, `/ws/logs`. Precedence per request: valid `mjsession` cookie → authenticated; else fall back to Basic; else `401`.
4. **Keep Basic working** for `curl`/CLI/webhooks and older clients (`/api/v1/set`, etc.). This is additive, not a replacement.

### WebUI side

- **Minimal path (recommended):** if majestic sets `mjsession` on the Basic-authenticated page load (step 1 above), the WS clients need **no change** — the cookie rides the handshake. `credentials: 'same-origin'` on existing `fetch()` calls already forwards cookies.
- **Optional UX upgrade:** replace the browser's Basic dialog with a real login form that `POST`s to `/api/v1/login`, plus a logout control that hits `/api/v1/logout`. Touch points: `www/cgi-bin/p/common.cgi` / `p/header.cgi` and a small `www/a/*.js`.
- **Fallback design (only if cookies are rejected server-side):** a `?token=` query param on the WS URLs — the WebUI fetches a short-lived token via an authenticated `fetch()` and appends it in `www/a/fw-update.js`, `www/a/preview.js`, `www/a/logs.js`. The cookie approach is preferred because it needs no WS-client changes.

## Security considerations

- `HttpOnly` (no JS access → limits XSS token theft), `SameSite=Strict` (CSRF), `Secure` when TLS is present, bounded expiry with refresh, per-boot secret.
- The cookie must be **required** — do not open any WS or API path to unauthenticated access.
- Preserve the default-password gate (`common.cgi:check_password`, which 303-redirects to `fw-interface.cgi` until the default password is changed).
- **Plaintext-HTTP caveat:** without TLS the cookie is still observable on the LAN, i.e. the same exposure Basic has today — no worse, but TLS remains a worthwhile orthogonal follow-up (not required to fix this).

## Acceptance criteria

- In **Safari on macOS**, with a non-default password: load the WebUI, then use **firmware upgrade**, **live preview**, and the **log viewer** — all work with **no credential prompt**.
- **Chrome** behaviour unchanged.
- **Basic auth still works** for `curl`/CLI/webhooks (no regression for `/api/v1/*` scripted access).
- No unauthenticated access to any protected endpoint or WebSocket.
- If login/logout is implemented: logout invalidates the session; a fresh session is required afterward.

## Background / evidence

- Discovered while investigating #149 (Safari firmware-upgrade failure). Full research, reporter logs, and the server-side handshake probes are in that thread.
- Root-cause confirmation (against a camera):
  - `/ws/upgrade` no creds → `401 WWW-Authenticate: Basic`; with creds → `101 Switching Protocols`. Same for `/ws/video` and `/ws/logs`.
  - majestic currently sends **no** `Set-Cookie`, and a cookie without Basic → `401` (cookie auth must be added).
- WebKit WebSocket + HTTP-auth limitation: WebKit bug 80362 (WebSocket does not use HTTP authentication credentials).

widgetii:
Closes #150.

## Problem

The WebUI authenticates with HTTP Basic. Safari/WebKit does **not** attach cached Basic credentials to a WebSocket *upgrade* handshake (WebKit bug 80362), so majestic `401`s the handshake and Safari pops (or the user cancels) a credential dialog. This breaks all three WS flows — `/ws/upgrade` (firmware update), `/ws/video` (live preview), `/ws/logs` (log viewer) — and is the root cause of the #149 firmware-upgrade failure. Chrome carries Basic into the handshake, so it "worked there."

## Fix (cross-repo)

majestic mints a `session` cookie and redirects unauthenticated browser navigations to a login page (see the companion majestic PR). Every browser, Safari included, sends that cookie on a same-origin WS handshake, so all three flows start working with no credential prompt. This PR adds the browser-side pieces:

- **`www/login.html`** — a self-contained sign-in page (inline CSS/JS, since `/a/*` assets themselves require auth). POSTs `username`/`password` to majestic's `/login`, then lands on the sanitised `?next=` path (default `status.cgi`).
- **`p/header.cgi` + `main.js`** — a **Sign out** control that `POST`s `/logout` (drops the server session) and returns to the login page.
- **`fw-update.js`** — a reboot clears the in-RAM session (and a form login leaves no cached Basic), so the post-reboot watch's authenticated polls now `401`. Treat that as *"camera is back, session expired"*: `rebootedAlready()` reads a `401` from `pulse.cgi` as positive proof of the reboot, and `confirmUpgrade()` sends the user to `/login.html?next=status.cgi` with **"Updated — please sign in again"** instead of degrading to a generic message. Liveness still rides the unauthenticated `/` ping.
- Docs (`CLAUDE.md`, `README.md`) describe the new flow.

Basic auth is retained for `curl`/CLI/webhooks. Requires a majestic build with the `/login` cookie path; older builds fall back to plain Basic (WS then fails only in Safari, as before).

## Verified on hardware (hi3516av300)

- Real **Safari** firmware upgrade over `/ws/upgrade` completed end-to-end (kernel + rootfs written & verified) — no credential prompt.
- Browser navigation, no creds → `302 /login.html?next=…`; `/login.html` unauth → `200`.
- Login `POST` with `password` last → `200 + Set-Cookie` (heap-fix case).
- `/ws/{upgrade,logs,video}` with cookie → `101`; without → `401`.
- `curl` with Basic → `200`; unauth `/api/v1/config.json` → `401 + WWW-Authenticate: Basic`.
- Logout → `200`, reused cookie → `401`. Post-reboot watch endpoints behave as the hardened code expects (`/`→200, `pulse.cgi`/`fw-update.cgi`→401).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,6 +8,15 @@ WebUI for [OpenIPC Firmware](https://github.com/openipc/firmware) — served on 
 
 Authentication is HTTP Basic against `/etc/shadow` (user `root`, default password `12345`); `common.cgi:check_password` forces a redirect to `fw-interface.cgi` until the default password is changed.
 
+**Session cookies (browser auth).** Majestic (the daemon serving port 80) also mints a `session` cookie so that browser flows Basic composes badly with — most visibly **WebSocket handshakes in Safari**, which never carry a Basic credential — keep working. The pieces:
+
+- `www/login.html` — a **self-contained** login page (inline CSS/JS; it can't pull `/a/*` because those need auth). It `POST`s `username`/`password` to majestic's `/login`, which validates against `/etc/shadow` and returns `Set-Cookie: session=…; HttpOnly; SameSite=Strict`. On success it redirects to the sanitised `?next=` path (default `status.cgi`).
+- Majestic redirects an **unauthenticated browser navigation** (a `GET` that `Accept`s `text/html` and isn't a WS handshake) to `/login.html?next=…` instead of answering `401 WWW-Authenticate: Basic`, so the native Basic dialog never pops. `curl`/CLI/XHR/WebSocket requests still get the `401`+Basic challenge, so scripted access is unchanged. It also auto-mints the cookie on any successful **root** Basic auth, so a client that still sends Basic transparently gets a session too.
+- The cookie rides every later request (same-origin `fetch` with `credentials: 'same-origin'`, and — the whole point — the three WebSocket handshakes `/ws/{upgrade,video,logs}`).
+- **Sign out** — a nav item in `p/header.cgi` (`#nav-logout`), wired in `main.js` to `POST /logout` (invalidates the server session) then navigate to `/login.html`.
+
+This is a majestic-side feature; the WebUI just provides the login page and the logout control. Older majestic builds without the `/login` cookie path fall back to plain Basic (the WS flows then fail only in Safari, as before).
+
 ## Deploying / running
 
 - `sbin/updatewebui [branch]` — fetches a branch zip from GitHub, wipes `/var/www`, then copies `sbin/*` → `/usr/sbin` and `www/*` → `/var/www`. This is the canonical "deploy from source" path. Default branch is `master`.
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@ Web interface uses system credentials for access. Default username is _root_,
 and password is _12345_. You will be asked to change the password at the first login.
 Please note that this will also affect your login via ssh!
 
+Signing in through the web form issues a session cookie; that cookie is what
+lets the live preview, log viewer and firmware-update progress work in every
+browser (Safari does not attach HTTP Basic credentials to a WebSocket
+handshake). HTTP Basic still works for `curl`/CLI/webhook access, and the
+navigation bar's **Sign out** link ends the session.
+
 ### Support
 
 OpenIPC offers two levels of support.
--- a/www/a/fw-update.js
+++ b/www/a/fw-update.js
@@ -303,9 +303,13 @@
 	// Read the version the camera is running NOW. Re-fetches this page instead of
 	// adding an endpoint — fw-update.cgi already renders it, and the value has to
 	// come from the rebooted camera rather than from this stale document.
-	// Returns null if it cannot be established (camera still coming up, or a
-	// --reset run bounced us to the password page), which callers treat as
-	// "unknown", not as "failed".
+	// Returns an object:
+	//   { needsAuth: true }  — reachable but the session is gone (the reboot
+	//                          cleared it); the caller must send the user to log in.
+	//   { version: string }  — the version string it came back on.
+	//   { version: null }    — reachable but the version couldn't be established
+	//                          (camera still coming up), which callers treat as
+	//                          "unknown", not "failed".
 	async function installedNow() {
 		const ctl = new AbortController();
 		const to = setTimeout(() => ctl.abort(), 5000);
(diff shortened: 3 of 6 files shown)


widgetii:
Both bot findings addressed in 2d044a0.

**1. 401 misread as reboot (`fw-update.js`).** Agreed the signal was ambiguous. `rebootedAlready()` no longer treats a 401/403 from `pulse.cgi` as reboot proof — reboot detection stays on the unauthenticated `/` ping. The lost-session handling now lives only in `confirmUpgrade()`, which is reached exclusively *after* a reboot is established (down-then-up ping, `noop`, or an uptime-proven reboot), so a session dropped without a reboot can no longer end the watch early. And since a failed `sysupgrade` reboots too and the version can't be read once the session is gone, the message is now the neutral **"Camera is back — please sign in again"** rather than "Updated".

**2. Logout ignores POST failure (`main.js`).** The Sign out handler now `await`s `/logout` and only navigates to the login page on `r.ok` (200 cleared / 204 nothing-to-clear). A network error or 5xx keeps the user on the page with a retry prompt, so a failed logout no longer implies a sign-out that didn't take.

Both files pass `node --check` and are deployed on the test camera; the server-side endpoints the reboot-watch depends on behave as expected (`/`→200 unauth, `pulse.cgi`/`fw-update.cgi`→401 without a session).

PR #155, opened 2026-08-17
login: a Stay signed in checkbox

widgetii:
WebUI half of the "Stay logged in" proposal from #120 — a checkbox on the login page that asks majestic for a persistent token, so a majestic restart (a firmware upgrade above all) recovers the session silently instead of presenting the login page again.

All mechanism is majestic-side: token format, 90-day sliding expiry, revocation on password change, and the `/etc/majestic.token` kill switch. The page only adds `remember=1` to the login POST when the box is ticked. An **older majestic ignores the extra field**, so the page works unchanged against both.

Verified on a lab hi3516av300 against the majestic counterpart, through this actual page in a browser: tick, sign in, restart majestic, navigate — lands on the status page with no prompt. Unticked preserves today's behaviour exactly (one session cookie, login page after restart).

Design discussion and the full hardware test matrix: see the majestic-side PR.

### DIFF
--- a/www/login.html
+++ b/www/login.html
@@ -35,7 +35,15 @@
 			display: flex; align-items: center; gap: .5rem;
 			margin-bottom: 1rem; font-size: .875rem; color: #adb5bd; cursor: pointer;
 		}
-		.check input { width: auto; margin: 0; accent-color: #0d6efd; }
+		/* The global input rule above styles text fields (full width, padding,
+		   dark background, border). A checkbox inheriting it renders as a padded
+		   dark box on engines that honour those properties on checkboxes, so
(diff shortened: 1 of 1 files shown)


widgetii:
Review addressed in `9d35ee9`: the finding was right — the global `input` rule (full width, padding, dark background, border) applied to the checkbox too, and my `.check input` override only reset `width`/`margin`. Engines differ in how much text-field styling they honour on checkboxes; the ones that honour more rendered a padded dark box. Every inherited property is now reset explicitly and the box gets a fixed 1rem square, keeping `accent-color` for the tick.

Issue #154, opened 2026-08-17
Firmware → Settings → Reset firmware: upgrade-like issues


usa-:
In **Firmware → Settings → Reset firmware**, I’m seeing issues similar to those we discussed for **Firmware → Upgrade** in #120.

The output stops here:

```text
root@openipc-gk7205v300:/tmp# /usr/sbin/sysupgrade -s -n -x
OpenIPC System Updater v1.0.57

Vendor	goke
SoC	gk7205v300
Model	gk7205v300
Kernel	03:54:53 2026-08-17
RootFS	master+6a3cde2, 2026-08-16

Offline operations.

Stop services, sync files, free up memory
```

In Safari, after the camera reboots, a **Basic Authentication** dialog also appears:

<img width="1327" height="582" alt="Image" src="https://github.com/user-attachments/assets/5dbf295e-8b7a-4241-970c-1208482b8019" />

widgetii:
Thanks — two separate things in this report. One is fixed and explained; the other I could not reproduce, so I need a bit more from you.

## The Basic Authentication dialog — fixed

Your instinct that this contradicts the cookie migration was right, and the reason is a gap in that migration rather than a leftover of the old scheme.

The redirect to the login page was only ever wired up for **page navigations**. Majestic decides by looking at the request: a `GET` that `Accept`s `text/html` and isn't a WebSocket handshake gets bounced to `/login.html`; everything else keeps the old `401 WWW-Authenticate: Basic` so that `curl`, webhooks and snapshot pollers carry on working.

The problem is that the WebUI's own background requests don't look like navigations. None of its 35 `fetch()` calls set an `Accept` header, so they arrive as `Accept: */*` — **indistinguishable from curl** at that decision point. The status bar alone polls every two seconds. So the moment a session stops being valid, the browser is handed a Basic challenge, and Safari renders that as the native dialog. Chrome quietly suppresses it for `fetch`, which is exactly why you only ever saw this in Safari.

**Reset firmware is the sharpest way to trigger it.** `-n` erases the overlay, which takes `/etc` with it — including `/etc/majestic.token`, the key behind the "Stay signed in" checkbox. The reboot clears the in-memory sessions too. So after a factory reset there is genuinely nothing left to authenticate with, and every background request on the page you left open gets challenged.

Two changes, both merged:

[...]

PR #157, opened 2026-08-17
webui: send a lapsed session to the login page instead of nowhere

widgetii:
Client half of #154. Pairs with the majestic change that stops answering a browser's own `fetch()` with `WWW-Authenticate` — that removes the Safari Basic dialog, but on its own it trades a confusing prompt for a confusing no-op: the header bar goes stale, a save quietly does nothing, and nothing says why.

## What changed

`apiFetch()` in `main.js` defaults `credentials` to `same-origin` and, on a 401, sends the browser to `/login.html?next=<current page>` — something the native dialog could never do, since it has no idea where you were going.

The returned promise deliberately **never settles**. The navigation is already under way, and letting callers run their `.then()` on a 401 body paints an error onto a page that is in the middle of leaving.

31 call sites across ten files move over.

## Why opt-in and not a `window.fetch` patch

A global wrapper is a smaller diff and catches pages we haven't written yet, but it fights `fw-update.js`, where a 401 is **expected**: `installedNow()` reads it as "the reboot cleared the session" and redirects with wording that depends on knowing an upgrade was in flight (#120, #146). A blanket redirect would race that and throw the transcript away mid-flash.

So `fw-update.js`'s four call sites stay on plain `fetch`, and the helper is something a page asks for.

## Why #154 surfaces it

`fw-reset.cgi` runs `sysupgrade -s -n -x`. `-n` wipes the overlay, which takes `/etc` — and with it `/etc/majestic.token`, the stay-signed-in key. The reboot clears the in-RAM sessions too, so nothing is left to authenticate with, and every background fetch on the still-open page gets a 401. The heartbeat alone fires every two seconds.

## Checks

`npm run build:dist` and `tools/lint-templates.sh` both pass (29 templates, 16 shell scripts). Deployed to a lab hi3516av300 alongside the majestic build; unauthenticated in-page fetches now come back as a bare 401 with no challenge header.

### DIFF
--- a/www/a/ext-ntfy.js
+++ b/www/a/ext-ntfy.js
@@ -10,7 +10,7 @@
 		btn.disabled = true;
 		out.className = 'small ms-2 text-secondary';
 		out.textContent = 'Sending…';
-		fetch('?send=test', { credentials: 'same-origin' })
+		apiFetch('?send=test', { credentials: 'same-origin' })
 			.then(r => r.text())
 			.then(d => {
 				const ok = d.trim() === 'OK';
--- a/www/a/files.js
+++ b/www/a/files.js
@@ -81,7 +81,7 @@
 		return attr(devZone + ' — ' + new Date(ms).toLocaleString() + ' your time');
 	}
 
-	fetch('/cgi-bin/j/pulse.cgi', { credentials: 'same-origin' }).then(r => r.json()).then(j => {
+	apiFetch('/cgi-bin/j/pulse.cgi', { credentials: 'same-origin' }).then(r => r.json()).then(j => {
 		devZone = j.timezone || '';
 		devIana = ianaOrNull(devZone);
 		devOffsetMs = parseTzOffsetMs(j.utc_offset); // main.js; null if unusable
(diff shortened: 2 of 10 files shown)


widgetii:
Both addressed in `7de3746` — the finding was right, and the first half was my error rather than an edge case.

**`makeTextFileLineIterator` reverted to plain `fetch`.** Its only caller is `runCmd()` on `fw-reset.cgi`, which streams `sysupgrade -s -n -x` — the factory reset that erases the overlay, takes `/etc/majestic.token` with it, and reboots. Losing the session is the *expected* end of that request, and showing the log until the camera goes is the whole point of the page. This is exactly the argument the PR description already makes for `fw-update.js`; I stated the rule and then applied it to only one of the two pages it covers.

Worth noting the failure mode is slightly different from the one described: the 401 can only land on the *initial* request, since once the response is streaming the reader just ends when the camera drops. So a mid-stream stall was not reachable — but "session already lapsed when you hit the button" was, and it would have blanked the transcript instead of showing it.

**`location.href` → `location.replace()`.** The bfcache point is a good catch and applies regardless of the above. The promise never settles by design, so anything awaiting it waits forever; leaving a history entry lets Back restore precisely that document and hand the user a page that looks alive and is not. `replace()` leaves nothing to go back to, which is also just right on its own terms — an unauthenticated page is not somewhere anyone wants in their history.

Net effect: `fw-update.js` (4 call sites) and `makeTextFileLineIterator` (1) stay on plain `fetch`; the other 30 go through `apiFetch`.

`npm run build:dist` and `tools/lint-templates.sh` both still pass.

PR #162, opened 2026-08-20
webui: make wiping the camera look different from rebooting it

### LINKED ISSUE #160: Make destructive firmware actions harder to trigger accidentally
There are several actions in the firmware settings that can result in losing the current configuration. I think these actions should be visually and behaviorally distinct from ordinary actions such as restarting the camera, so that they are harder to trigger accidentally.

### Firmware → Settings

The **Restart camera** and **Reset firmware** buttons are currently too easy to confuse:

* they are located relatively close to each other;
* both use the same `Are you sure?` confirmation dialog.

The **Reset firmware** button is already highlighted in red, which certainly helps, but I don't think this alone is enough to prevent accidental activation when both actions have essentially the same confirmation flow.

I think the two actions should be made more clearly distinguishable. For example:

* move the buttons farther apart and/or
* make their confirmation behavior different.

One simple option would be to remove the confirmation dialog from **Restart camera**, since restarting is non-destructive. Conversely, the **Reset firmware** confirmation should contain enough warning text to make the user stop and pay attention before proceeding, rather than being just another simple `Are you sure?` prompt.

### Firmware → Update

The **Reset config (wipe overlay)** option is also destructive, but currently it is presented as an ordinary switch:

> Erases all your settings — equivalent to a fresh flash

I think it should follow the same approach as **Firmware → Settings → Maintenance → Reset firmware**, with appropriate adjustments for a switch rather than a button. In particular, it should:

* be displayed in red;
* have the destructive part of the description (e.g. **Destroys all changes.**) in red;
* require an explicit confirmation in `alert()` with enough warning text to make the user stop and pay attention before the action is performed, as described in the first section.

The goal is not to prevent users from performing these actions, but to make accidental configuration loss much less likely.

widgetii:
Closes #160.

Three actions can throw away every setting on a camera, and none of them looked or behaved differently from the harmless controls beside them.

## Firmware → Settings → Maintenance

`Restart camera` and `Reset firmware` sat in one flex row a gap apart and asked the identical `Are you sure?`. The only thing separating a reboot from a factory wipe was reading the button.

The reset now lives below a rule under a **Danger zone** heading — arrived at rather than landed on — and each prompt says what it is actually about to do:

> **Restart camera** — Restart the camera now?
> Settings are kept. Video and recording stop for about half a minute while it comes back.

> **Reset firmware** — Reset this camera to factory state?
> Every configuration change is erased: network and Wi-Fi, video and image settings, timezone, passwords, and every extension you have configured.
> You will need to set the camera up again from scratch, and reach it on whatever address DHCP gives it. This cannot be undone.

I kept a prompt on the restart rather than dropping it as @usa- suggested. It is non-destructive but not free — it drops video and recording — and differentiating by *wording* fixes the "same confirmation flow" complaint without making an accidental click on a nav button reboot the camera outright. Easy to change if you disagree.

Varying the text needed a mechanism, so `.btn-danger`/`.btn-warning`/`.confirm` now honour `data-confirm`, falling back to the generic question. The restart-required banner's button gets the same text as the one in Settings, since it is the same action.

## Firmware → Update → Reset config (wipe overlay)

The worst of the three: rendered exactly like `Upgrade kernel` directly above it, one click, no question asked.

`field_switch` now takes an optional confirm message, which marks the row `.destructive` — red label, red switch when on — and `main.js` asks **as it is switched ON**, putting it back if the answer is no.

Confirming on the change rather than at submit matters: by the time the form is submitted the user is looking at a different button entirely, and a prompt there would be about the upgrade, not about the wipe.

Turning a destructive switch **off** never asks. Nothing that has not opted in changes — the other 16 `field_switch` call sites render exactly as before.

## Drive-by

`field_switch` only tested for `"true"`, so an off switch emitted its value straight into the tag:

```html
<input type="checkbox" id="fw_reset" ... class="form-check-input" false>
```

A boolean attribute literally named `false`. Harmless, since nothing reads it, but invalid and the sort of thing a later attribute selector trips over. Fixed on the line I was already editing.

## Verified

Rendered on an hi3516av300 rather than read off the template:

```
fw_kernel  ... class="form-check-input" checked
fw_reset   ... class="form-check-input"  data-confirm="Wipe ALL settings…"
fw_force   ... class="form-check-input"
```

plus `Danger zone` and both distinct `data-confirm` strings on the Settings page. `build:dist` and `lint-templates.sh` pass.

### DIFF
--- a/www/a/bootstrap.override.css
+++ b/www/a/bootstrap.override.css
@@ -472,3 +472,26 @@ pre#output[data-cmd] {
 .form-select {
 	transition: border-color var(--mj-ease), box-shadow var(--mj-ease);
 }
+
+/* Destructive switches (issue #160).
+ *
+ * "Reset config (wipe overlay)" sat in Advanced options rendered exactly like
+ * "Upgrade kernel" directly above it — same weight, same colour, one click, no
+ * question asked. Nothing about it looked like the thing that throws away every
+ * setting on the camera. field_switch marks such a row .destructive when it is
+ * given a confirm message; this is what makes it read as one.
+ */
+.boolean.destructive .form-check-label {
+	color: var(--bs-danger);
+	font-weight: 500;
+}
+
+.boolean.destructive .form-check-input:checked {
+	background-color: var(--bs-danger);
+	border-color: var(--bs-danger);
+}
+
+.boolean.destructive .form-check-input:focus {
+	border-color: var(--bs-danger);
+	box-shadow: 0 0 0 .25rem rgba(var(--bs-danger-rgb), .25);
+}
--- a/www/a/main.js
+++ b/www/a/main.js
@@ -263,14 +263,32 @@ function initAll() {
 	$$('form').forEach(el => el.autocomplete = 'off');
 
 	// For .warning and .danger buttons, ask confirmation on action.
+	//
+	// data-confirm overrides the wording. One shared "Are you sure?" meant that
+	// rebooting the camera and wiping it back to factory asked the identical
+	// question, on two buttons sitting next to each other — so the prompt carried
+	// no information and became something to click through (issue #160). The
+	// generic text stays for everything that has not said otherwise.
 	$$('.btn-danger, .btn-warning, .confirm').forEach(el => {
+		const ask = el.dataset.confirm || "Are you sure?";
 		// for input, find its parent form and attach listener to it submit event
 		if (el.nodeName === "INPUT") {
 			while (el.nodeName !== "FORM") el = el.parentNode
-			el.addEventListener('submit', ev => (!confirm("Are you sure?")) ? ev.preventDefault() : null)
+			el.addEventListener('submit', ev => (!confirm(ask)) ? ev.preventDefault() : null)
 		} else {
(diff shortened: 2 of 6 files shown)


widgetii:
Valid, fixed in `5bb7770`.

Worth being precise about the severity: these strings are author literals in the CGI, not user input — the same trust model as the existing `hint` argument, which is interpolated raw *by design* so it can carry a link. So it isn't reachable as an injection. What it is, is a trap for whoever writes the next prompt: a quote would end the attribute early and take the rest of the tag with it, silently.

`attr_escape()` neutralises `&`, `<`, `>` and `"`, then **restores numeric character references**. That last step is the whole difficulty — the prompts use `&#10;` for their line breaks, so escaping the ampersand would put `&#10;` on screen instead of breaking the line.

Checked against the camera's busybox `sed` rather than GNU:

```
newline entity survives : Line one&#10;Line two
quote neutralised       : He said &quot;wipe it&quot; now
tag injection blocked   : &quot;&gt;&lt;script&gt;alert(1)&lt;/script&gt;
bare ampersand escaped  : Wi-Fi &amp; network
```

Rendered output for the existing prompt is unchanged, verified on the camera. `build:dist` and `lint-templates.sh` pass.

PR #165, opened 2026-08-21
Add a camera selector for openipc.local

widgetii:
Every camera on a link answers `openipc.local`, so a browser that goes there has not said *which* one it wants. With a majestic that serves `GET /api/v1/peers`, that request now lands here instead of on one camera's UI, and this page lists what the camera found: itself, and every other camera it has heard announce itself.

**One click each.**

```
OpenIPC
Cameras on this network

  openipc-hi3516av300   YOU ARE HERE          10.216.128.39
  openipc-hi3516ev300                         10.216.128.68
  openipc-ssc30kq                             10.216.128.33
  openipc-hi3516cv300                         10.216.128.52
  openipc-t31                                 10.216.128.35
```

## Notes on the choices

**The serving camera is in the list.** Leaving it out would mean the one camera you are certainly connected to is the one you cannot open — and finding it again would mean looking up its address. It is marked only so you know which one you are already talking to.

**Links are built from addresses, not names.** A camera's own `.local` name resolves only in a browser that speaks mDNS; the address came out of the packet the camera announced itself in. A published URL is preferred when it is `http`/`https` and ignored otherwise — it arrives over the network like everything else here, and `javascript:` in an `href` would run in this page. Everything else is written with `textContent` for the same reason: a name is whatever a device on the link chose to call itself.

**Each entry points at a specific camera**, never back at the shared name. A session belongs to the camera that issued it, and the next request under `openipc.local` may well reach a different one.

**Self-contained**, like `login.html` and for the same reason — served without authentication, since arriving at `openipc.local` you have not yet picked a camera to sign in to, so it cannot pull `/a/*.css` or `/a/*.js`.

**A camera on its way out of the roster is marked**, not offered as though it were up.

**Plain `fetch`, not `apiFetch`** — the latter redirects a 401 to the login page, which is the wrong answer on a page reached before signing in to anything.

## Verified

Rendered in Chromium against a real link of five OpenIPC cameras: five entries, correct hostnames, every link resolvable. Against a build without the endpoint the fetch 404s and the page says so rather than hanging.

Valid HTML5, no deprecated tags, no `/a/` dependencies, and it lands in `majestic-webui-dist.tar.gz` alongside `login.html` with no change to `tools/build-dist.sh`.

### DIFF
--- a/www/cameras.html
+++ b/www/cameras.html
@@ -84,8 +84,12 @@
 					// the camera published is preferred, but only if it is plain
 					// http/https — it arrives over the network like everything else
 					// here, and javascript: in an href would run on this page.
+					//
+					// Used as sent, not with a slash appended: an authority on its
+					// own is already a valid URL, and appending to one that turned
+					// out to carry a path, a query or a fragment would corrupt it.
 					var href = 'http://' + cam.address + '/';
-					if (cam.url && /^https?:\/\//i.test(cam.url)) href = cam.url + '/';
+					if (cam.url && /^https?:\/\//i.test(cam.url)) href = cam.url;
 
 					var a = document.createElement('a');
 					a.className = 'cam';
@@ -132,7 +136,13 @@
 			// login page, which is the wrong move here: this page is reached before
 			// signing in to anything, so an empty list belongs on screen rather
 			// than a redirect to a camera the user has not chosen.
-			fetch('/api/v1/peers', { credentials: 'same-origin', headers: { 'Accept': 'application/json' } })
+			//
+			// And deliberately without credentials. The roster needs none — the
+			// camera decides on the caller's address, not a session — while the
+			// origin here is the shared name, which every camera answers. Sending
+			// the cookie would hand a session minted by one camera to whichever
+			// one happens to reply next.
+			fetch('/api/v1/peers', { credentials: 'omit', headers: { 'Accept': 'application/json' } })
 				.then(function (r) {
 					if (r.status === 404) throw new Error('This camera does not publish a camera list.');
 					if (!r.ok) throw new Error('The camera list is not available here.');
--- a/www/cameras.html
+++ b/www/cameras.html
@@ -98,7 +98,9 @@
 			}
 
 			function addressHref(address) {
-				if (!address) return null;
+				// A non-string reaches split() and colon tests below and throws,
+				// so anything that is not a string is simply not an address.
+				if (typeof address !== 'string' || !address) return null;
 				if (isIPv4(address)) return 'http://' + address + '/';
 				var v6 = ipv6Host(address);
 				return v6 ? 'http://' + v6 + '/' : null;
@@ -145,6 +147,12 @@
 				var ul = document.createElement('ul');
 
 				cameras.forEach(function (cam) {
+					// The array was checked, but its entries were not: a null or a
+					// non-object among them would throw on cam.address and put an
+					// internal error on screen, the very thing the array check
+					// above was there to avoid.
+					if (!cam || typeof cam !== 'object') return;
+
 					// Addressed numerically rather than by name: a camera's own
 					// .local name only resolves in a browser that speaks mDNS, while
 					// the address came out of the packet it announced itself in.
--- a/www/cameras.html
+++ b/www/cameras.html
@@ -79,7 +79,10 @@
 			function isIPv4(s) {
 				var p = s.split('.');
 				return p.length === 4 && p.every(function (o) {
-					return /^\d{1,3}$/.test(o) && +o <= 255;
+					// No leading zeros: "010" is octal to some URL parsers, so a
+					// value that looks like the shown address could resolve to a
+					// different host in the click. Canonical dotted-decimal only.
+					return /^(0|[1-9]\d{0,2})$/.test(o) && +o <= 255;
 				});
 			}
 
@@ -182,10 +185,14 @@
 						a.appendChild(here);
 					}
 
-					if (cam.quietFor > QUIET_AFTER) {
+					// A finite number only: the comparison alone would let a
+					// string or a missing value through and print "no reply for
+					// undefineds".
+					if (typeof cam.quietFor === 'number' && isFinite(cam.quietFor)
+							&& cam.quietFor > QUIET_AFTER) {
 						var q = document.createElement('span');
 						q.className = 'quiet';
-						q.textContent = 'no reply for ' + cam.quietFor + 's';
+						q.textContent = 'no reply for ' + Math.round(cam.quietFor) + 's';
 						a.appendChild(q);
 					}
 
--- a/www/cameras.html
+++ b/www/cameras.html
@@ -83,19 +83,25 @@
 				});
 			}
 
-			// A bracketed literal only — the address field is a host on its own,
-			// never host:port, so any bare colon here is part of an IPv6 address.
-			function isIPv6(s) {
-				if (s.indexOf(':') === -1) return false;
-				try { return new URL('http://[' + s + ']/').hostname === '[' + s + ']'; }
-				catch (e) { return false; }
+			// The address field is a host on its own, never host:port, so any bare
+			// colon makes it an IPv6 address. Returns the parser's own bracketed
+			// form when it is one, so uppercase or uncompressed spellings are
+			// accepted rather than dropped for not matching character-for-character
+			// — and null otherwise. The parser only reads brackets as an IPv6 host,
+			// so nothing else can slip a different authority through here.
+			function ipv6Host(s) {
+				if (s.indexOf(':') === -1) return null;
+				try {
+					var h = new URL('http://[' + s + ']/').hostname;
+					return h.charAt(0) === '[' ? h : null;
+				} catch (e) { return null; }
 			}
 
 			function addressHref(address) {
 				if (!address) return null;
 				if (isIPv4(address)) return 'http://' + address + '/';
-				if (isIPv6(address)) return 'http://[' + address + ']/';
-				return null;
+				var v6 = ipv6Host(address);
+				return v6 ? 'http://' + v6 + '/' : null;
 			}
 
 			// The camera's own URL is preferred when it is plainly http or https,
@@ -109,7 +115,7 @@
 						var u = new URL(cam.url);
 						var host = u.hostname.replace(/^\[|\]$/g, '');
 						if (!u.username && !u.password
-								&& (isIPv4(host) || isIPv6(host))) {
+								&& (isIPv4(host) || ipv6Host(host))) {
 							return cam.url;
 						}
 					} catch (e) { /* unparseable: fall through to the address */ }
--- a/www/cameras.html
+++ b/www/cameras.html
@@ -69,6 +69,37 @@
 			// so rather than offer it as though it were up.
 			var QUIET_AFTER = 60;
 
+			// Where a link is allowed to point. Everything in the roster reached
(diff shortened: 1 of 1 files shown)


widgetii:
Both fixed in deccfb9.

**Preauth fetch sends cookies** — right, and it is the exact hazard this page exists to steer people away from, committed by the page itself. On `openipc.local` the origin is the *shared* name that every camera on the link answers, so `credentials: 'same-origin'` would hand a session minted by one camera to whichever replied next. It needs no credentials either way: the camera decides whether to answer the roster by the caller's address, not by a session. Now `credentials: 'omit'`.

**Mutates advertised camera URL** — right. Appending a slash is harmless against what majestic sends today (a bare scheme and authority), but an authority on its own is already a valid URL, and the day one carries a path, query or fragment the append corrupts it. Used as sent.

Re-rendered in Chromium against the link, which had gained a sixth camera by then:

```
6 links
  http://10.216.128.39:80   openipc-hi3516av300 | you are here | 10.216.128.39
  http://10.216.128.68:80   openipc-hi3516ev300 | 10.216.128.68
  http://10.216.128.49:80   openipc-gk7205v200  | 10.216.128.49
  http://10.216.128.33:80   openipc-ssc30kq     | 10.216.128.33
  http://10.216.128.52:80   openipc-hi3516cv300 | 10.216.128.52
  http://10.216.128.35:80   openipc-t31         | 10.216.128.35
```

Each entry now points at exactly the URL its camera published, with no trailing slash added.

widgetii:
Both fixed in 6b5615d.

**Address href injection** — right, and the reasoning is the important part: everything in the roster arrived over the network, and this page cannot see whatever guarantee the camera makes about it. `'http://' + address + '/'` with an address of `10.0.0.1@evil.tld` makes the part before the `@` userinfo and sends the click to `evil.tld`; a `/`, `?` or `#` does the same with a path, query or fragment. A link that goes somewhere the list never named is about the worst outcome available on a page whose whole job is "click the camera you want".

An entry is now turned into a link only once it is known to name a host and nothing else — digits, dots and colons, which is the whole of an IPv4 or IPv6 literal and cannot begin any of those parts. A published URL is checked the same way by its parsed hostname, with userinfo rejected. Anything left over is dropped from the list rather than linked.

Checked against the cases that motivated it:

```
  address                 url                      -> href
  10.216.128.39           -                        -> http://10.216.128.39/
  fe80::1                 -                        -> http://[fe80::1]/
  10.0.0.1@evil.tld       -                        -> (dropped)
  10.0.0.1/../../etc      -                        -> (dropped)
  evil.tld                -                        -> (dropped)
  10.0.0.1?x=1            -                        -> (dropped)
  10.0.0.1#frag           -                        -> (dropped)
  10.0.0.1                http://x@evil.tld        -> http://10.0.0.1/
  10.0.0.1                javascript:alert(1)      -> http://10.0.0.1/
  10.0.0.1                https://10.0.0.1:443     -> https://10.0.0.1:443
```

That also fixed IPv6 in passing — an address with colons needs brackets before a port can follow it, and was being pasted in bare.

**Fetch can hang forever** — right. Bounded with `AbortController` at five seconds, following `main.js`'s heartbeat, and a timeout now says so rather than looking like an empty network.

Re-rendered against six cameras on the link afterwards; all six still listed.

widgetii:
Both fixed in d6d8eda. Both existed because my check tested which *characters* an address contained rather than whether it *was* an address — a lazy approximation where structure was needed.

**Hex hostname allowed** — right. `deadbeef` or `1234` is all hex-and-dots, passed the character class, and became `http://deadbeef/` — a DNS name the roster never meant. An IPv4 is now four decimal octets each ≤ 255, and an IPv6 is whatever the URL parser accepts as a bracketed host. `deadbeef` is neither.

**IPv4:port link broken** — right. Any colon was read as IPv6 and bracketed, so `10.0.0.1:8080` became `http://[10.0.0.1:8080]/`. The roster's `address` field is a host on its own — a port lives in `url` — so a v4 address with a colon in it is simply not a valid address here and is dropped, not mangled.

Re-tested, 16 cases, all pass:

```
deadbeef              -> dropped     (was http://deadbeef/)
1234                  -> dropped
10.0.0.1:8080         -> dropped     (was http://[10.0.0.1:8080]/)
999.0.0.1             -> dropped     (octet out of range)
10.0.0               -> dropped      (too few octets)
10.216.128.39         -> http://10.216.128.39/
fe80::1               -> http://[fe80::1]/
http://x@evil.tld     -> http://10.0.0.1/   (userinfo rejected, falls back)
https://[fe80::1]:443 -> https://[fe80::1]:443
```

Rendered against six cameras on the link afterwards; all six listed.

widgetii:
Both fixed in cada575.

**Unvalidated cameras response type** — right. `render()` checked `.length` then called `.forEach`, so a `data.cameras` that was not an array threw, and the catch put an internal error on screen. Anything not an array is now treated as no cameras and lands on the empty state.

**Blank list after filtering** — right, and it is the interaction with the previous fix: dropping every entry whose address was not safe to link to could reach `replaceChildren` with an empty `<ul>` — a blank card. If nothing survives the filter it now says "No cameras with a usable address."

Driven against a fake DOM:

```
normal (2 good)      -> list of 2
all filtered out     -> "No cameras with a usable address."
empty array          -> "No cameras answered."
null                 -> "No cameras answered."
object not array     -> "No cameras answered."
string               -> "No cameras answered."
```

No case throws.

widgetii:
Fixed in e77f91f. Right — and it is the flip side of the fix two rounds ago: tightening the check from "address-shaped characters" to "a real address" made it reject valid IPv6 written any way but the parser's own. `2001:DB8::1` (uppercase) and `2001:db8:0:0:0:0:0:1` (uncompressed) normalise to `[2001:db8::1]` and so failed the exact-match, and those cameras dropped off the list.

It now accepts anything the URL parser reads as a bracketed IPv6 host and uses that canonical form for the link, so the spelling the camera sent no longer decides whether it shows up. A published `url` is still used verbatim — only its host is validated — so this touches address-derived links only.

```
2001:DB8::1            -> http://[2001:db8::1]/     (was dropped)
2001:db8:0:0:0:0:0:1   -> http://[2001:db8::1]/     (was dropped)
deadbeef               -> dropped
10.0.0.1:8080          -> dropped
http://x@evil.tld      -> falls back to http://10.0.0.1/
```

Rendered against six cameras on the link afterwards; all six listed.

widgetii:
**Advertised URL can misdirect** — fixed in 5ecfa60. Right, and a real gap: a published `url` was accepted whenever its host was a literal address, but never required to be the *same* address the row shows. A hostile announcement could display `10.0.0.1` and link to `10.0.0.9` — both are literal hosts, and the roster is untrusted. The URL's host must now equal the displayed address (compared on canonical hrefs, so an equivalent IPv6 spelling still matches); otherwise the link falls back to the shown address. A URL that only adds a port or scheme to the same host is still honoured.

```
address    url                    -> href
10.0.0.1   http://10.0.0.9/       -> http://10.0.0.1/       (mismatch: use address)
10.0.0.1   http://10.0.0.1:8080/  -> http://10.0.0.1:8080/  (same host: honoured)
fe80::1    http://[FE80::1]:80/   -> http://[FE80::1]:80/   (equivalent v6 spelling)
fe80::1    http://[fe80::2]/      -> http://[fe80::1]/       (different addr: use address)
```

**Selector not default landing** — I'd like to leave this one; I believe it's a false positive from the reviewer not seeing the other repo. The root routing is not this page's job and isn't left unchanged — it is handled on the server side, in the majestic PR this depends on (widgetii/majestic#375, now merged): `GET /` with `Host: openipc.local` and peers known returns `302 → /cameras.html` *before* `index.html` is considered. `index.html` still refreshing to `status.cgi` is correct and intended for a browser that reached a camera by its own name or address — that browser *did* choose a camera and should land on it, not on the selector. Verified on hardware:

```
Host: openipc.local              -> 302 /cameras.html
Host: <camera-ip>                -> 200 (status.cgi via index.html)
```

So the selector is reachable exactly when it should be, and the page needs no root-routing change of its own. Happy to reconsider if a maintainer sees a case I'm missing.

widgetii:
**Timeout cleared too early** — fixed in 6bb5a76. Right, and it undercut the timeout added last round: `clearTimeout` ran when the headers arrived, but `r.json()` reads the body afterward, so a camera that answered and then stalled mid-body escaped the deadline. The timer is now cleared only after the body is parsed, so the five seconds cover the whole exchange.

The other two still showing open:

- **Fetch can hang forever** (#10) — this is the parent of the above; the timeout fix plus this refinement is its resolution.
- **Selector not default landing** (#5) — I believe this is a false positive; my reply above lays it out. The `/` → `/cameras.html` redirect lives server-side in the majestic PR this depends on (merged), so the selector *is* reachable from `openipc.local`, and `index.html` refreshing to `status.cgi` is correct for a browser that reached a camera by its own name. Happy to be shown a case I'm missing.

widgetii:
**Malformed entries can crash** — fixed in 84eb192. Right: `render()` guarded the array but not its elements, so a `null` or a string among them threw on `cam.address`. Each entry is skipped now unless it is an object, and `addressHref` treats a non-string address as no address rather than calling `.split()` on it — which a numeric or object address would also have thrown on. Driven against a fake DOM with all those shapes mixed in; every good entry still renders and nothing throws.

**IPv6 links never built** — I do not think this one is correct, and I want to show the check rather than just assert it. `ipv6Host()` requires `new URL('http://[' + s + ']/').hostname` to start with `[` — and for an IPv6 host, it does; the hostname is returned **with** the brackets:

```
> new URL('http://[fe80::1]/').hostname
'[fe80::1]'                    // starts with '[' -> passes
> addressHref('fe80::1')
'http://[fe80::1]/'
> cameraHref({address:'2001:db8::1'})
'http://[2001:db8::1]/'
```

So an IPv6 camera does get a link, and the earlier round's fix (accepting uppercase/uncompressed spellings) still holds. If there is a JS engine where `URL.hostname` strips the brackets from a v6 literal I would genuinely like to know, but on V8 it keeps them, and the WHATWG URL standard specifies the bracketed serialization. Leaving `ipv6Host` as is unless a maintainer can point at an environment where this fails.

widgetii:
Both fixed in 9437159.

**Non-canonical IPv4 accepted** — right. `isIPv4` allowed leading zeros, and since the link is built from the string as written, `010.0.0.1` could be displayed while a URL parser that reads `010` as octal navigated somewhere else — the same show-one/reach-another gap as the advertised-URL finding, via the address field. Canonical dotted-decimal only now (a lone `0` stays valid; a zero-prefixed multi-digit octet does not).

```
010.0.0.1  -> dropped     10.00.0.1 -> dropped     01.2.3.4 -> dropped
0.0.0.0    -> http://0.0.0.0/         127.0.0.1 -> http://127.0.0.1/
```

**Quiet marker shows garbage** — right. `cam.quietFor` went straight into the text, so a missing or non-numeric value printed "no reply for undefineds". It shows only for a finite number now, and rounds it. Checked against a string, undefined, null, NaN and Infinity — none render a marker.

---

A note on where this stands: this is the ninth review round, and I want to flag that the last few have shifted from finding new problems to circling the guards added in earlier rounds — including one round that reported `IPv6 links never built` against code I demonstrated works (`URL.hostname` keeps the brackets on V8, per the WHATWG serialization). These two are genuine and worth taking, but if the next round surfaces only re-readings of the same input-validation code rather than new behaviour, I'd suggest the roster-hardening is at a reasonable stopping point — the page is a read-only list of links whose targets are all validated to be literal addresses matching what is shown.

PR #167, opened 2026-08-21
Add an in-UI camera switcher to the navbar

widgetii:
The `openipc.local` camera selector (#165) only helps someone who already knows to type that name. Anyone administering a camera by its IP or its own hostname never learns the others on the LAN exist. This surfaces them from **inside the authenticated UI** — a switcher in the top navbar, on every page, listing the other OpenIPC cameras on the link with one-click jump to each.

```
Information  Majestic  Firmware  Tools  Extensions  Preview  📷 openipc-hi3516av300 +4 ▾  Sign out
                                                                    │
                                              openipc-gk7205v200   10.216.128.49
                                              openipc-hi3516ev300  10.216.128.68
                                              openipc-hi3516cv300  10.216.128.52
                                              openipc-t31          10.216.128.35
                                              ─────────────────────────────────
                                              See all cameras…
```

## How it works

Fed by `GET /api/v1/peers` — the same roster the `openipc.local` page uses. That endpoint is **subnet-gated** and **absent on older majestic** (a 404 either way), and the switcher treats both as "nothing to show" and stays hidden. So there is **no new disclosure and no hard version dependency**: on a camera without the backend, or reached from off the discovery link, the control simply does not appear. Its presence means there are others here.

The label is **this camera's own hostname**, coloured with the brand accent and given a camera glyph so a user-set name reads as *the device you are on* rather than another menu word; a muted **+N** says how many others exist. The dropdown lists them by hostname and address, with **See all cameras…** through to `/cameras.html`.

## Shape

- Lives in `header.cgi` (the nav, included on every page) + a new `/a/cameras-switch.js` loaded after `main.js`, whose `apiFetch` it reuses. No new page, no backend change.
- Re-reads the roster on a lazy **60 s** timer, not off `main.js`'s 2 s heartbeat — peers change slowly (30 s discovery sweep).
- The href validation is **copied from `www/cameras.html`**, not shared: that page is served pre-auth and cannot pull `/a/*.js`. The roster is untrusted network input in both, so the copies must stay identical. A link is built only when its target is a literal IP **matching the address shown**; a published `url` is honoured only when its host equals that address and carries no userinfo; malformed entries are dropped; names are written with `textContent`.

## Verified

Against a link of six OpenIPC cameras (headless Chromium, live authenticated session):

- switcher shows this camera + the five others, each a validated link, `See all` → `/cameras.html`
- a roster of **only this camera**, an **empty/non-array** body, or a **404** endpoint all leave it **hidden**
- **malformed** (`null`, non-string address, `evil.tld`, `10.0.0.1@evil.tld`, leading-zero octets) and **misdirecting** (`url` host ≠ shown address) entries are dropped
- IPv6 addresses (incl. uppercase/uncompressed) bracket correctly

Valid HTML5, no new dependency, no `/a/*` requirement (self-populates via JS); ships in the dist tarball with no `build-dist.sh` change.

Needs a majestic that serves `GET /api/v1/peers`; against a build without it the switcher stays hidden.

### DIFF
--- a/www/a/cameras-switch.js
+++ b/www/a/cameras-switch.js
@@ -100,10 +100,13 @@
 			var a = document.createElement('a');
 			a.className = 'dropdown-item d-flex align-items-center gap-2';
 			a.href = entry.href;
-			// main.js forces every http(s) link to target=_blank, so these
-			// cross-origin peer links open in a new tab. Sever the opener so the
-			// peer page cannot reach back through window.opener and navigate this
-			// admin tab (tabnabbing).
+			// Open peers in a new tab, matching how the rest of the UI treats
+			// external http links. main.js sets target=_blank on the anchors
+			// present at load, but the switcher builds these later (async, and
+			// again on every refresh), so set it here rather than rely on that
+			// one-time pass. rel=noopener severs window.opener so the peer page
+			// cannot reach back and navigate this admin tab (tabnabbing).
+			a.target = '_blank';
 			a.rel = 'noopener noreferrer';
 
 			var label = document.createElement('span');
--- a/www/a/cameras-switch.js
+++ b/www/a/cameras-switch.js
@@ -134,13 +134,29 @@
 		var box = document.getElementById('cam-switch');
 		if (!box) return;
 
-		// apiFetch (main.js) carries the session and never 401s this endpoint —
-		// 200 on-link, 404 off-link or on a majestic without it. Anything but a
-		// good JSON body leaves the switcher hidden.
-		apiFetch('/api/v1/peers', { headers: { 'Accept': 'application/json' } })
+		// Plain fetch, not apiFetch, and without credentials — matching
+		// www/cameras.html. The roster needs no session (the camera answers on
+		// the caller's address, and the endpoint is subnet-gated, not
+		// auth-gated), and on the shared openipc.local origin a cookie could
+		// otherwise be handed to whichever camera replies next. apiFetch's only
+		// extra — a redirect to login on 401 — is wrong here anyway: this runs
+		// unprompted in the background, not on a user action.
+		//
+		// Bounded so a stalled reply self-terminates well before the next tick
+		// rather than piling up behind the timer. 200 on-link, 404 off it or on
+		// a majestic without the endpoint; anything else hides the switcher.
+		var ctl = new AbortController();
+		var giveUp = setTimeout(function () { ctl.abort(); }, 5000);
+
+		fetch('/api/v1/peers', {
+			credentials: 'omit',
+			signal: ctl.signal,
+			headers: { 'Accept': 'application/json' }
+		})
 			.then(function (r) { return r.ok ? r.json() : null; })
 			.then(function (data) { build(box, data && data.cameras); })
-			.catch(function () { hide(box); });
+			.catch(function () { hide(box); })
+			.finally(function () { clearTimeout(giveUp); });
 	}
 
 	function start() {
--- a/www/a/cameras-switch.js
+++ b/www/a/cameras-switch.js
@@ -100,6 +100,11 @@
 			var a = document.createElement('a');
 			a.className = 'dropdown-item d-flex align-items-center gap-2';
 			a.href = entry.href;
+			// main.js forces every http(s) link to target=_blank, so these
+			// cross-origin peer links open in a new tab. Sever the opener so the
+			// peer page cannot reach back through window.opener and navigate this
+			// admin tab (tabnabbing).
+			a.rel = 'noopener noreferrer';
 
 			var label = document.createElement('span');
 			label.textContent = entry.cam.name || entry.cam.address;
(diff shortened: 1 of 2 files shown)


widgetii:
Both fixed in 9a3dd9d — and both were the switcher diverging from what `cameras.html` already does for this exact endpoint, which is a fair catch.

**Peers request leaks session** — right. The roster needs no credentials: it's subnet-gated, not auth-gated, and answered on the caller's address. `apiFetch` sent the session cookie anyway, and on the shared `openipc.local` origin that cookie could reach whichever camera answers next — the crossing `cameras.html` uses `credentials: 'omit'` to avoid. `apiFetch`'s only other effect, a redirect to login on 401, is wrong for a request that runs unprompted in the background. Now a plain `fetch` with `credentials: 'omit'`, matching `cameras.html`.

**Unbounded interval fetch overlap** — right. No deadline meant a stalled reply could pile up behind the 60s timer. Bounded with an `AbortController` at 5s, the way `cameras.html` and `main.js`'s heartbeat already bound theirs — each poll self-terminates long before the next tick, so nothing accumulates.

Re-verified live on the six-camera link: with credentials omitted the switcher still populates on-link (the endpoint doesn't need the session) — this camera + the four others, `+4`, See all.

widgetii:
Addressed both new findings in 9586477:

- **404 polls never stop** — a 404 from \`/api/v1/peers\` (older majestic without the endpoint, or off-subnet) is now treated as definitive: the refresh interval is cleared, so the tab stops re-requesting every 60s. Transient non-404 errors (e.g. 500) still keep polling so a recovering camera is picked up.
- **Missing noopener on links** — \`main.js\` rewrites every \`http(s)\` anchor to \`target=_blank\`, so these cross-origin peer links open in a new tab; they now carry \`rel="noopener noreferrer"\` to sever \`window.opener\` (tabnabbing). The same-origin "See all" link is left unchanged.

Verified with the offline DOM harness (16 assertions): peer links carry the rel and the "See all" link does not; a 404 clears the interval while a 500 does not; self-only and non-array rosters still hide the control.

widgetii:
> 404 polls never stop

This one is already addressed in 9586477 — the finding is quoting the pre-fix snippet. The current code (right-side of the diff) captures the interval and clears it on a definitive 404:

```js
var timer = null;
function stop() { if (timer) { clearInterval(timer); timer = null; } }
// ...
    .then(function (r) {
        if (r.status === 404) { stop(); return null; }   // absent endpoint or off-subnet: stop polling
        return r.ok ? r.json() : null;
    })
// ...
function start() {
    refresh();
    timer = setInterval(refresh, REFRESH_MS);
}
```

So a stable 404 (older majestic without the endpoint, or off-subnet) clears the timer after the first request and never polls again. Non-404 errors (e.g. a transient 500) intentionally keep polling so a recovering camera is still picked up. The offline DOM harness asserts exactly this: 404 → `clearInterval` called once; 500 → interval untouched.

widgetii:
Addressed in 89e390d:

- **Peer links lose _blank** — correct catch. My previous comment leaned on `main.js`s one-time `target=_blank` pass, which runs at load and never sees the switchers async-built links. Now set `target="_blank"` explicitly (alongside `rel="noopener noreferrer"`), so peers open in a new tab consistently with the rest of the UI, on every rebuild.

- **404 polls never stop** — this is fixed as of 9586477; the finding is quoting the pre-fix hunk (anchored to the old line range). Current code captures the interval and clears it on a definitive 404 (`if (r.status === 404) { stop(); return null; }`, where `stop()` calls `clearInterval`). The harness asserts 404 → `clearInterval` once, 500 → interval untouched. This new commit re-shuffles the diff, so a fresh pass should re-anchor and resolve it.

widgetii:
Both remaining findings are by-design / already-fixed rather than open defects:

**Duplicated href validation logic** — this duplication is deliberate and documented. `www/cameras.html` is served *before* authentication (you have not chosen a camera to sign in to yet), so it cannot pull `/a/*.js`, which require a valid session. That makes a shared module impossible: the switcher (authenticated, loads `/a/*.js`) and `cameras.html` (pre-auth, self-contained) cannot import from a common file. The risk it raises — a validation fix landing in one copy but not the other — is mitigated in-tree: both files open with a comment stating the helpers must stay identical ("keep in step with www/cameras.html" / "change one, change the other"), and the copy is verbatim. Deduplicating would require either inlining the same JS into `cameras.html` at build time or serving the validator unauthenticated — a larger architectural change, out of scope for a discoverability PR. Flagging the constraint is fair; the current shape is the intended trade-off.

**404 polls never stop** — this is fixed (9586477); the finding is quoting the pre-fix hunk (anchored to the old `R156-165` range, which my commits shifted). Current pushed code:

```js
var timer = null;
function stop() { if (timer) { clearInterval(timer); timer = null; } }
// ... in refresh():
    .then(function (r) {
        if (r.status === 404) { stop(); return null; }   // absent endpoint / off-subnet: stop
        return r.ok ? r.json() : null;
    })
// ... in start():
    refresh();
    timer = setInterval(refresh, REFRESH_MS);
```

There is exactly one `setInterval` in the file, captured to `timer`, and a definitive 404 clears it after the first request — so a stable 404 (older majestic, or off-subnet) never polls again. Non-404 errors still poll so a recovering camera is picked up. The offline DOM harness asserts this directly (404 → `clearInterval` once; 500 → interval untouched).

PR #169, opened 2026-08-22
endpoints: show the address the browser actually used

### LINKED ISSUE #164: Majestic → Endpoints: use request host/IP for endpoint links instead of physical interface address
The **Majestic → Endpoints** page currently generates endpoint URLs using the camera's physical network interface address.

This can lead to incorrect links when the WebUI is accessed through another network path, such as VPN, reverse proxy, or a DNS hostname.

For example, if the camera has a local address:

```
192.168.0.201
```

but the user opens the WebUI through VPN:

```
http://10.8.0.5/
```

the endpoint links are still generated using the local address:

```
rtsp://192.168.0.201/stream=0
http://192.168.0.201/image.jpg
http://192.168.0.201/api/v1/config.json
```

They should instead use the address by which the WebUI was accessed:

```
rtsp://10.8.0.5/stream=0
http://10.8.0.5/image.jpg
http://10.8.0.5/api/v1/config.json
```

The endpoint URL generation should use the request hostname/IP (for example, the HTTP `Host` header) instead of the camera's physical interface address.

This would make the generated links work correctly in multi-interface setups, VPN environments, and installations where the camera is accessed through a hostname or proxy.

If the request hostname cannot be determined, the current behavior can be used as a fallback.

widgetii:
Fixes #164.

### The bug

`mj-endpoints.cgi` built all 27 URLs from `$network_address` — the source address of the camera's default route (`p/common.cgi:540`). Reach the WebUI over a VPN, a reverse proxy or a hostname and every link points at a LAN address the client can't resolve.

### Why it isn't a `HTTP_HOST` fix

The issue suggests reading the `Host` header server-side. That can't work here. Majestic's CGI bridge (`on_cgi_bin`, `src/websrv/api.c`) sets exactly `QUERY_STRING`, `SCRIPT_NAME`, `HTTP_HOST`, `HTTP_REFERER`, `HTTP_COOKIE`, `REQUEST_METHOD`, `CONTENT_TYPE`, `CONTENT_LENGTH` — no `SERVER_PORT`, no scheme. And `HTTP_HOST` comes from libevent's `evhttp_request_get_host()`, which deliberately **strips the port** off the header. So the haserl would:

- **lose the port** — `http://proxy:8080/` renders as `http://proxy/image.jpg`, which is the reverse-proxy case the issue is about (`system.webPort` is configurable 1–65535);
- **guess the scheme wrong** — majestic serves real TLS on `system.httpsPort`, so the links would stay `http://`/`ws://` and get blocked as mixed content;
- **render an unvalidated attacker-supplied header** into the page.

### What this does instead

Host and scheme become spans that `main.js` fills from `location` — the same source `logs.js`, `preview.js` and `fw-update.js` already use to build their WS handshake URLs:

| span | filled with |
|---|---|
| `.ep-http` | `http` / `https` per `location.protocol` |
| `.ep-ws` | `ws` / `wss` |
| `.ep-host` | `location.host` — host and port |
| `.ep-addr` | `location.hostname` — no port, for RTSP |

The server still renders `$network_address` into those spans, so with JS off the page is byte-for-byte what it is today. `.cp2cb` reads `textContent` at click time, so click-to-copy picks up the rewritten text with no change.

RTSP is the one endpoint not on the WebUI's origin: it gets the host without a port, plus `rtsp.port` from the live config when that isn't the default 554 — which the page ignored entirely before. The same `config.json` fetch that already backed the `system.unsafe` notice is reused, so there's no extra request.

The remote-webhook snippets on **Settings** and **Telegram** carried the same hardcoded address and get the same two spans.

### Caveat

An `rtsp://` link aimed at a VPN or proxy host only resolves if 554/TCP is forwarded too — the WebUI can't know that. It's still strictly better than today's LAN address, which is unreachable by definition in that scenario.

### Testing

Headless chromium against the rendered page, three ways:

| accessed as | RTSP | WS | HTTP |
|---|---|---|---|
| `http://127.0.0.1:8731/` | `rtsp://127.0.0.1:8554/stream=0` | `ws://127.0.0.1:8731/ws/video?stream=0` | `http://127.0.0.1:8731/mjpeg` |
| `https://127.0.0.1:8732/` | `rtsp://127.0.0.1:8554/stream=0` | `wss://127.0.0.1:8732/ws/video?stream=0` | `https://127.0.0.1:8732/mjpeg` |
| `http://[::1]:8733/` | `rtsp://[::1]:8554/stream=0` | `ws://[::1]:8733/ws/video?stream=0` | `http://[::1]:8733/mjpeg` |

The IPv6 row caught a bug in the first draft: `location.hostname` already brackets an IPv6 literal, so bracketing it again gave `[[::1]]`. Both `host` and `hostname` keep the brackets; no special case needed.

`tools/lint-templates.sh` passes (29 templates, 16 shell scripts).

### DIFF
--- a/www/a/main.js
+++ b/www/a/main.js
@@ -335,6 +335,20 @@ function initAll() {
 		});
 	});
 
+	// Endpoint URLs are rendered server-side against the camera's default-route
+	// address, which is wrong the moment the WebUI is reached over a VPN, a
+	// reverse proxy or a hostname (issue #164). Rewrite them to the address the
+	// browser actually used. This has to happen here rather than in the haserl:
+	// majestic's CGI bridge passes neither the port nor the scheme, and libevent
+	// strips the port off HTTP_HOST. Runs before the .cp2cb wiring below so a
+	// copy picks up the rewritten text.
+	const epTls = location.protocol === 'https:';
+	$$('.ep-http').forEach(el => el.textContent = epTls ? 'https' : 'http');
+	$$('.ep-ws').forEach(el => el.textContent = epTls ? 'wss' : 'ws');
+	// host and hostname both keep the brackets an IPv6 literal needs in a URL
+	$$('.ep-host').forEach(el => el.textContent = location.host);      // host[:port]
+	$$('.ep-addr').forEach(el => el.textContent = location.hostname);  // RTSP has its own port
+
 	// click-to-copy for .cp2cb snippets (HTTPS uses the clipboard API, plain
 	// http falls back to a hidden textarea + execCommand)
 	$$('.cp2cb').forEach(el => {
--- a/www/cgi-bin/ext-telegram.cgi
+++ b/www/cgi-bin/ext-telegram.cgi
@@ -76,7 +76,7 @@ fi
 			<h3>Remote send</h3>
 			<dl class="small list mb-0">
 				<dt>Webhook</dt>
-				<dd class="text-break cp2cb">http://root:PASSWORD@<%= $network_address %>/cgi-bin/ext-telegram.cgi?send=image</dd>
+				<dd class="text-break cp2cb"><span class="ep-http">http</span>://root:PASSWORD@<span class="ep-host"><%= $network_address %></span>/cgi-bin/ext-telegram.cgi?send=image</dd>
 			</dl>
 			<p class="small text-secondary mt-2">Call this URL to trigger an image send. Click to copy, then replace <code>PASSWORD</code> with your WebUI password.</p>
 		</div></div>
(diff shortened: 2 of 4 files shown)


widgetii:
Addressed in b0fa4b8 — the RTSP port is now range-checked to 1–65535, not just digit-checked.

Majestic does declare `CND_CFG_RANGE(rtsp, port, 1, 65535)`, but that only binds the write path; a hand-edited `majestic.yaml` reaches `config.json` unchecked, so the review is right that `rtsp://host:0/…` was reachable.

The glob arm now also drops 6-digit and longer values before the numeric test, so `[ -ge ]` never sees an operand too big for a long. Exercised under busybox ash — `554 8554 1 65535` keep their suffix, `0 65536 999999 99999999999999999999 "" abc 12x -1` all fall back to 554.

PR #185, opened 2026-08-26
preview: offer WebRTC alongside MSE, and fall back between them

widgetii:
The Preview page has played MSE over `/ws/video` since it stopped being an MJPEG
refresher, and that is a good default: it negotiates nothing, so it plays
whatever the browser's decoder takes. What it cannot do is get below about a
second of latency, carry audio in the camera's direction, or tell the encoder
that the link is congested. Majestic now serves WebRTC at `/ws/webrtc` on lite
firmware as well, so the page can offer all three.

## Shape

`www/a/preview-webrtc.js` is a second body behind `preview.js`'s façade — the
same seven methods (`setStream`, `requestIdr`, `setAudio`, `setVolume`,
`audioSupported`, `destroy`, `supported`), the same `onState`/`onCodec`/`onAudio`
callbacks. So `preview-page.js` is written once and chooses a transport at
attach time rather than knowing which one it got.

MSE stays the default until WebRTC has had field time. The **WebRTC** toggle
opts in and `localStorage` remembers the choice per browser — what decides it is
what that browser can negotiate, and that travels with the browser rather than
with the camera. `mj-settings.cgi` shares the `preview()` markup but loads
`preview.js` alone, so the toggle stays hidden there.

## The fallback chain is WebRTC → MSE → MJPEG → note

The middle step is load-bearing rather than tidy. WebRTC negotiates and can
therefore fail where MSE cannot: Firefox's WebRTC stack offers only H.264
Baseline whatever its decoder can do, so a camera on `profile: main` has nothing
to give it — while the same browser plays that stream over MSE without
complaint. A player reporting `'fallback'` is asking for the other transport,
not for MJPEG, and the reason lands on the toggle's tooltip rather than
vanishing.

## Verified

Against a hi3516cv500 camera, driving the real page in Chromium (not a mock):

| what | result |
| ---- | ------ |
| H.264 main, 3840x2160 | plays over both transports; the badge names which one |
| Main → Sub over WebRTC | re-negotiates, lands on 1280x720 |
| reload | preference survives; switching back to MSE works |
| audio | unmuting brings a live Opus track and enables the slider; muting again keeps the picture |
| H.265 on both channels | Chromium decodes it in neither transport — the page walks the whole chain to MJPEG within five seconds, unticks the toggle and rewrites the preference |

## One thing worth knowing

With H.265 on `video0` and H.264 on `video1`, WebRTC gives Chromium the
substream, because the camera picks a channel the peer can actually decode. MSE
asks for a stream by number and has no such move, so it drops to MJPEG. The
transport that can fail to negotiate is also the one that can adapt.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -157,6 +157,19 @@ Small `#!/bin/sh` scripts that emit JSON for the front-end. `pulse.cgi` is polle
 - `main.js:initAll` runs on `load`: wires `.btn-danger`/`.btn-warning`/`.confirm` to `confirm()`, links `input[type=range]` to a sibling `…-show` and hidden input, makes external links open in a new tab, and starts the heartbeat.
 - `main.js:runCmd(msg)` streams `/cgi-bin/j/run.cgi` line-by-line via `fetch`/`ReadableStream` and appends to a `pre#output` element whose `data-cmd` carries the base64-encoded command; used by `fw-reset.cgi` (overlay erase).
 - `timezone.js` holds the `TZ` array used by `fw-time.cgi` for the city → `TZ` string mapping.
+- **The live player is two implementations behind one façade.** `preview.js`
+  (`MajesticVideo`, MSE over `/ws/video`) and `preview-webrtc.js`
+  (`MajesticWebRTC`, WebRTC over `/ws/webrtc`) return the same object —
+  `setStream`, `requestIdr`, `setAudio`, `setVolume`, `audioSupported`,
+  `destroy`, `supported` — so `preview-page.js` is written once and picks a
+  transport at attach time. MSE is the default; WebRTC is opt-in via the
+  `#mj-transport` toggle and remembered in `localStorage` under `mj-transport`.
+  The fallback chain is **WebRTC → MSE → MJPEG → note**, and the middle step
+  matters: WebRTC negotiates, so it can fail where MSE cannot (Firefox offers
+  only H.264 Baseline whatever it can decode), which is why a player reporting
+  `'fallback'` asks for the other transport rather than for MJPEG. `mj-settings.cgi`
+  shares the `preview()` markup but loads `preview.js` alone, so the toggle
+  stays hidden there.
 
 ### FPV variant
 
--- a/www/a/preview-page.js
+++ b/www/a/preview-page.js
@@ -101,6 +101,25 @@
 	let stream = 0, audioOn = false, vol = 1;
 	let audioConfigured = false;
 
+	// Which attachment is the live one. Two things need it, and neither is
+	// hypothetical:
+	//
+	// A player can report 'fallback' from inside attach() — MajesticWebRTC does
+	// exactly that when RTCPeerConnection or addTransceiver throws — so the
+	// handler runs, attaches MSE, and then the outer `player = impl.attach(…)`
+	// assignment completes and overwrites the working MSE player with the dead
+	// WebRTC façade nobody can now destroy.
+	//
+	// And a destroyed player is not a silent one. WebRTC's createOffer and
+	// getStats continuations settle after close, so a transport switched away
+	// from can still call onState and, seeing a `usingWebRTC` that has since
+	// flipped, hide a video that is playing perfectly well.
+	//
+	// So the generation is stamped at attach and captured by that attachment's
+	// handlers: a callback from a superseded player is dropped rather than
+	// interpreted as if the current one had sent it.
+	let attachSeq = 0;
+
 	// Whether to offer the audio control at all: the camera has to have audio
 	// configured and this transport has to be able to carry it. MSE can only
 	// when the browser decodes a codec majestic can produce, so the answer
@@ -114,6 +133,7 @@
 	// own different machinery, and the state worth carrying over is three
 	// values.
 	function attachPlayer(webrtc) {
+		const gen = ++attachSeq;
 		if (player) { try { player.destroy(); } catch (e) {} player = null; }
 		usingWebRTC = !!webrtc;
 		// MSE leaves a MediaSource object URL on the element; WebRTC's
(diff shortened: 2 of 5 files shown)


widgetii:
All three fixed in `e98df23`. Each was reproduced against the real page on a hi3516cv500 before fixing and re-checked after — they are the same mistake in three shapes: state shared between an attachment and its replacement, so the loser of a handover keeps a voice.

**1. Fallback overwrites active player — confirmed, and worse than described.** Reproduced by stubbing `RTCPeerConnection` to throw on construction, which is the real synchronous-fallback path:

```
after fallback: badge "H264 3840×2160", playing, toggle off   <- looks fine
after Sub:      badge "MJPEG", video hidden, 3840×2160 frozen  <- controls on the dead façade
```

The picture kept playing, so the only symptom was that the controls did nothing — and then clicking Sub drove the dead WebRTC façade, which reopened, failed, and took the whole page to MJPEG.

**2. Destroyed player controls UI — confirmed.** My first attempt at this raced and passed, so I forced the window instead: delay `createOffer` by 2.5 s, switch transports at 1 s. The rejection then arrives against a player that no longer exists.

```
t+ 4s badge "MJPEG", vidShown false, currentTime  1.89
t+16s badge "MJPEG", vidShown false, currentTime 13.95
```

MSE was playing the whole time — `currentTime` climbing steadily behind an MJPEG overlay the user had no way to dismiss.

Both are fixed by stamping each attachment with a generation. Handlers capture it and drop callbacks that are not current, and `attachPlayer()` destroys what it built rather than assigning it when a nested attach has superseded it.

**3. Reconnect loses signaling socket — confirmed, and it leaks a camera session.** `RTCPeerConnection.close()` is specified not to fire state-change events, so the trigger has to be the event itself. Driving the page's own ICE handler with `iceConnectionState === 'failed'`:

| | sockets created | still open |
| --- | --- | --- |
| before | 2 | **2** |
| after | 2 | 1 |

The orphan is not just a dangling reference. majestic's log shows the camera holding both sessions at once — client `0x7f7c1070` still connected while its replacement `0x7f783488` ran, for twenty-two seconds until the tab closed. `RTC_MAX_SESSIONS` is 3, so two ICE blips on a single tab are enough to lock other viewers out. That is a stronger consequence than the finding claimed.

Fixed by capturing every socket in its own handlers so it only clears the shared reference while it still holds it, closing the socket being replaced in `reconnect()` and `stop()`, and numbering each connection attempt so a continuation from a retired one cannot report a failure the live attempt is not having.

**Re-verified after the fixes**, all against the camera in Chromium: the two regressions above now pass; 4K H.264 over both transports; Main→Sub renegotiation; preference surviving a reload; switching back to MSE; a live Opus track on unmute and the picture surviving re-muting; and WebRTC recovering from the forced ICE failure with one socket instead of two.

widgetii:
**Hung sessions skip fallback** — real, and fixed in `75c21bc`.

The diagnosis is exactly right: nothing else in the player could ever notice. The socket is open, so no `onclose`; ICE is connected, so no failure; the offer was answered. The deadline said "no signal" and stopped, and since the page treats that as a display state rather than a request to change transport, it stayed on the bars for as long as the tab was open — with MSE, which asks for none of this, sitting unused behind a toggle.

The deadline now retires the attempt as well as reporting it, and the escalation already there does the rest: three fruitless attempts and the page falls back.

Reproduced by stubbing `getStats` to report zero `bytesReceived`, which is what a dropped media path looks like from up here. Worth noting for anyone repeating it: `RTCStats` entries are read-only, so the stub has to hand back copies — my first attempt edited them in place, failed silently, and "passed".

```
before: badge "no signal", 60 s and counting, never tried MSE
after:  "no signal" at 9 s, again at 18 s, then MSE at 4K,
        toggle unticked, tooltip "WebRTC: negotiated but no media arrived"
```

Which failure it was now travels with the message, because "negotiated but no media arrived" and "could not establish a session" send whoever reads that tooltip to different halves of the interop notes. That also fixed a stutter I had not spotted: the page renders `'WebRTC: ' + reason`, so reasons that named the transport again came out as "WebRTC: WebRTC …".

Full suite re-run against the camera after this change — happy path, both earlier regressions, and the forced ICE failure — all green.

widgetii:
`03cceaf` — two of the three addressed, one declined with a measurement.

**Missing audio stays listening — real, fixed.** `preview.js` reports audio on every stream init (`onAudio(wantAudio ? (info.audioCodec || null) : null)`) precisely so the page can reset a control that promises something the camera is not delivering, and my player only ever reported a track it received. The answer's completion now decides it: `ontrack` has already run for everything that answer accepted, so an audio track absent at that point is not coming. It also mirrors `preview.js` in re-muting and clearing `wantAudio`, so a later unmute actually flips and renegotiates instead of no-opping.

Reproduced by turning the camera's audio off and asking for it anyway:

```
before: 🔊 Listening, 0 audio tracks, slider enabled, still there at t+16s
after:  🔇 No audio,  0 audio tracks, slider disabled, video untouched at 4K
```

**Audio play loses activation — partly.** The premise does not hold: `preview.js` has the identical `setAudio` (unmute, then `reopen()` with the same 300 ms delay) and the identical `video.play().catch(function () {})`, and documents no requirement about activation — so this is not something the new player does differently. Chrome's sticky activation also survives the delay; the measured audio test plays unmuted after the toggle click.

What *is* a real hole is swallowing the rejection: if unmuted playback ever is refused, the old code left a paused video and no explanation, which is the worst outcome available. Now it takes the picture over the sound — re-mute, play, and report `onAudio(null)` so the control stops lying.

**Early ICE candidates discarded — declined, measured.** The mechanism is right: `addIceCandidate()` is called without waiting for `setRemoteDescription()` to resolve, and the camera does trickle candidates immediately behind the answer. But nothing is lost, because both chain onto the same `RTCPeerConnection` operations queue — one queued behind a pending answer runs after it, per spec.

```
camera sent 3; added 3, rejected 0, handed over pre-answer 1
```

One candidate really did arrive before the answer was installed, and it was still added. A queue of my own would have guarded against nothing, so there is a comment recording the measurement where the queue would have gone.

---

**One caveat on verification.** The lab camera went down mid-session — majestic exited silently during a SIGHUP config reload (widgetii/majestic#430, which this reproduces: log ends at "Signal HUP received, reloading config" with nothing after it), and the board is now unreachable and needs a power cycle. Everything above was measured before that, including this commit's final code shape. What has *not* been re-run against hardware since this last edit is the wider suite — happy path, the two player-lifetime regressions, the forced ICE failure, the no-media fallback — all of which passed one commit earlier and are untouched by this change apart from the new `play()` rejection handler.

widgetii:
`38ea511` — camera is back, full suite re-run against the final tree, and it caught two mistakes of my own in the previous commit. Both reported "no audio" over a working Opus stream, which is the failure mode that commit existed to prevent.

**An ordering assumption.** I decided whether the camera had given us audio by checking a flag `ontrack` sets, from inside `setRemoteDescription`'s `.then()`, believing the track events had already fired. They are queued tasks and so is the promise resolution; nothing orders them. Now it asks the transceivers — `currentDirection` is the negotiated answer and is settled by the time that handler runs.

**The play() rejection handler I added for the autoplay finding.** It treated every rejection as autoplay policy refusing sound. But `ontrack` fires once per track, so the audio track's call reassigns `srcObject` and aborts the video track's `play()` with `AbortError` — an ordinary event on every session that has audio. It muted the element and reported null. Only `NotAllowedError` means the sound is the problem.

How they surfaced is worth recording, because it is the argument for re-running everything rather than the last thing touched: **the camera-has-no-audio test passed throughout.** When there is genuinely no audio, a false negative and a true negative look identical. Only the test that unmutes a camera that *does* have audio could tell them apart.

Final state, all measured against the camera:

| | |
| --- | --- |
| happy path (4K both transports, Main→Sub, reload, back to MSE) | pass |
| fallback buries the live player | pass |
| late callback from a destroyed transport | pass |
| forced ICE failure — sockets left open | 1, was 2 |
| no-media session falls through to MSE | pass, reason "negotiated but no media arrived" |
| trickled candidates reaching ICE | 3 of 3 |
| audio on unmute / control resets when the camera has none | pass |

---

**Unrelated, but reproduced twice while doing this and worth reporting:** majestic exited silently on `GET /api/v1/set?audio.enabled=…` while a native WebRTC session was live. Both times the log ends at `Signal HUP received, reloading config` with nothing after it, and the process is simply gone. That is widgetii/majestic#430, which has been unexplained — this looks like a repeatable recipe for it.

widgetii:
**Post-start stalls never recover** — real, fixed in `bf38703`, and the fix needed a second pass.

The diagnosis is exact: `bytesReceived` only ever climbs, so "greater than zero" said a session had once worked, not that it still did. After the first bytes the no-signal watchdog was disarmed and nothing replaced it. This matters more here than on the MSE path, where media rides the signalling socket — a camera that stops sending closes it or errors the element, and either reconnects. WebRTC's media is out-of-band, so signalling notices nothing.

The stats poll now watches progress rather than totals, and a stall retires the attempt exactly as a failure does.

**The first version of that fix detected every stall and still never fell back**, which the test caught: `failCount` was cleared on the first byte of *any* attempt, so a session that started and died over and over reset the count each time and retried for ever — the one outcome the escalation exists to prevent. Failures are now forgiven only once a session has held up for as long as it would take to call it stalled.

Reproduced by letting a session start properly, then freezing `bytesReceived` — what the page sees when the camera goes away without ICE noticing:

```
before: badge "H264 3840×2160 · WebRTC", frozen, indefinitely
after:  three attempts over ~30 s, then MSE at 4K,
        tooltip "WebRTC: media stopped arriving"
```

Worth noting this is not hypothetical on this hardware: majestic exiting under a SIGHUP reload (widgetii/majestic#430) produces precisely this shape, and I walked into it twice today from the other side.

**Whole suite re-run against this tree**, all against the camera:

| | |
| --- | --- |
| happy path — 4K over both transports, Main→Sub, reload, back to MSE | pass |
| fallback buries the live player | pass |
| late callback from a destroyed transport | pass |
| forced ICE failure — sockets left open | 1 |
| no-media session falls through | pass |
| post-start stall falls through | pass |
| trickled candidates reaching ICE | 3 of 3 |
| audio on unmute | pass |
| camera with audio off resets the control | pass |

widgetii:
**Muted fallback keeps audio streaming** — real, fixed in `036716b`.

It also contradicts this file's own reasoning a few lines up, where the audio transceiver is added only when `wantAudio` is set, *"because negotiating a track nobody listens to has the camera encode for nobody"*. A track the browser has declined to play is nobody. The branch now renegotiates via `setAudio(false)` rather than just flipping the local flag.

Reproduced by rejecting `play()` with `NotAllowedError` whenever the element is unmuted, which is what a browser that has not granted the page sound does:

```
before: 1 audio track still arriving 20 s after the refusal
after:  0, picture still playing, control reset to "No audio"
```

**Whole suite re-run against this tree** — nine cases, every one against the camera, each of the fixed findings verified failing before and passing after:

| | |
| --- | --- |
| happy path — 4K over both transports, Main→Sub, reload, back to MSE | pass |
| fallback buries the live player | pass |
| late callback from a destroyed transport | pass |
| forced ICE failure — sockets left open | 1 |
| no-media session falls through | pass |
| post-start stall falls through | pass |
| autoplay refusal renegotiates the audio away | pass |
| trickled candidates reaching ICE | 3 of 3 |
| audio on unmute / control resets when the camera has none | pass |

One finding stands declined: **early ICE candidates**. Measured — three trickled, one handed over before the answer was installed, all three added, none rejected, because `addIceCandidate` and `setRemoteDescription` share the same operations queue. There is a comment recording that where a queue of my own would have gone.

widgetii:
The one finding still listed open — *Muted fallback keeps audio streaming* — is quoting the previous pass's code:

```js
video.muted = true;
wantAudio = false;
video.play().catch(function () {});
```

Those three lines were replaced in `036716b`; the branch now calls `setAudio(false)`, which mutes and reopens, so the session comes back without the audio m-line. Measured: one audio track still arriving twenty seconds after the refusal before, none after.

Merging.

PR #186, opened 2026-08-26
preview: make WebRTC the default, and be careful what that remembers

widgetii:
WebRTC has earned the default: sub-second where MSE is about a second behind, audio in both directions, and a receiver whose reports let the camera match the encoder to the link. MSE is one tick away for the browsers and cameras where negotiation cannot be made to work, and the WebRTC → MSE → MJPEG chain is unchanged.

## The part that actually needed thought

Not the default — what a failure is allowed to conclude from it.

Opt-in, a demotion was harmless: the person had asked for WebRTC and could ask again. As the default it runs for everybody on every load, and a single stored "use MSE" would quietly park a browser on the slower transport with nothing on screen to say why. So the memory is split in two:

| key | what it is | how long |
| --- | --- | --- |
| `mj-transport` | what the person chose, either direction | permanent, and beats everything below |
| `mj-transport-auto` | what a failure decided for them, timestamped | 6 h, then discarded |

`mj-transport-auto` is never written against an explicit choice of WebRTC — someone who ticked the box gets it back rather than being overruled in silence. And anything unparseable in it counts as expired, so a value from another version cannot pin the transport for ever.

A camera that is merely **full** is not a failure at all. With widgetii/majestic#456 it says so — `busy` rather than `error` — and that is not remembered, because it will not be true for long. Without it a third viewer arriving would demote the fourth for six hours over a slot that freed a minute later. An older majestic sends `error` for both, and that is exactly what happens; the rest of this still works.

## Verified

On a hi3516cv500, on top of the nine cases already in the suite (all re-run green):

- an unprimed browser lands on WebRTC and stores nothing;
- unticking stores `mse`, survives a reload, re-ticking stores `webrtc`;
- a browser where `RTCPeerConnection` throws is demoted — under the auto key, not as a choice — and takes MSE;
- a six-hour-old demotion is discarded and WebRTC tried again;
- four browsers at once: three get WebRTC, the fourth is told the camera is full and takes MSE **remembering nothing**, and has WebRTC back once a slot frees.

```
viewer 4  H264 3840x2160        <- MSE, auto=null, pref=null
          "WebRTC: the camera is already serving as many WebRTC viewers as it
           can; try again shortly"
after a slot frees: H264 3840x2160 · WebRTC
```

## Worth knowing

A WebRTC viewer is in the camera's bitrate loop — REMB drives the encoder — where an MSE viewer never was. That is now the default on the Preview page. `mj-settings.cgi` still loads `preview.js` alone, so the settings preview, where someone is judging quality, is unaffected.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -164,11 +164,14 @@ Small `#!/bin/sh` scripts that emit JSON for the front-end. `pulse.cgi` is polle
   `destroy`, `supported` — so `preview-page.js` is written once and picks a
   transport at attach time. **WebRTC is the default**; the `#mj-transport`
   toggle switches to MSE. What is remembered is split in two on purpose:
-  `mj-transport` is the person's explicit choice and is permanent, while
+  `mj-transport-pick` is the person's explicit choice and is permanent, while
   `mj-transport-auto` is a demotion a failure decided for them and carries a
   timestamp so it expires (6 h) — otherwise one bad session parks a browser on
   the slower transport for good. A camera that is merely out of session slots
   answers `busy` rather than `error`, and that is not remembered at all.
+  `mj-transport` is the previous release's single key: read once, migrated and
+  deleted, because it could not tell a choice from a fallback (both wrote
+  `mse`).
   The fallback chain is **WebRTC → MSE → MJPEG → note**, and the middle step
   matters: WebRTC negotiates, so it can fail where MSE cannot (Firefox offers
   only H.264 Baseline whatever it can decode), which is why a player reporting
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -162,8 +162,13 @@ Small `#!/bin/sh` scripts that emit JSON for the front-end. `pulse.cgi` is polle
   (`MajesticWebRTC`, WebRTC over `/ws/webrtc`) return the same object —
   `setStream`, `requestIdr`, `setAudio`, `setVolume`, `audioSupported`,
   `destroy`, `supported` — so `preview-page.js` is written once and picks a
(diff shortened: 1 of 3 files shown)


widgetii:
`ea6aa50` — both real, and the first was a migration I would have shipped wrong.

**Legacy fallbacks become permanent.** Exactly right: the release merged this morning kept one key and its *automatic* fallback wrote `mse` into it — the same value unticking the box wrote. So every browser that has already fallen back is indistinguishable from someone who chose MSE on purpose, and reading it as a choice pins them off the new default for good.

The explicit choice moved to `mj-transport-pick`; `mj-transport` is now read once and discarded. `webrtc` under the old key was unambiguous — only the toggle wrote it — so it survives as a choice. `mse` becomes a demotion: nothing changes for that browser today, and in six hours it tries WebRTC again instead of never.

**Future timestamps never expire.** Also right, and `9999999999999junk` is a good example — `parseInt` reads the digits happily. The window is bounded at both ends now and the value has to be all digits.

Measured against the camera:

```
legacy mse    -> auto=<now>, pick=null, old key gone, MSE today
legacy webrtc -> pick=webrtc, auto=null, WebRTC now
auto=now+9e9  -> discarded, WebRTC
auto='9999999999999junk' -> discarded, WebRTC
```

---

**One thing that should hold this PR, found while testing it.** Concurrent WebRTC session churn was segfaulting majestic — widgetii/majestic#430, root-caused and fixed in widgetii/majestic#458 (a use-after-free in the signalling handler). Making WebRTC the default is exactly what multiplies concurrent sessions, so this should not merge until that fix has shipped. The camera-side companion (widgetii/majestic#456, the `busy` reply) is independent and safe either way.

PR #190, opened 2026-08-27
settings: keep field positions still when rows appear and disappear

### LINKED ISSUE #189: Keep form field positions stable when dynamic fields appear or disappear
When changing the value of a select field causes other fields to appear or disappear dynamically, the current two-column layout can rearrange the remaining fields.

For example, fields that were at the bottom of the first column may move to the top of the second column, or vice versa, causing the entire second column to shift. In some cases, the field currently being edited moves to a different position as a result of changing its value. This is especially inconvenient because the user can suddenly lose sight of the control they are working with.

I think the positions of existing fields should remain stable when other fields are dynamically shown or hidden. Adding or removing a field should not cause the other fields to be redistributed between the two columns.

It may make sense to recalculate the column layout when the browser window is resized. 

The main goal is to keep the form predictable and prevent the control currently being edited from unexpectedly moving.

openipc-ai:
Fixes #189.

### The bug

`.mj-cols` was a CSS multi-column box (`column-count: 2`). A column box re-balances itself whenever its content changes height, so every `visibleWhen` row that appeared or disappeared re-flowed the entire section — including rows nowhere near the control that was touched.

Measured on the lab hi3516av300 (headless Chromium, `osd.anchor` changed from `proportional`), `x` is the row's left edge, so a change in `x` means it crossed into the other column:

| width | new anchor | rows that jumped columns | the select being edited |
|---|---|---|---|
| 1440 | `center` | `osd.outline` (x 419 → 905) | moved down 87px |
| 1200 | `center` | `osd.outline` (x 344 → 763) | moved down 87px |
| 992 | `center` | `osd.outline` (x 285 → 636) | moved down 108px |
| 900 | `top-left` | — | **anchor itself, x 560 → 299, y 295 → 922** |
| 900 | `bottom-right` | — | **anchor itself, x 560 → 299, y 295 → 922** |

The last two are the case the issue calls out: pick a value and the dropdown you just used leaves the column, so the cursor is now pointing at whatever slid into its place.

### The fix

Deal the rows into two *real* columns rather than let a column box flow them.

- `.mj-cols` is a flex row of two `.mj-col` children (`flex: 1 1 0; min-width: 0`), so both are the same width whatever they hold and long labels wrap inside their column.
- `layoutCols()` decides the cut — at mount and on a debounced `resize`, **never** when a row toggles. Showing or hiding a row then only moves what is under it in its own column.
- `rowSteps()` measures a row as the step down to the next *visible* row, so the margins between rows count and hidden rows cost nothing.
- The cut is the one that leaves the taller column shortest, and never falls immediately after an `<h5>` group heading (a heading belongs to the rows under it).
- Re-dealing re-parents nodes, which would blur a control mid-edit, so `grabFocus`/`restoreFocus` carry focus **and** text selection across a resize.
- Below `md` the columns stack and the deal is skipped — stacked, every split reads the same, and the rows stay in document order.

### Verification

All against the lab hi3516av300 with the branch deployed to `/var/www/a`, driven by headless Chromium through an authenticating pass-through proxy (the camera itself untouched).

**Stability** — every case above, re-run:

| tab | change | jumps | edited control |
|---|---|---|---|
| osd | anchor → `center` | 0 | stayed put |
| osd | anchor → `top-left` | 0 | stayed put |
| osd | anchor → `bottom-right` | 0 | stayed put |
| isp | `iris.type` → `DC` (reveals 8 rows) | 0 | stayed put |

at 1440 / 1200 / 992 / 900 px.

**Resize still re-balances, and does not eat the caret** — caret parked at chars 3–7 of `osd.template`, four rows dropped from the form, then a resize:

```
mounted       6/7  h=522/559
after toggle  6/7  h=522/350   focus=mjf-osd-template     <- split held
after resize  5/8  h=435/438   focus=mjf-osd-template sel=3-7
```

**No regression to the initial layout** — all 19 reachable sections rendered at 1440px and 390px: no console errors, no horizontal overflow, no `.mj-ctl-in` spill, and column heights within ~10 % of each other on every multi-row section (e.g. `video0` 717/788, `sip` 911/837, `osd` 522/559). At 390px every section is a single stacked column, as before.

### Known trade

Holding the split still means a big reveal is no longer re-balanced: setting `isp.iris.type` to `DC` reveals eight rows into one column and leaves it at 525 vs 1261 px until the next resize or navigation. That is the trade the issue asks for — the whole `Iris` group staying together under its own heading reads fine, and predictable beats balanced when the control under your cursor is at stake.

https://claude.ai/code/session_01UKXGGYZaF1pxsZTxqhEz7u


openipc-ai:
Addressed in b659316 — the finding is valid in mechanism, so I removed the dependency rather than lean on the fact that it currently cancels.

**Verifying it first.** `rowSteps()` measured each row as the distance to the next row *in its own column*, so the last row of each column was measured to that column's bottom instead of to the next document-order row. Those costs were then reused to score every candidate cut. To see whether that actually changed a result, I forced four different starting splits (all rows left, all rows right, 1/rest, quarter/rest), kicked a resize after each, and compared the cut against the one a fresh mount picks:

```
osd     mounted cut=6   from 13 -> 6   from 0 -> 6   from 1 -> 6   from 3 -> 6
isp     mounted cut=6   from 21 -> 6   from 0 -> 6   from 1 -> 6   from 5 -> 6
video0  mounted cut=8   from 15 -> 8   from 0 -> 8   from 1 -> 8   from 3 -> 8
sip     mounted cut=9   from 17 -> 9   from 0 -> 9   from 1 -> 9   from 4 -> 9
system  mounted cut=4   from  9 -> 4   from 0 -> 4   from 1 -> 4   from 2 -> 4
```

It converges everywhere, because `p.mj-row` is uniformly `margin-top: 0; margin-bottom: 1rem` — "to the next row's top" and "to my column's bottom" are the same number. It stops cancelling for a row that *precedes* an `<h5>` group heading, whose `mt-4` collapses in only when the two share a column: 8px, latent until a schema grows more nested-object groups than today's single `isp.iris`.

**The fix.** `rowBoxes()` now returns each visible row's height and its own margins, read from its styles; `y[]` lays them out as if they all ran down one column; and each candidate's two column heights are read off as differences — the left column ending at its last row's bottom margin, the right starting with its first row's top margin whole, since a flex item is its own block formatting context and cannot collapse either away. Nothing consults the current split.

**Checked against the browser, not by argument.** For the cut actually chosen, the modelled heights match what the browser rendered, across all 19 sections at 1440px and 992px:

```
system   cut=4  model=368/397  rendered=368/397   isp    cut=6  model=525/489  rendered=525/489
video0   cut=8  model=716/787  rendered=717/788   osd    cut=6  model=521/558  rendered=522/559
rtsp     cut=6  model=698/890  rendered=699/891   sip    cut=9  model=910/836  rendered=911/837
```

(≤1px, from `offsetHeight` being integral.) The rest of the suite re-ran clean on the lab hi3516av300 with this commit deployed: 0 column jumps and a stationary edited control for `osd.anchor` → center/top-left/bottom-right and `isp.iris.type` → DC at 1440/992/900px; resize still re-balances 6/7 → 5/8 while keeping `sel=3-7` in a focused text field; no console errors, no overflow, no `.mj-ctl-in` spill, and identical splits to the previous commit on all 19 sections at 1440 and 390px.

Issue #184, opened 2026-08-25
Preview source and encoding for slow communication channels


usa-:
This proposal concerns majestic-webui and majestic.

I have around a dozen cameras deployed at various remote locations. They stream video to YouTube over connections with relatively high bandwidth, while camera management is performed over a separate VPN connection with much lower bandwidth.

In this setup, the camera preview frequently freezes or develops artifacts in both of these UI locations:

* `Preview`
* `Majestic → Settings → Live Adjustments`

I would like to propose a possible solution to this problem, but first I would like to discuss the concept.

I suggest adding a new `Preview` section under `Majestic → Settings → Video & Audio`, where the video source and encoding method used for the preview can be configured.

### Source

Possible values:

**Video**

Use one of the enabled, uncropped sources: `Main stream` or `Sub stream`.

Among the available sources with a resolution at least as high as required by the particular preview location, select the one with the lowest resolution. If all available sources have a lower resolution than required, use the one with the highest resolution.

**MJPEG**

Offer this option only on SoCs where Majestic supports a separate hardware MJPEG channel.

**JPEG**

This is the most flexible mode and the one I intend to use.

Select the actual source according to the same principle as for `Source = Video`, with one difference: if a separate hardware MJPEG pipeline is available on the SoC, consider it as a candidate as well.

The following settings are available in this mode.

### **Reduce size**

`Yes / No`

Determines whether the source frame should be reduced to the resolution required by the particular preview location using CPU resources.

### **Maximum size**
*`WxH` format, empty = no limit.*

Optional text field. If `Reduce size = Yes` and the frame size required for the preview exceeds this value, reduce the source frame directly to this size. If `Reduce size = No` and the source frame size exceeds this value, reduce it to the specified dimensions using CPU resources.

### **Frame rate**

Integer field followed by a unit selector with:

* `frames per second`
* `seconds per frame`

### **Enable OSD**

Same as the existing setting:

`Majestic → Settings → Video & Audio → MJPEG → Enable OSD`

widgetii:
Your diagnosis matches what we found independently, and some of this may already be solved for you — though not all of it, and the part that isn't is worth separating out.

[...]

PR #191, opened 2026-08-27
preview: say that WebRTC lets the camera change the stream for everyone

widgetii:
Raised by @usa- in #184, and it is a fair hit.

The toggle said `WebRTC` and nothing else. Switching it on puts the viewer in the camera's bitrate loop: the receiver's estimate reconfigures the encoder for whichever channel it is watching. On a thin link that *is* the feature — it is why the preview holds up where MSE tears — but the encoder is shared, so the change reaches every other viewer of that stream and whatever is recording it.

It is bounded and temporary. The configured `videoN.bitrate` is a ceiling, so a viewer can only ever lower it, and the encoder is handed back when the last peer disconnects. But temporary is not the same as disclosed, and deciding it on someone else's behalf from a button labelled with a protocol name is not a fair trade.

## What changes

- A **tooltip** on the toggle: what the transport buys, and what it costs.
- A **line under the control**, shown only while WebRTC is the live transport:
  > The camera is adapting this stream's bitrate to your connection. Anyone else watching the same stream sees that too.

The tooltip alone would not do — it needs a pointer to find, and this is not a detail reserved for people who own a mouse.

The failure path already borrowed that tooltip to explain why WebRTC gave up, so the two now share it: the reason first, because it is the news, then the standing explanation. Otherwise the next person to hover learns only that something once broke.

```
normal:        "Sub-second video and two-way audio, and the camera fits the
                stream to your connection — which changes it for everyone else
                watching that stream too."

after failure: "WebRTC: negotiated but no media arrived
                <blank>
                Sub-second video and two-way audio, and the camera fits …"
```

## Verified

On a hi3516cv500, driving the real page:

| | |
| --- | --- |
| default (WebRTC) | note shown, tooltip carries the explanation |
| switched to MSE | note hidden |
| back to WebRTC | note returns, tooltip unchanged |
| forced failure (`RTCPeerConnection` throws) | note hidden, tooltip has reason **and** explanation |

Existing preview suite re-run and still passing: happy path, both player-lifetime regressions, no-media fallback, and the transport-preference cases.

## Not in this PR

@usa- also asked for an explicit **"allow adjusting bitrate: yes/no"**, which is the right answer for a camera whose streams are already committed to other uses. That needs camera-side work and a decision about where the knob lives, so it is separate. This change only stops the side effect being a surprise.

Refs #184

### DIFF
--- a/www/a/preview-page.js
+++ b/www/a/preview-page.js
@@ -88,6 +88,28 @@
 	const mute = $('#mj-mute'), muteLbl = $('#mj-mute-lbl'), volCtl = $('#mj-vol');
 	const audioCtl = $('#mj-audio-ctl');
 	const transportCtl = $('#mj-transport'), transportGrp = $('#mj-transport-ctl');
+	const transportLbl = $('#mj-transport-lbl'), transportNote = $('#mj-transport-note');
+
+	// What the toggle says when nothing has gone wrong. Kept because a failure
+	// replaces it with the reason, and switching back has to put the explanation
+	// there again rather than leave the tooltip stuck on a complaint about a
+	// session that is long over.
+	const TRANSPORT_TITLE = transportLbl ? transportLbl.title : '';
+
+	// The bitrate adaptation is the thing worth disclosing. It is what makes
+	// WebRTC work on a thin link, and it reaches past the person who switched it
+	// on: the encoder is shared with everyone else watching that stream, and
+	// with whatever is recording it. So say it where it cannot be missed, and
+	// only while it is actually happening — a tooltip needs a pointer to find,
+	// and this is not a detail for people who happen to own a mouse.
+	function syncTransportNote() {
+		if (transportNote) {
+			transportNote.hidden = !usingWebRTC;
+		}
+		if (transportLbl && usingWebRTC) {
+			transportLbl.title = TRANSPORT_TITLE;
+		}
+	}
 	const s0 = $('#mj-stream-0'), s1 = $('#mj-stream-1');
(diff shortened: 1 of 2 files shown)


widgetii:
Declining this one — the behaviour is deliberate, and I have pinned it with a test rather than leaving it to argument.

After a fallback the toggle is unticked and the note is hidden, so the page already says "you are not on WebRTC". What the tooltip then answers is the obvious next question: *why not*. Clearing it would leave a control that is off for a reason nobody can discover. That is why the reason and the standing explanation share the tooltip rather than replacing one another.

The reading that would make this a bug is "the MSE control reports a failure that never happened", and that does not occur. Choosing MSE deliberately never writes a failure title:

```
1. default (WebRTC)      title = "Sub-second video and two-way audio, and the camera fits …"
2. switched to MSE       title = "Sub-second video and two-way audio, and the camera fits …"   <- unchanged
3. back to WebRTC        title = "Sub-second video and two-way audio, and the camera fits …"   <- restored
4. forced failure        title = "WebRTC: RTCPeerConnection failed\n\nSub-second video and …"
```

Step 2 now asserts both that the tooltip is untouched and that it does not start with `WebRTC: `, so a future change that leaks a failure message onto a control the user chose will fail the test instead of shipping.

The window in which the reason persists is bounded the way it should be: a successful WebRTC attach restores the plain title (step 3, via `syncTransportNote()`), and a page load starts from the markup. Both are covered above.

Verified on a hi3516cv500 along with the rest of the preview suite.

PR #192, opened 2026-08-27
preview: start on the substream, the way this equipment is meant to be used

widgetii:
CCTV has settled this convention already: the **main channel** carries the best picture the sensor can give and is what an NVR or the SD card records; the **substream** exists to be watched over whatever link happens to be available. Previewing the channel that is being recorded — and then adapting its bitrate down to suit one browser on a hotel wifi — is the wrong way round.

So the preview starts on Sub where one exists. Main is one click away for anyone who wants a closer look, and the note from #191 says what that costs while they are there.

## Why this is better than the switch it replaces

#184 asked for an explicit *"allow adjusting bitrate: yes/no"*, on the grounds that a stream already committed to another purpose should not be quietly re-tuned to suit a preview. That is a fair objection, and this answers most of it structurally rather than with another setting: the adaptation now lands on the channel whose *job* is preview, so the stream someone is recording is untouched unless they ask for it by name.

What remains is the narrower case of a substream already committed to something else, which is a camera-side question rather than a page one.

## Gated, not assumed

The default is conditional on `video1.enabled`, because the two transports disagree about what a missing stream means:

- **WebRTC** treats `?stream=N` as a preference and falls back to a channel it can actually serve.
- **`/ws/video`** subscribes to whatever number it is handed and then delivers nothing — which reads as "no signal" rather than as a misconfiguration.

That is also why the first attach now waits on the config fetch. It costs nothing: `mjConfig()` is memoised and the night-mode controls have already asked for it by the time the player section runs. The three controls that reach the player are guarded for the gap.

## Verified

On a hi3516cv500, over both transports:

| camera | result |
| --- | --- |
| `video1.enabled` | both start on Sub at 1280x720; Main switches to 3840x2160 |
| `video1` disabled | Sub control hidden, both start on Main, nothing stalls |

The rest of the preview suite passes. Two of its tests were **corrected rather than worked around**: both proved a control was live by clicking Sub, which is now where the page already is — they would have passed whether the control worked or not. They click Main instead, which restores what they were actually testing.

Refs #184

### DIFF
--- a/www/a/preview-page.js
+++ b/www/a/preview-page.js
@@ -330,19 +330,52 @@
 		};
 	}
 
-	attachPlayer(wantWebRTC());
+	// Start on the substream, which is the convention this equipment is built
+	// around: the main channel carries the best picture the sensor can give and
+	// is what an NVR or the SD card records, while the substream exists to be
+	// watched over whatever link happens to be available. Previewing the
+	// channel that is being recorded, and then adapting its bitrate down to
+	// suit one browser, is the wrong way round.
+	//
+	// It also settles most of what a bitrate opt-out would have been for: the
+	// adaptation now lands on the channel whose job is preview. Main is one
+	// click away for anyone who wants to look closely, and the note under the
+	// toggle says what that costs.
+	//
+	// Gated on video1.enabled rather than assumed, because the two transports
+	// disagree about what a missing stream means. WebRTC treats ?stream=N as a
+	// preference and falls back to a channel it can serve; /ws/video subscribes
+	// to whatever number it is given and simply delivers nothing, which reads
+	// as "no signal" rather than as a misconfiguration.
+	mjConfig().then(cfg => {
+		if (mjGet(cfg, 'video1.enabled') === true) {
+			stream = 1;
+			if (s1) s1.checked = true;
+		}
+		attachPlayer(wantWebRTC());
+	});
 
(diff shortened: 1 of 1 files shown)


widgetii:
Fixed in `335b113` — the finding is right, and it was the wrong trade twice over. `/api/v1/config.json` has no timeout of its own, and it would hang on exactly the thin, flaky link this default exists for.

The attach now races a 1500 ms deadline. In practice the answer is already there — `mjConfig()` is memoised and the night-mode controls asked for it before this code runs — so the race resolves immediately with the right channel. When it does not, the preview starts on Main and corrects itself if the answer ever turns up. A picture on the wrong channel beats no picture.

**The guard against overruling a person needed a second attempt**, and my own test caught it. Arming it on the `change` event was wrong: clicking the radio that is *already selected* fires no `change`, so someone who starts blind on Main and presses Main — because Main is the one they want — looked exactly like someone who pressed nothing, and the late answer moved them to Sub anyway. It is armed on `click` now, since a click is an expression of intent whether or not it alters anything.

Verified with the config fetch stubbed at the browser:

```
never answers                        starts on Main, plays, stays there
answers after 5 s                    starts on Main, corrects itself to Sub
answers after 6 s, Main clicked at 2.5 s   stays on Main
```

Fourteen cases across the preview suite still pass.

PR #193, opened 2026-08-27
settings: give Live adjustments the low-latency transport, on the substream

widgetii:
This page is where latency is the feature. Someone is dragging a saturation slider and watching for the effect, and MSE is about a second behind where WebRTC is not.

**The reasoning that kept it on MSE was mine and it was wrong.** I argued that a WebRTC viewer joins the encoder's bitrate loop and would disturb a judgement about image quality. That does not survive looking at what the page actually offers: Brightness, Contrast, Saturation, Hue, Mirror and Flip are all ISP knobs. Nothing here judges an encoder setting, and whoever is tuning `videoN.bitrate` is on another tab with no preview at all. The objection applied to nobody, while the cost — a second of lag on a control being dragged — landed on everybody.

It takes **the substream** too, by the convention #192 established: the main channel is what an NVR or the SD card records, and this preview only has to show what the ISP is doing, which looks the same on either channel because both are the same picture scaled. Main where no substream is configured, because `/ws/video` subscribes to whatever number it is handed and then quietly delivers nothing.

## One extraction, deliberately narrow

`preview-transport.js` now owns **which transport to prefer and what to remember about it**. Two pages need those rules and they are subtle enough that two copies would drift within a release:

- a legacy key migrated exactly once, distinguishing a choice from a fallback;
- a demotion that expires after six hours;
- a window bounded at *both* ends, so a clock running backwards cannot suppress WebRTC for a week;
- the difference between a camera that cannot serve this browser and one that is merely busy.

Divergence there would present as *"the preview behaves differently on the settings page"*, which is the kind of bug nobody files.

**The attach-and-fall-back dance stays in both callers, on purpose.** What the Preview page does on a fallback — badge, MJPEG, keeping a toggle and a note in step — has nothing to do with this panel, and a shared version would be an abstraction over two callers with one of them bent to fit. It is a dozen lines either side; the rules above are not.

## Verified

On a hi3516cv500:

| | |
| --- | --- |
| Live adjustments | plays 1280x720 over WebRTC (`srcObject` set) |
| `RTCPeerConnection` made to throw | same panel, same substream, over MSE |

Sixteen cases across the preview suite still pass — which is what the extraction had to earn, since it touched the Preview page's preference handling as well.

Refs #184

### DIFF
--- a/www/a/mj-settings.js
+++ b/www/a/mj-settings.js
@@ -677,25 +677,48 @@
 	function attachLivePreview(video) {
 		const stream = getDotted(state.config, 'video1.enabled') === true ? 1 : 0;
 
+		// Which attachment is the live one. MajesticWebRTC can report 'fallback'
+		// from inside attach() — it does exactly that when RTCPeerConnection or
+		// addTransceiver throws — so the handler below runs, installs MSE, and
+		// then the outer assignment completes and buries it under the façade
+		// that just failed. The picture would look right and the handle would be
+		// wrong, which is worse: state.previewPlayer is what the tab teardown
+		// destroys, so every visit would leave a live socket behind.
+		let gen = 0;
+
 		function attach(kind) {
+			const mine = ++gen;
+
+			// Whatever is running loses its handle here, so it has to be closed
+			// here. In the synchronous case there is nothing yet to close.
+			if (state.previewPlayer) {
+				try { state.previewPlayer.destroy(); } catch (e) {}
+				state.previewPlayer = null;
+			}
+
 			const impl = window.MajesticTransport.impl(kind);
 			if (!impl) return;
-			state.previewPlayer = impl.attach(video, {
+
+			const p = impl.attach(video, {
 				stream: stream,
-				onState: (st, detail) => {
-					if (kind !== 'webrtc') return;
+				onState: (st) => {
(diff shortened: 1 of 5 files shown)


widgetii:
Fixed in `710d013`. The finding is right, and it is the same bug I fixed on the Preview page and then walked straight into again in the new caller — `MajesticWebRTC` reports `'fallback'` from inside `attach()`, the handler installs MSE, and the outer assignment then buries it under the façade that just failed.

Worth being precise about why it matters, because the picture looks right either way: `state.previewPlayer` is what the tab teardown destroys, so every visit to Live adjustments left a streaming socket open behind it.

Same fix as the other caller — a generation stamped at attach, so an attachment superseded while it was still being built drops what it made rather than burying the replacement. Re-attaching now also destroys the outgoing player, which the asynchronous fallback path needed and did not have.

**My test missed this for exactly the reason the bug is easy to ship**: it asserted a picture arrived, and a picture arrives with the handle wrong. It now counts open WebSockets and then leaves the tab:

```
old code:  1 socket on the panel, 1 still open after leaving   <- fails
new code:  1 socket on the panel, 0 after leaving              <- passes
```

Checked in both directions rather than assumed — the test fails on the pre-fix build.

Fourteen test files across the preview and settings suites pass.

PR #194, opened 2026-08-27
Preview: talkback and per-session stats

widgetii:
majestic's `/webrtc` debug page carries two things this player does not — a microphone button and a stats readout. That page is being retired (majestic side, separately): it duplicates this player, and its script lives inside a C string literal where review found four bugs in an afternoon. This moves the parts worth keeping here first, so the redirect lands on a page that lost nothing.

## Talkback

The camera will not take a send-only audio section. It reads the direction as what it may send *into*, so it accepts `recvonly` and `sendrecv` and refuses `sendonly`. So talking necessarily opens the camera's audio as well, and `setMic()` unmutes to match rather than leaving the camera encoding Opus into a muted element. The tooltip says so.

The button appears only where all three preconditions hold: `audio.outputEnabled` on the camera, WebRTC as the live transport, and a secure context — a browser hands over no microphone on plain HTTP, so it would otherwise be a dead end.

Every way talkback can end is handled, because each one leaves a live capture behind if it is not:

| ending | handled by |
|---|---|
| permission refused | `catch`, reported with the reason |
| device unplugged / permission revoked mid-session | `micTrack.onended` |
| camera declines our half (`audio.outputEnabled` off) | `currentDirection` check after the answer |
| second click while the prompt is up | `micBusy` |
| player destroyed by a transport switch while the prompt is up | `closed` recheck after the await |
| leaving the page | `destroy()` releases first |

That list is not hypothetical — it is what review found in the page this replaces, three of them in the same shape: a control whose state was decided before an await and trusted after it.

## Stats

Both ends side by side, because they disagree exactly where it matters. The browser's `getStats` gives picture, rate, cumulative loss, jitter and round trip; the camera's own once-a-second line — previously read by nothing but the debug page — gives ICE and DTLS state, its REMB estimate, keyframe requests and audio-in. REMB is the camera's *opinion* of what the link will carry rather than a measurement of what arrived, and audio-in is the far half of talkback: the browser says it sent, the camera says it arrived. A microphone showing `sending` with the camera's counter stuck at zero is the case worth being able to see.

## Testing

No test suite in this repo, so the mic lifecycle was driven headless against stubs (`vm` context, stubbed `RTCPeerConnection`/`getUserMedia`/`WebSocket`) before landing — seven cases, each asking the same question: did the microphone stop? All pass, including the renegotiation timing that made the first version of the camera-declines test lie.

MSE keeps the façade with no-ops; there is no way to send a microphone up a media-source stream, and the page still never asks which transport it attached. `mj-settings.cgi` shares this markup but does not load `preview-page.js`, so the new controls stay hidden there.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,9 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## What this is
 
-WebUI for [OpenIPC Firmware](https://github.com/openipc/firmware) — served on port 80 of the camera by an embedded httpd that runs `haserl` CGI scripts. There is no compile step, no package manager, no test suite. "Building" means copying the tree onto a running camera.
+WebUI for [OpenIPC Firmware](https://github.com/openipc/firmware) — served on port 80 of the camera by an embedded httpd that runs `haserl` CGI scripts. There is no compile step and nothing the browser loads comes from a package manager. "Building" means copying the tree onto a running camera.
+
+`npm test` runs what tests there are: plain Node, no framework, no dependency — `tests/talkback.test.js` drives the WebRTC player's microphone lifecycle against stubs, and `tests/transport.test.js` covers the STUN/TURN list built for `RTCPeerConnection`. Both exist because their subject is timing or configuration parsing that a browser cannot be made to reproduce on demand, and because getting either wrong is silent: a microphone that stays captured, or a session that negotiates and carries nothing. Most of this repo is CGI and DOM wiring that these cannot reach; do not read the suite as coverage of the WebUI.
 
 Authentication is HTTP Basic against `/etc/shadow` (user `root`, default password `12345`); `common.cgi:check_password` forces a redirect to `fw-interface.cgi` until the default password is changed.
 
--- a/package.json
+++ b/package.json
@@ -2,10 +2,11 @@
   "name": "majestic-webui",
   "version": "0.0.0",
   "private": true,
-  "description": "OpenIPC majestic web UI — dev tooling that builds the minified distribution tarball consumed by the firmware. The shipped source stays unminified and readable; minification happens only in CI.",
+  "description": "OpenIPC majestic web UI \u2014 dev tooling that builds the minified distribution tarball consumed by the firmware. The shipped source stays unminified and readable; minification happens only in CI.",
   "license": "MIT",
   "scripts": {
-    "build:dist": "sh tools/build-dist.sh"
+    "build:dist": "sh tools/build-dist.sh",
+    "test": "node tests/talkback.test.js && node tests/transport.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/assert.js
+++ b/tests/assert.js
@@ -1,0 +1,30 @@
+// The smallest thing that can fail a build.
+//
+// No framework on purpose: this repo ships to a camera's rootfs and has no
+// build step, so a test dependency would be the only reason anyone needed npm
+// to work on it. Node alone runs these.
+'use strict';
+
+let failures = 0;
+
+function check(name, cond, detail) {
+	if (cond) {
+		console.log('  ok   ' + name);
+	} else {
+		failures++;
+		console.log('  FAIL ' + name + (detail ? ' — ' + detail : ''));
+	}
+}
+
+function group(name) {
+	console.log(name);
+}
+
+// Call once at the end. Exits non-zero if anything failed, which is what makes
+// `npm test` mean something.
+function done() {
+	console.log(failures ? '\n' + failures + ' failed' : '\nall passed');
+	process.exit(failures ? 1 : 0);
+}
+
+module.exports = { check: check, group: group, done: done };
--- a/tests/talkback.test.js
+++ b/tests/talkback.test.js
@@ -1,0 +1,272 @@
+// preview-webrtc.js's talkback state machine, driven against stubs.
+//
+// Every case asks the same question: did the microphone stop? A capture that
+// outlives the control able to stop it is the worst thing this player can do,
+// and each of these is a route to it that review found in the debug page this
+// replaced — a permission grant that outlives the mode it was asked for, two
+// clicks racing into two grants, a track that ends on its own, a camera that
+// declines the direction after the browser has already lit the microphone.
+//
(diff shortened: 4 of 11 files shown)


widgetii:
All five addressed in 03e5e68.

**1. Send-only talkback rejected** — right, and fixed. I checked the camera's behaviour before changing it: it answers `sendrecv` or `sendonly` on the audio section and never `recvonly`, so with today's answerer this end never actually settles on `sendonly` and the described session cannot occur. But that is a fact about one answerer, not about the negotiation — a camera with no audio source that still accepts talkback *would* correctly answer `recvonly` — and testing for `sendrecv` alone would then stop the capture and tell the user the camera was not listening. `micSending()` now accepts both directions that contain "send", which is what the question actually is. New test case covers it.

**2. Talkback permanently disabled** — real, and the nastiest of the five. The path is exact: switching transport mid-prompt destroys the player, the grant lands in a closed one, which releases it *silently* and reports no ending, so the control kept `disabled` and "Asking…" from a player that no longer exists. `syncTalkCtl()` now returns the button to rest unconditionally rather than depending on an ending that in this one path never arrives.

**3. ICE config misses the initial session** — real. The getter is only read at `open()`, so a blind attach stayed host-candidates-only until something else happened to reconnect. Off-LAN that costs the full no-signal timeout and is exactly the sequence that demotes a browser to MSE — which would have made the very regression this PR exists to prevent show up anyway, just later. The late-config correction now re-attaches for ICE when it has not already reopened for the stream change.

**4. Talkback shown before acceptance** — real. There is now a `live` state between `asking` and `on`: capturing, not yet accepted. It leaves the button **enabled**, unlike `asking` — by then a microphone is running, and a control that cannot stop a running microphone is the wrong control. `on` waits for the answer, and so does unmuting the speaker.

**5. Stats panel does not return** — real. Both the transport switch and the checkbox now go through `syncStatsCtl()`, so "is the panel showing" has one answer instead of two that had to agree.

Tests: 24 checks now, including the microphone-only session from #1 and assertions that `on` follows the answer rather than the grant, and that a declined session never claims `on`. All passing.

widgetii:
Second round done in 3525337 — two fixed, one already fixed.

**Rejected talkback leaves audio** (new) — real, and a good catch. `setMic()` unmuted at the moment of capture, so a camera that then declined left its audio audible while the Listen control still read "Muted": sound the page could not account for and the user could not directly stop.

The fix removes the cause rather than adding an undo. Nothing needed unmuting that early — `open()` builds the sendrecv transceiver from `wantMic`, not from `wantAudio`, so the offer never depended on it. The element is now unmuted in the acceptance branch, where the direction is settled and the page is told in the same breath. The refusal path has nothing left to reverse.

**Stats polls overwrite newer data** — real. The interval is not async-aware, so a slow `getStats()` and a fast one could settle out of order. Worth spelling out why it is worse than a misdrawn row: `rateBytes` is the divisor for the *next* tick, so one stale write prints a bitrate of zero or of twice the truth and keeps doing it. Now one poll at a time — a skipped tick loses a sample, interleaving them corrupts every sample after.

**Talkback remains permanently disabled** — this one is already fixed and I think it was carried forward without re-evaluation. The description still matches the pre-fix code ("the rebuilt control only clears `checked`"), but on the reviewed head `syncTalkCtl()` reads:

```js
if (talk) { talk.checked = false; talk.disabled = false; }
if (talkLbl) { talkLbl.textContent = '🎤 Talk'; talkLbl.title = TALK_TITLE; }
```

It returns the control to rest unconditionally on every attach, which is exactly the path the finding describes — precisely because the closed player releases the grant silently and reports no ending. Happy to look again if there is a route to it I have not seen.

Tests: 26 checks, including the element being muted after a refusal and unmuted only after acceptance. All passing.

widgetii:
Added a test suite in 448b2ff, so the coverage that majestic's C suite had for this does not just disappear with the page.

The camera used to build this list itself, for the debug page, and had tests for it there; that code went with the page and the rules moved to `preview-transport.js`. Without this the relay-credential rule would have gone from tested-in-CI to nobody's-job.

Two files, plain Node, no framework and no dependency — this tree ships to a camera's rootfs, and a test dependency would be the only reason anyone needed npm to work on it. `npm test` runs both.

- **`tests/transport.test.js`** — the ICE list: the camera default when unset, every YAML spelling of "none" (`iceServers: off` arrives as the string `"false"`), separators, and a relay dropped on its own rather than taking the STUN entries with it.
- **`tests/talkback.test.js`** — the microphone lifecycle, eight cases asking one question: did the capture stop?

Verified by mutation rather than asserted. Reverting `micSending()` to sendrecv-only, unmuting at the grant again, and removing the `micBusy` guard each fail the suite:

```
FAIL track kept
FAIL reported on — [["asking",""],["live",""],["off","the camera is not accepting audio"]]
FAIL element still muted
FAIL one getUserMedia — 3 calls
```

Worth saying plainly: this is eight cases over two files, not coverage of the WebUI. CLAUDE.md now says so, and says why these two subjects earned tests when the CGI and DOM wiring around them did not — their failures are silent ones. A microphone that stays captured does not throw.

widgetii:
Both remaining findings are stale — each cites code the fix removed, and the reviewer carried them forward rather than re-evaluating.

**1. Rejected talkback leaves audio.** Cites `wantMic = true; wantAudio = true; video.muted = false;` in the grant handler. That is exactly what 3525337 deleted. The unmute now happens only in the acceptance branch:

```js
if (micSending()) {
    // Accepted, so the camera's audio comes with it. Now is when the element
    // may be unmuted: the direction is settled and the page is told in the
    // same breath, so the Listen control and the sound agree.
    wantAudio = true;
    video.muted = false;
    onMic('on', '');
```

A refusal therefore has nothing to undo. `tests/talkback.test.js` asserts both halves — `element still muted` on the refusal path, `unmuted on acceptance` on the other — and reintroducing the unmute at the grant fails the first.

**3. Talkback remains permanently disabled.** Cites `if (talk) talk.checked = false;` from `attachPlayer()`. That line is gone; `syncTalkCtl()` does the whole reset unconditionally, which is the point — the closed player releases the grant silently and reports no ending, so nothing can be left to an `onMic` that never arrives.

Neither line is in the tree. Merging on that basis; if either has a route I have not seen, I will take it as a follow-up rather than hold the branch.

PR #197, opened 2026-08-28
Preview: a stream choice per page, and a tighter control row

widgetii:
All from #184, where the reporter has now run this on his own cameras and reported back.

## A remembered stream per page

> I think the remembered source selection should be stored **separately** for `Preview` and `Live Adjustments`, since the two pages may serve different purposes

Done, keyed by page.

Splitting it alone would have made things **worse**, which is worth saying because it nearly shipped that way: Live adjustments had no Main/Sub control, so a separate key would have pinned it to the substream permanently — and the case that makes the remembered choice worth having at all, `video0` cropped and `video1` not, is exactly the case where the substream is the wrong picture.

So that panel has its own control now, on the header line beside "Live preview", offering Sub only where one exists.

## The control row

> the `Stats` button unnecessarily takes up valuable vertical screen space. It could be placed on the same line as the other buttons, but aligned to the right.

Every control now sits on one flex line, with Stats pushed to the far end — it is the only one of them that opens a panel rather than changing what is playing. The adaptation notice stays a block of its own, because it is a sentence and not a control.

> The `This browser` and `This camera` columns … could be renamed to `Browser` and `Camera`

Done. They were paying rent on a narrow panel.

## Testing

`npm test`. The split is covered including the part that matters — choosing on one page leaves the other alone — and verified by mutation: collapsing both back to a single key fails two cases.

## Not in this PR

Two reports from the same comments are still open and are answered on the issue rather than here: a substream set to 1 FPS appearing to update faster (most likely a pending pipeline reload, since `videoN.fps` is not one of the live-applied fields), and a flicker when switching to WebRTC on a camera with no substream. The second is a real report but the proposed fix — disabling the toggle when Sub is off — would be wrong: WebRTC does not require a substream, and does work on Main for browsers that accept its profile.

### DIFF
--- a/tests/transport.test.js
+++ b/tests/transport.test.js
@@ -22,8 +22,8 @@ const SRC = path.join(__dirname, '..', 'www', 'a', 'preview-transport.js');
 // without — every read and write in the module is wrapped, and a module that
 // threw without storage would fail in a private window too, which is a real
 // browser state and not a hypothetical one.
-function load(withStorage) {
-	const store = {};
+function load(withStorage, seed) {
+	const store = Object.assign({}, seed || {});
 	const ctx = { window: {}, console: console };
 	if (withStorage) {
 		ctx.localStorage = {
@@ -113,5 +113,26 @@ check('no storage reads as no preference',
 let threw = false;
 try { noStore.chooseStream('preview', 1); } catch (e) { threw = true; }
 check('and writing without storage does not throw', !threw);
+
+group('upgrading from the single shared key');
+// A previous release wrote one key for both pages. Both inherit it — that is
+// what the person was looking at — and it is then thrown away. Dropping it
+// instead would return anyone who had chosen Main to the substream default,
+// and the reason to choose Main is that the substream shows the wrong picture.
+const up = load(true, { 'mj-preview-stream': '0' });
+check('preview inherits the old choice', up.chosenStream('preview') === 0);
+check('so does live', up.chosenStream('live') === 0);
+// Read once: choosing on one page afterwards must not be undone by the old
+// value coming back the next time the other page asks.
+up.chooseStream('live', 1);
+check('and the old key does not resurrect', up.chosenStream('live') === 1);
+check('while preview keeps what it inherited', up.chosenStream('preview') === 0);
+
+// A per-page answer already recorded wins over the legacy one.
+const both = load(true, {
+	'mj-preview-stream': '0', 'mj-preview-stream:preview': '1' });
+check('an existing per-page choice is not overwritten',
+	both.chosenStream('preview') === 1);
+check('while the page without one still inherits', both.chosenStream('live') === 0);
 
 done();
--- a/tests/transport.test.js
+++ b/tests/transport.test.js
@@ -90,18 +90,28 @@ group('the remembered Main/Sub choice');
 // the preview box, wants Main. The default is still Sub — that is the common
 // case — but the answer has to survive a page load or they re-pick it for ever.
 const t = load(true);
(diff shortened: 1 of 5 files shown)


widgetii:
Both fixed in 8bda9b2, and the second one had a twin.

**Legacy stream choice discarded** — correct, and it would have hit precisely the wrong people. The previous release wrote one unsuffixed key for both pages; reading only the new per-page keys silently discarded it, returning everyone to the substream default on upgrade. Since the reason to choose Main is that the substream shows a different picture — the cropped-`video0` case this whole feature exists for — the setting would have been lost by exactly the users who needed it.

Both pages now inherit the old value, because one value serving both is what the person was actually looking at, and it is read once and thrown away. A per-page answer already recorded wins over it. Four new cases cover the upgrade, including that the old key does not resurrect and undo a later choice.

**Hidden Sub remains selectable** — also correct. Bootstrap's `btn-check` is a real radio behind CSS, so hiding the label leaves the input in the tab order and arrow-key navigation could select a stream the camera does not have.

It is worth noting where else that lives: **the Preview page has had the same hole since its Main/Sub control existed** — `#mj-sub` is hidden and `#mj-stream-1` was never disabled. Only the new instance was in this diff, but shipping the fix for one and not the other would leave the older, more used page broken. Both are guarded now.

Verified by mutation rather than by the suite going green: removing the migration call fails four cases.

widgetii:
Fixed in 3226867. The finding is right and the reasoning behind it is the part worth keeping: `write()` swallows storage failures on purpose — a private window must not break the page — so "it did not throw" was never evidence the value landed. Deleting the legacy key first and then failing to write leaves nothing to try again from.

New keys first now, old key only once the new ones read back.

The transport preference above it had the identical shape — `migrate()` dropped `mj-transport` before writing its replacement — and is fixed with it. Not in this diff, but it is the same hazard on the same storage in the same file, and shipping the fix beside an untouched copy of the bug is not a thing to do.

Tested against a stub that accepts removals and refuses writes, which is what an exhausted quota looks like:

```
legacy key after a failed migration: {"mj-preview-stream":"0"}
KEPT — correct
```

Verified by mutation — restoring the delete-first order fails that case:

```
FAIL and the legacy key is kept for the next attempt — {}
```

One thing I corrected while adding it: my first version of that assertion was `stuck.__store === undefined || true`, which is true unconditionally. It passed, proved nothing, and would have sat there looking like coverage. The store is now reachable from the test and actually asserted.

usa-:
Forgot to mention: the 1 FPS problem no longer occurs in today's midday version.

widgetii:
This one is stale now — it matches the line text but not its context.

The finding cites `write(STREAM_KEY, null);` at what is now line 167. There are two of those, and they are different things:

```js
if (old !== '0' && old !== '1') {
    write(STREAM_KEY, null);      // ← line 167, the one cited
    return;
}
let carried = true;
['preview', 'live'].forEach(function (w) {
    if (read(streamKey(w)) !== null) return;
    write(streamKey(w), old);
    if (read(streamKey(w)) !== old) carried = false;
});
if (carried) write(STREAM_KEY, null);   // ← the guarded one
```

Line 167 is the branch where the stored value is neither `0` nor `1` — corrupt, or written by something that is not this code. There is no fallback to lose there: the value cannot be carried anywhere, so removing it is the point rather than a hazard. Keeping it would mean re-parsing the same garbage on every call for ever.

The delete the finding was originally about is the second one, and it now runs only after both per-page writes have been read back — which is what the quota test covers, and what fails when the order is put back the old way.

Merging on that basis.

PR #198, opened 2026-08-28
fw-reset: show the erase as it happens, and hand the camera back

widgetii:
Closes the three trailing points on #154: the reset stalls for five seconds, shows no progress while it erases, and never takes you back to the UI afterwards.

All three come from the same place — `fw-reset.cgi` ran `sysupgrade -s -n -x --web`, and every one of those flags was written for a sysupgrade that no longer behaves that way.

### `-x` is inert and expensive

It used to mean "leave the reboot to me", back when this page hopped to `fw-restart.cgi` at the end to take it. sysupgrade refuses `--no_reboot` when the run rewrites the flash the camera runs from, and `-n` qualifies: `rootfs_data` is the upper layer of the overlay the running root is assembled from. So it prints four lines of NOTICE and `sleep 5` waiting for a Ctrl-C no web page can send, then reboots anyway.

### `-s` was hiding the erase meter

`set_progress()` is `busybox "$@" | awk '{print NR, $1}'` when silent mode is on. `flash_eraseall` redraws one line with a bare `\r` and emits no newline until it is done, so awk saw the whole erase as a single record and printed `1 Erasing` — once, after the fact. That is exactly the `1 Erasing` in the reporter's transcript. Dropping `-s` puts the real meter back: one redraw per erase block, ~5/s, ~30s of a 34s run on a 8.6 MB overlay. No firmware change needed.

### The hand-back was unreachable code

`makeTextFileLineIterator()` put `location.href = '/cgi-bin/fw-restart.cgi'` *after* its read loop. The stream does not end — it dies with the camera, so `rd.read()` rejects and everything after the loop is skipped. Even had it run it was wrong twice: it aimed at a camera already going down, and `fw-restart.cgi` issues its own `reboot -d1`.

`fw-reset.js` now streams the run, watches for `Protected: flashing` / `Unconditional reboot` / `Aborting.`, catches the stream error, waits for the camera to go and come back, and navigates to `status.cgi` — where majestic serves the login page itself, since a reset clears the session and takes the stay-signed-in key with it.

The second commit fixes something the hardware run exposed: the watch was armed by the stream ending, and on a real reset the announcement came at 34.6s while the read did not reject until 81.2s. Forty-six seconds of "do not navigate away" about a camera that had already gone. It is armed by the announcement now, the way `fw-update.js` has always done it, for the reason its comment gives.

### Shared terminal writer

Rendering a `\r` meter in a `<pre>` needs something that plays the cursor back, which `fw-update.js` already had (#134). Extracted as `termWriter()` in `main.js` and used by both. `runCmd()` and `makeTextFileLineIterator()` are gone — `fw-reset.cgi` was their only caller, and they reached for an `el` the page had to leak into global scope.

### Verification

Hardware, hi3516ev300 (`master+2f558d4`), real `sysupgrade -n --web` driven from a real browser:

```text
[ 4.6s] Erasing 64 Kibyte @ 10000 -   0% complete.
[33.1s] Erasing 64 Kibyte @ 860000 -  96% complete.
[34.6s] Unconditional reboot            <- status flips to "waiting for it to come back"
        camera back; navigation to status.cgi → 302 /login.html?next=%2Fcgi-bin%2Fstatus.cgi
```

No NOTICE, no stall, live meter, correct hand-off. Also exercised in a browser against a recorded 34s reset stream: the abort path (camera untouched, transcript kept, heartbeat resumed), the connection-dies-mid-erase path, and a socket that lingers 20s past the announcement.

`termWriter()` was checked byte-identical to the previous inline renderer over that recorded stream at eight different chunk boundaries, so the upgrade page renders exactly as before.

### DIFF
--- a/www/a/fw-reset.js
+++ b/www/a/fw-reset.js
@@ -1,0 +1,175 @@
+// Factory reset: stream `sysupgrade -n --web` through j/run.cgi into the pane on
+// fw-reset.cgi, then wait for the camera to come back and hand the user over to
+// it. Vanilla JS; `$` and `termWriter` come from main.js.
+//
+// Not the /ws/upgrade socket fw-update.js uses: that endpoint exists to drive a
+// firmware flash, and a reset writes no image. run.cgi is the plain streaming
+// pipe this page has always used. What changed is what happens around it.
+//
+// The old version lived inline in the CGI and did two things that no longer
+// hold. It read the stream a line at a time and appended each one, which is
+// fine for `Stopping crond: OK` and useless for a meter that redraws itself
+// with a bare \r — hence the terminal writer. And when the stream ended it
+// navigated straight to fw-restart.cgi, on the assumption that sysupgrade had
+// been told -x and had left the rebooting to us. It hasn't for a while:
+// wiping rootfs_data rewrites the flash the camera is running from, so
+// sysupgrade reboots regardless and says so. Two things followed from that.
+// The hop landed on a camera that was already going down — and it would have
+// asked for a *second* reboot if it had arrived. And it never even ran: the
+// stream does not end, it dies with the camera, so the read rejects and the
+// navigation after the loop was never reached (issue #154). Hence the catch
+// below, and a watch that waits for the camera instead of racing it.
+(function () {
+	const out = $('#output');
+	if (!out) return;
+	const term = termWriter(out);
+	const dec = new TextDecoder('utf-8');
+
+	// Same two markers fw-update.js reads, for the same reasons. "Protected:
+	// flashing continues" is sysupgrade's point of no return, printed just before
+	// the erase; "Unconditional reboot" is it announcing the reboot it is about
+	// to take. die() prints "<reason> Aborting." — before the RAM pivot that
+	// leaves the camera untouched and running, so it is the one ending that must
+	// not start a reboot watch.
+	const flashMarker = /Protected: flashing/i;
+	const rebootMarker = /Unconditional reboot|Rebooting now/i;
+	const abortMarker = /Aborting\./;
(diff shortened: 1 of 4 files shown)


widgetii:
Good catch on `ping()`, and it was worse than flagged — the finding applies to the `run.cgi` stream too, which is the request that actually matters here.

Confirmed on a camera rather than argued about:

```console
$ curl -D - 'http://cam/cgi-bin/j/run.cgi?cmd=...'          # unauthenticated
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Basic realm="Authentication"              # <- Safari's dialog

$ curl -D - -H 'X-Requested-With: XMLHttpRequest' ...
HTTP/1.1 401 Unauthorized                                   # bare, nothing to prompt from
```

So a session that lapsed before the reset started would have popped exactly the dialog #154 was opened about, on the page fixing #154. `/` happens to be served unauthenticated on this build, so `ping()` was the milder half — but it is the same mistake and gets the same treatment.

Fixed by rebasing onto master, which had meanwhile landed `rawFetch()` for precisely this: the `X-Requested-With` header without `apiFetch`'s redirect-to-login. Both call sites in `fw-reset.js` use it now. This page must keep opting out of the redirect — losing the session is the expected end of a factory reset and the transcript has to survive it — but that was never a reason to opt out of the header.

Re-verified after the rebase: all four stream paths in a browser (reboot, abort, connection-dies-mid-erase, socket lingering past the announcement), `termWriter()` still byte-identical to master's inline renderer at eight chunk boundaries, and a real reboot end to end on a hi3516ev300 — live meter, watch armed on the announcement, hand-off landing on the status page.

PR #203, opened 2026-08-29
Preview: try a transport before committing the screen to it

widgetii:
Fixes the flicker reported three times on #184, using the reporter's own rule:

> nothing on the screen should change until it is known that WebRTC is actually available

## What it did

A transport switch destroyed the running player, cleared the video element, and attached the replacement to it. So clicking **WebRTC** on a camera that could not serve this browser cost a blank frame, a visible reconnect and a trip back to MSE — for a session that never worked.

The fix originally proposed, disabling the button when no substream exists, would have been wrong: WebRTC does not require a substream and runs on the main channel for any browser that accepts its profile. Hiding the control would have removed a working feature from them.

## What it does now

One element cannot hold two players — MSE drives it through `src`, WebRTC through `srcObject` — so there are two, and only ever one visible. A switch attaches the new player to whichever is idle and **leaves the other playing**. The picture changes exactly once, when the trial reports that it is playing.

A trial that fails is destroyed and the screen never learns it happened, beyond the toggle coming back up with the reason in its tooltip. Everything a trial says short of `playing` is dropped on purpose — `connecting`, `nosignal`, `error` are precisely the states that used to leak onto the badge while a switch flailed.

**The chain still ends at MJPEG.** When the live player dies its replacement is staged like any other switch, and the dead one keeps its last frame rather than blanking. If that replacement also fails, both are released and the page falls through. Marking the dead one is load-bearing: without it a second failure would leave a frozen frame on screen with nothing left to move it.

## Both players, one implementation

Live adjustments had the same shape and needed the same fix — less visible there, since that panel has no toggle and only ever switches automatically, but the same thing happens when a WebRTC session dies mid-watch.

Writing it twice was the obvious next move and the wrong one, so the swap is now `preview-swap.js` and both pages use it. The note in `preview-transport.js` explaining why the fall-back dance is *not* shared said "the dance is a dozen lines; the rules below are not" — true of a dozen lines, not of a state machine where two players are live at once, each attachment needs its own id because "is this the current generation" cannot describe two current things, and a dead player must be marked or its last frame stays for ever.

What stays with each caller is every decision: what to try next, what to put on the badge, what a failure means. The Preview page falls through to MJPEG and keeps its toggle in step; the panel has neither and stops at MSE.

Teardown on the settings page goes through the swap now. Destroying only the visible player would leave a transport still being judged behind on every visit — a live socket with no handle, which is the leak `state.previewPlayer` was tracked to prevent in the first place.

## Testing

A new suite drives the switch against stub players — 27 checks, including the two that matter most:

- a failed trial does **not** destroy the working player
- a trial in any non-playing state does **not** touch the badge

Verified by mutation, not by going green. Restoring the old tear-down-first order fails ten checks; letting a trial report to the badge fails the state-leak one. The same 27 pass unchanged after the extraction, which is what made sharing the code safe to do rather than a rewrite to be careful about.

This page has twice shipped the opposite bug — a fallback burying the player that was working — and both times the picture looked right while the handle was wrong, which is exactly what looking at it cannot catch. That is why the harness exists rather than a manual check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/talkback.test.js && node tests/transport.test.js"
+    "test": "node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/staging.test.js
+++ b/tests/staging.test.js
@@ -15,7 +15,10 @@ const path = require('path');
 const vm = require('vm');
 const { check, group, done } = require('./assert');
 
-const SRC = path.join(__dirname, '..', 'www', 'a', 'preview-page.js');
+const A = (f) => path.join(__dirname, '..', 'www', 'a', f);
+// The swap machinery is a module of its own now; the page is what decides what
+// its outcomes mean, and that division is part of what these check.
+const SRCS = [A('preview-swap.js'), A('preview-page.js')];
 
 const IDS = [
 	'live-mjpeg', 'live-video', 'live-video-b', 'mj-audio-ctl', 'mj-badge',
@@ -79,6 +82,7 @@ function load(pickedTransport) {
 
 	const ctx = {
 		window: win,
+		MajesticSwap: null,   // preview-swap.js assigns it onto window below
 		console: console,
 		MajesticVideo: impls.mse,
 		MajesticWebRTC: impls.webrtc,
@@ -93,7 +97,11 @@ function load(pickedTransport) {
 		Promise: Promise,
 	};
 	vm.createContext(ctx);
-	vm.runInContext(fs.readFileSync(SRC, 'utf8'), ctx);
+	vm.runInContext(fs.readFileSync(SRCS[0], 'utf8'), ctx);
+	// The browser reaches window.MajesticSwap through the global scope; a vm
+	// context has no such link, so hand it over explicitly.
+	ctx.MajesticSwap = win.MajesticSwap;
+	vm.runInContext(fs.readFileSync(SRCS[1], 'utf8'), ctx);
 	env.el = (id) => env.els['#' + id];
 	return env;
 }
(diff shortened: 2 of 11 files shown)


widgetii:
All four addressed in 36b4f1e. Two of them would have put the flicker back by a different door, which is the useful part of this review.

**MSE invalidates stored elements** — real, and the one I would not have found. The MSE player replaces its video element on every reconnect (`cloneNode` + `replaceChild`, keeping the id), so the nodes the swap captured were detached within a session: picking a spare by node identity returned the detached original, a trial attached to something nobody could see, and every show/hide wrote to nothing. Elements are getters now and everything resolves by slot. The stub player in the tests clones like the real one, so this cannot come back quietly.

It was marked resolved in the report, but it was not — the code still held nodes. I checked `preview.js` rather than the marker.

**Promotion reconnects the replacement** — real, and the sharpest of the four. `settle()` ran after promotion and called `setAudio(true)` for anyone listening; both players answer that by renegotiating, which blanks the picture. So a trial that had just proved itself was torn down at the moment it took over — precisely the flicker this PR exists to remove, moved one step later and only for people with audio on. The preferences go in at `attach()` now, so a trial is opened as the thing it will become.

**Staging promotes stale stream** — real. Main/Sub changes reached the visible player only, so a trial judged across a channel change was promoted onto the channel the viewer had already left. Both are told now.

**Failed MSE switch misreported** — real. Only a WebRTC failure restored the toggle, so switching to MSE and having it fail left the control unchecked and the preference stored as MSE while WebRTC carried on playing: the UI reporting the opposite of the screen, and the next load retrying the failure. The toggle now describes whatever survived.

Also fixed while in there, from the *Late config overrides toggle* finding: a delayed config answer no longer restarts WebRTC while a trial is in flight. That trial is the viewer's own click, and it was opened after the ICE list landed anyway.

## Tests

33 checks. The two structural faults were each verified by mutation rather than by the suite going green:

```
── mutation: cache the node on first use (the original bug)
  FAIL the old element is hidden
  FAIL while the old node is hidden

── mutation: setAudio after promotion
  FAIL and is not told to turn audio on afterwards
```

widgetii:
Rebuilt on master after #201 landed. The two touched the same block — #201 added the per-channel note sync to the Main/Sub handlers, this PR replaced those handlers so a stream change also reaches the trial being judged — and both are wanted, so the resolution does all three: the visible player, the trial, and the note.

Suite green on the merge.

PR #204, opened 2026-08-29
Preview: an Auto source that follows the size it is drawn at

widgetii:
The `Auto` source from #184, built now that the objection to it has been withdrawn.

That objection was that Auto cannot see cropping, so on a camera whose `video0` is cropped it would silently show a different picture. The reporter disposed of it:

> If the user needs the full frame, the `Sub` button is still available. If the user wants to see the cropped `Main` stream, they can select `Main`.

He is right, and the part I had missed is that **the preview already defaults to Sub** — so a cropped `video0` was already not what you saw until you asked for it. It was never an argument about Auto.

## The rule

A third button beside Main and Sub, offered only where there are two streams to choose between. **Not the default**: a browser with no preference still opens on the substream, and Auto is remembered like the other two once chosen.

**By area**, his refinement and the correct one — `704x576` and `1280x720` are not ordered the same way by width as by the picture they carry, and it is pixels that cost bandwidth. Nearest at or above the box the player is drawn in, else the largest below it: scaling a bigger picture down loses nothing visible, scaling a smaller one up does.

**CSS pixels, not device pixels.** On a 2x display the larger stream is sharper, but this feature exists for links that cannot carry the larger stream at all, and doubling the demand on every phone is the wrong side to err on. Worth stating because it is a real trade and the other choice is defensible.

**At most one change a second**, his limit. A change inside that window is *deferred, not dropped* — the size that triggered it is the size it still is, and forgetting it would leave the wrong stream up until the window next happened to move. That distinction is the one the tests are really for.

Which source is in use was already covered: the badge names the resolution.

## Testing

Both halves are invisible when they go wrong — a rule that compares widths is right about half the time, and a rate limit that drops changes looks like it works until you resize once and stop. So both are tested and mutation-checked:

```
── mutation: compare by width, not area          → 5 failures
── mutation: drop rate-limited changes           → FAIL but it is applied once the window passes
```

## Not included

Live adjustments does not get an Auto button. Its preview is a small fixed panel, so Auto would pick the substream essentially always — which is already its default. Adding a control whose answer never changes would be clutter; say the word if you would rather have it for consistency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js"
+    "test": "node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/auto-source.test.js
+++ b/tests/auto-source.test.js
@@ -1,0 +1,201 @@
+// The Auto source rule: pick the stream closest to the size the player is
+// drawn at, and follow the window without cutting the session every frame.
+//
+// Tested here rather than by eye because both halves are invisible when they
+// go wrong: a rule that picks by width instead of area is right about half the
+// time, and a rate limit that drops changes instead of deferring them leaves
+// the wrong stream up until the window happens to move again.
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const vm = require('vm');
+const { check, group, done } = require('./assert');
+
+const A = (f) => path.join(__dirname, '..', 'www', 'a', f);
+const SRCS = [A('preview-swap.js'), A('preview-page.js')];
+
+const IDS = [
+	'live-mjpeg', 'live-video', 'live-video-b', 'mj-audio-ctl', 'mj-badge',
+	'mj-lightmon', 'mj-mute', 'mj-mute-lbl', 'mj-note', 'mj-stats',
+	'mj-stats-btn', 'mj-stats-ctl', 'mj-stream-0', 'mj-stream-1',
+	'mj-stream-auto', 'mj-auto', 'mj-sub', 'mj-talk', 'mj-talk-ctl',
+	'mj-talk-lbl', 'mj-transport', 'mj-transport-ctl', 'mj-transport-lbl',
+	'mj-transport-note', 'mj-vol', 'toggle-ircut', 'toggle-light',
+	'toggle-night',
+];
+
+function makeEl(id) {
+	return {
+		id: id, style: {}, hidden: false, checked: false, disabled: false,
+		textContent: '', title: '', value: 100, src: '', srcObject: null,
+		clientWidth: 640, clientHeight: 360, handlers: {},
+		addEventListener(ev, fn) { (this.handlers[ev] = this.handlers[ev] || []).push(fn); },
+		removeAttribute() { this.src = ''; },
+		fire(ev) { (this.handlers[ev] || []).forEach((f) => f()); },
+	};
+}
+
+// cfg: { box: [w,h], main: 'WxH', sub: 'WxH', subEnabled: bool, picked: choice }
+function load(cfg) {
+	const env = { made: [], els: {} };
+	IDS.forEach((id) => { env.els['#' + id] = makeEl(id); });
+	// As the markup ships them: the Sub and Auto labels start hidden and are
+	// revealed only where a second stream exists.
+	['mj-sub', 'mj-auto'].forEach((id) => { env.els['#' + id].hidden = true; });
+	if (cfg.box) {
+		['live-video', 'live-video-b'].forEach((id) => {
+			env.els['#' + id].clientWidth = cfg.box[0];
+			env.els['#' + id].clientHeight = cfg.box[1];
+		});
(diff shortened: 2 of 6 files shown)


widgetii:
Both fixed in 7c66d64, and the first would have made Auto useless on most cameras.

**Unset Main excluded from Auto** — right, and I checked the camera side rather than the symptom: `video1.size` ships with a default of `1280x720`, `video0.size` ships with **none at all**. Empty means whatever the sensor gives, which is the largest picture the camera has. Parsing that to `null` and dropping it left Auto choosing between one option — so on every camera whose main stream had never been set by hand, which is most of them, it would have sat on the substream for ever while looking like it was deciding.

Unknown now counts as sensor native. Ties go to the substream: two channels the same size, or two unknown ones, cost the same to decode and the cheaper one to carry.

**Element resizes go unnoticed** — also right. The decision is about the element, so the element is what should be watched. This page is responsive: a column reflowing, or something above it changing height, resizes the player without the viewport moving at all. `ResizeObserver` on both video elements where there is one, the window event where there is not.

## A test bug worth naming

While adding coverage for the first finding I found my own harness was lying. It built the stub config with:

```js
'video0.size': cfg.main || '1920x1080',
```

so the case that deliberately passed an empty size got the default back, and the unset-Main test asserted nothing. It passed against the broken code and the fixed code alike. Now `undefined` picks the default and `''` passes through, and both fixes are mutation-checked:

```
── mutation: exclude an unset size again
  FAIL Auto took Main for a box bigger than the substream — stream=1

── mutation: watch only the window
  FAIL both video elements are observed — 0
```

That is the second time in this PR series a `||` default has quietly disarmed a test, so it is worth writing down rather than just fixing.

widgetii:
Fixed in 6c94e3c, and this is the third time in this series the same trap has been hit, so it is worth naming plainly:

**the MSE player replaces its video element on every open and every reconnect** (`cloneNode` + `replaceChild`, keeping the id). Anything that holds a reference to one is holding a detached node shortly afterwards.

- The swap hit it with stored element references — fixed by resolving through getters.
- This hit it with a `ResizeObserver`, which would have stopped reporting the moment MSE reconnected, leaving Auto on the wrong stream indefinitely and looking like it had simply decided.

The observer now watches `#mj-player`, the container, which nothing replaces and whose width is what moves the video's anyway. The *measurement* is unchanged — still the video element, resolved fresh through the swap each time.

Mutation-checked, since the previous version also passed its test:

```
── mutation: observe the video node again
  FAIL the container is what is observed — live-video
```

widgetii:
Fixed in b6bb6dd. This is the best finding of the series and the worked example in it is exactly right — I reproduced the loop from those numbers before changing anything.

The video element is `width: 100%` with no height constraint, so **its height is the aspect ratio of whatever is playing**. Measuring the element's area therefore fed the decision its own result: at 800px wide, `1280x720` renders 800x450 and by area points at a `704x576` substream, which renders 800x655 and points back at the main stream. A switch every second, on a window nobody was touching, for ever.

The decision is taken on the container's width now, which does not move when the stream does.

## On area versus width

The finding also settles something I had argued for in the other direction, so it is worth being explicit. With a width-constrained box the arithmetic collapses: a stream is large enough exactly when **its own width** is, because the rendered height scales by the same factor as the rendered width. Comparing rendered areas was not a stricter version of comparing widths — it was the same comparison with the playing stream's aspect ratio mixed in, which is the bug.

So width decides and **area breaks ties**, which keeps the reporter's point — that one dimension must not be trusted alone — exactly where it still applies: two streams of equal width and different pictures.

## A test that asserted the bug

One existing case expected Sub at 800px wide against `1280x720` and `704x576` — and that expectation *was* the oscillating choice. It has been replaced by the two cases that describe the rule: a stream narrower than the box loses however its area compares (it is being upscaled, which is the one outcome worth avoiding), and area separates two streams of equal width.

There is also a case that reproduces the reported loop directly: it lets the element take the aspect ratio of whatever is playing, pokes the observer four times across the rate limit, and asserts the choice never moves.

```
── mutation: compare by area against the box again
  FAIL took Main — stream=1
  FAIL opens on Main, the only stream at least 800 wide — stream=1
```

widgetii:
Both halves fixed in 2993f8c.

**Reachable before the configuration lands** — right, and it applied to Sub as well, which I had only guarded from the page. A `btn-check` is a real radio behind CSS, so hiding the label leaves the input in the tab order; between page load and `mjConfig()` resolving, a keyboard could select a stream the camera may not have. Both inputs ship `disabled` in the markup now and the page enables them once it knows there is a substream — the page is no longer the only thing standing between a keyboard and a stream that does not exist.

**A stale Auto left running** — also right, and worse than it sounds: the callback disabled the control without clearing it, so on a single-stream camera you got no visible button selected and Auto still active behind a control nobody could see or unset. It is cleared, and Main is shown as the selection.

## The test for this was asserting nothing

Worth reporting rather than quietly fixing. The case existed and passed, but the stub never had Auto *checked* — the only code path that checks it is the one a single-stream camera skips — so `checked === false` was true before the fix and after it.

It now does what the report describes: delays the configuration, selects the invisible control in that window, and then lets the answer arrive. Which is the only version of the test that can fail:

```
── mutation: leave a stale Auto running
  FAIL the stale Auto choice is cleared
```

That is the third test in this series I have had to repair after finding it green against broken code — the others were a `||` default swallowing an empty value and an assertion that was `x === undefined || true`. Worth naming as a pattern: a case that passes the moment it is written deserves a mutation before it is trusted.

PR #205, opened 2026-08-29
updatewebui: add --help, and stop hiding the firmware's own WebUI

### LINKED ISSUE #202: Improve `updatewebui`: add `--help` and clarify overlay file handling during upgrades
I have a few questions and concerns about `updatewebui`.

First, there does not seem to be a way to display the program's usage information or available options. Could `updatewebui --help` be supported?

Second, `updatewebui` appears to copy files from the current WebUI version into the overlay. This means that once a file from the WebUI is present in the overlay, a newer version of the same file delivered by a subsequent upgrade may no longer take effect, because the overlay copy takes precedence over the updated file below it. In other words, an upgrade can update the underlying WebUI files, but those changes can remain hidden by copies of the old files that were previously placed in the overlay. This seems potentially problematic for future WebUI upgrades.

Finally, I have a personal modification in `/overlay/root/var/www/cgi-bin/p/motor.cgi`, where I replaced the original PTZ buttons with buttons for controlling the illumination via UART. After running `updatewebui`, my modified file is replaced by a version containing the original PTZ buttons. Is this the intended behaviour of `updatewebui`? In particular, is the purpose of copying the current WebUI files into the overlay to deliberately replace any existing files there, including user modifications?

openipc-ai:
Fixes #202.

### `--help` did not just print nothing, it wiped the WebUI

The script read `${1:-master}` as a branch name, so `updatewebui --help` fetched
`refs/heads/--help.zip`, got GitHub's 404 body, unzipped nothing — and then ran
`rm -f $(find /var/www -type f)` anyway, because the wipe came *before* any check
on the download. Reproduced on a lab hi3516av300:

```
# updatewebui --help
unzip: short read
Copy files to web directory
cp: can't stat '/tmp/tmp.qW1T8v//sbin/*': No such file or directory
cp: can't stat '/tmp/tmp.qW1T8v//www/*': No such file or directory
Delete temporary files
# echo $?
0
```

72 served files → 0, `status.cgi` 200 → 404, exit status 0. And because the
deletions went through the merged path, the overlay was left holding a whiteout
for every file, hiding the firmware's own copies for good — a reboot does not
undo it.

### The overlay concern in the issue is the same mechanism from the other side

Every file the tool installs is an overlay copy that hides the firmware's file
underneath, and it outlives the firmware upgrade that would have replaced it. So
an upgrade can ship a newer WebUI that never shows.

### What changed

- **Nothing is touched until the download is fetched, verified and unpacked**,
  and confirmed to hold a WebUI tree. Every failure path — 404, unreachable
  host, not a zip, truncated zip, wrong payload — now leaves the running WebUI
  exactly as it was.
- **Only files that differ from `/rom`'s are written**, and the overlay copy of
  a file the firmware already has byte-identical is dropped. On a current
  camera that is **37 overlay files instead of 80**: every `.cgi` matches what
  buildroot installed and is served from `/rom`, so a firmware upgrade can
  still replace it. Overlay use 844K → 468K.
- **Removals happen in the overlay's upper directory**, never through the
  merged path — which is also what clears the whiteouts an older run left. And
  only entries, never the directories holding them: removing a directory from
  the upper layer leaves the merged parent listing empty until reboot, with the
  files still readable one by one, so only a directory listing notices.
  Afterwards the script drops caches and `md5sum -c`s every path it touched,
  and says so if the merged view has not caught up.
- **`--restore`** takes the installed copies away so the firmware's WebUI shows
  through again. It also recovers a camera the old script wiped.
- **`--dry-run`** reports what would change and touches nothing.
- **`bin/*` → `/usr/bin`** is installed too. It was silently skipped, so
  `ntfy.sh` never updated.
- busybox `unzip` applies the caller's umask, 0077 on OpenIPC, so the old run
  installed the whole WebUI root-only (0600/0700). Modes are normalised to what
  the firmware ships.
- Writes are renamed into place rather than written over, so replacing this
  script while it runs cannot corrupt the shell reading it.
- A manifest at `/etc/webui/updatewebui.manifest` records what was installed,
  which lets the next run tell a local edit apart from a file the release
  simply changed. Local edits are still replaced — this installs a released
  tree, it does not merge into one — but they are archived to
  `/etc/webui/updatewebui-local-<stamp>.tar.gz` first, and `--dry-run` names
  them beforehand.

### Hardware validation

All on a lab hi3516av300 (imx415, kernel 4.9, jffs2 overlay), from a clean
firmware-only baseline each time:

| | result |
|---|---|
| `--help`, `--frobnicate` | usage printed / rejected, nothing touched |
| install (GitHub master and `--url=`) | 42 written, 37 overlay files, 75 served, manifest 80 lines |
| re-run | `Already up to date.` |
| local edit to `p/motor.cgi` | named, archived to `/etc/webui`, replaced |
| pages | status, mj-settings, preview, fw-update, info-logs, tool-files, `j/pulse.cgi` all 200 |
| `--restore` | back to 72 files, `diff -r /var/www /rom/var/www` identical, no reboot |
| 404 / unreachable / not-a-zip / truncated / non-WebUI zip | 72 files intact, `status.cgi` 200 |
| old script's `--help`, then `--restore` | 0 files recovered to 72, 404 → 200 |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,7 +21,15 @@ This is a majestic-side feature; the WebUI just provides the login page and the 
 
 ## Deploying / running
 
-- `sbin/updatewebui [branch]` — fetches a branch zip from GitHub, wipes `/var/www`, then copies `sbin/*` → `/usr/sbin` and `www/*` → `/var/www`. This is the canonical "deploy from source" path. Default branch is `master`.
+- `sbin/updatewebui [options] [branch]` — fetches a branch zip from GitHub and installs `www/*` → `/var/www`, `sbin/*` → `/usr/sbin`, `bin/*` → `/usr/bin` (the same payload `tools/build-dist.sh` hands buildroot). This is the canonical "deploy from source" path; default branch is `master`. `--help` lists the options, `--dry-run` reports what would change, and `--restore` takes the installed copies away again so the firmware's own WebUI shows through.
+
+    **It is overlay-aware, and has to be.** The rootfs is a read-only squashfs pivoted to `/rom` with a jffs2 overlay on top, so every installed file is an overlay copy that hides the firmware's own file underneath — and outlives the firmware upgrade that would have replaced it, hiding the newer file that upgrade delivered (#202). So the installer writes only the files that actually differ from `/rom`'s, drops the overlay copy of any file the firmware already has byte-identical, and prunes overlay entries this release no longer ships. On a current camera that is 37 overlay files instead of 80 — every `.cgi` matches what buildroot installed and comes straight from `/rom`.
+
+    Removals happen **in the overlay's upper directory** (`upperdir=` from `/proc/mounts`; `/overlay/root` on 4.x, `/overlay` on the 3.10 out-of-tree `overlayfs`), never through the merged path — a plain `rm /var/www/x` writes a whiteout that hides the firmware's copy for good. Only entries are unlinked, never the directories holding them: removing a directory from the upper layer leaves the merged parent listing empty until the next reboot. After the surgery the script drops caches and `md5sum -c`s every path it touched, and says so if the merged view has not caught up.
+
+    Nothing on the camera is touched until the download is fetched, verified and unpacked. The version this replaced wiped `/var/www` *before* checking anything, so `updatewebui --help` — an unknown branch — 404'd, unzipped nothing, and left the camera with no WebUI at all and whiteouts over the firmware's copies.
+
+    A manifest at `/etc/webui/updatewebui.manifest` records what was installed, which is what lets the next run tell a local edit apart from a file the release simply changed. Local edits are replaced, not merged, but they are archived to `/etc/webui/updatewebui-local-<stamp>.tar.gz` first (`--no-backup` opts out).
 - Edits to a running camera can also be made directly under `/var/www/cgi-bin/` and `/usr/sbin/`.
 - There is no local way to run the UI off-camera — every script assumes camera-side binaries (`majestic`, `yaml-cli`, `ipcinfo`, `fw_printenv`, `haserl`, `chpasswd`, `sysupgrade`).
 
--- a/sbin/updatewebui
+++ b/sbin/updatewebui
@@ -30,6 +30,12 @@ KEEP_BACKUPS=3
 # Kept in step with tools/build-dist.sh, which is what buildroot consumes.
 PAYLOAD='www:/var/www sbin:/usr/sbin bin:/usr/bin'
 
+# Everything this project has ever installed outside /var/www. A camera updated
+# by the manifest-less version of this script left no record of what it copied
+# there, and --restore exists largely for those, so it falls back to this list.
+LEGACY_SHARED='/usr/sbin/openwall /usr/sbin/setnetwork /usr/sbin/telegram
+/usr/sbin/updatewebui /usr/bin/ntfy.sh'
+
 branch=
 zip_url=
 mode=install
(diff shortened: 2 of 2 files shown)


openipc-ai:
Review addressed in edcd679. Five of the six changed code; one did not, and here is why.

### `unzip -t` — not the bug it looked like, removed anyway

The finding says busybox `unzip` has no `-t`. It does, on the target firmware:

```
# busybox unzip --help
Usage: unzip [-lnojpqK] FILE[.zip] [FILE]... [-x FILE]... [-d DIR]
	-t	Test
# unzip -t ok.zip >/dev/null 2>&1; echo $?
0
# head -c 4000 ok.zip > bad.zip; unzip -t bad.zip >/dev/null 2>&1; echo $?
1
```

The full install ran end-to-end against GitHub on that same box before the review, so valid updates were never being rejected. But busybox's own usage line omits `-t` while the option list below it includes it, so I would rather not build on that — and the check was redundant regardless. Unpacking into the scratch directory rejects both a non-zip and a truncated zip using nothing beyond the options busybox `unzip` has always had, so that is the only gate now. Both cases still refuse before anything on the camera is touched.

### The other five

**Backup failure permitted replacement.** Fixed, and it is the one I am most glad was caught. `backup_local` runs before any write, so refusing there leaves the camera untouched — and a failed archive is exactly the moment worth stopping for, because the run is about to replace work that exists nowhere else. The review's note about the pipeline was right too: busybox `tar` has no `-z`, and a shell pipeline reports only its last command, so `tar` and `gzip` are now two separately checked steps rather than one whose first half could fail unseen.

```
updatewebui: cannot write /nonexistent/dir/x.tar.gz, nothing has been changed
  rc=1
  local edit preserved: YES
  manifest untouched:   80 lines
```

**Removal failures were ignored.** Fixed. `remove_entries` unlinks and then checks the entry is actually gone; a survivor is named and the run refuses to commit on top of it. Verified by remounting `/overlay` read-only:

```
Drop 0 redundant and 1 stale overlay entry(ies)
Could not remove:
  /overlay/root/var/www/zzz-stale.cgi
updatewebui: the overlay did not end up as planned, so the manifest was left alone
  rc=1
  manifest untouched: YES
```

and the next run cleans up once the flash is writable again.

**Manifest write failure reported success**, and **the no-op path skipped the manifest entirely.** Both fixed; the manifest write is now a checked step of installing, including on the run that had nothing to install — that was the one case that never got a baseline, so a later local edit could not be recognised or archived. Note for anyone reproducing: busybox `mkdir -p` happily creates the whole chain, so a bogus path does not fail; I used the read-only squashfs.

**Restore left the previous updater's helpers behind.** Real, and the sharpest of the six, because it broke the recovery path on exactly the cameras it exists for. Confirmed on hardware — a legacy install leaves four shadowing files and no manifest:

```
  /usr overlay before restore: 4 files
    /overlay/root/usr/sbin/openwall
    /overlay/root/usr/sbin/setnetwork
    /overlay/root/usr/sbin/telegram
    /overlay/root/usr/sbin/updatewebui
```

Restore now falls back to the shared paths this project has ever installed, and only where the firmware has its own copy to reveal — an overlay file with nothing underneath it is somebody's own, not ours to delete. After the fix: `/usr` overlay 0 files, `/usr/sbin/updatewebui` back to the firmware's, `/var/www` identical to `/rom`.

### Regression, on the lab hi3516av300 from a clean firmware-only baseline

| | |
|---|---|
| `--help` / bad option | rc=0 / rc=1 |
| `--dry-run` | rc=0, overlay untouched |
| install | www 75, overlay 37, manifest 80 |
| re-run | `Already up to date.` |
| local edit | archived, `motor.cgi` back to the firmware's |
| install from GitHub master | rc=0, www 75 |
| pages (status, mj-settings, preview, fw-update, info-logs, tool-files, `j/pulse.cgi`) | 200 ×7 |
| `--restore` | www 72, `diff -r /var/www /rom/var/www` identical, `/usr` overlay 0 files |

PR #211, opened 2026-08-29
SD card: say when the card cannot be recorded to

### LINKED ISSUE #208: SD card: detect and surface a corrupt or read-only filesystem
The SD card page reports capacity, mountpoint, filesystem and usage, but it has no idea whether the card is actually **usable**. A card whose filesystem has gone bad reads as healthy right up until you notice nothing has been recorded for hours.

### What happened on a lab hi3516av300

Recording stopped silently at 12:04. Everything the UI shows still looked fine — card present, mounted, 18 GB free — and the only evidence was in `dmesg` and the majestic log:

```
FAT-fs (mmcblk0p1): error, corrupted directory (invalid entries)   (x12)
majestic: Cannot open file for writing: /mnt/mmcblk0p1/2026-08-29/16-02.mp4
```

The kernel had remounted it read-only, which `/proc/mounts` states plainly:

```
/dev/mmcblk0p1 /mnt/mmcblk0p1 vfat ro,relatime,... errors=remount-ro
```

`df` still showed 18 GB free, so the storage bar looked perfectly healthy. The recording that was in flight was left truncated and carries ~109 NAL errors. After an unmount the card would no longer mount at all (`mount: ... Invalid argument`).

### What the UI should say instead

1. **Read-only mount is the cheap, decisive signal.** `j/sdcard.cgi` already parses `/proc/mounts` to find the mountpoint and fs type; the option list is right there. If it contains `ro`, say so loudly — a card mounted read-only cannot record, whatever the free space says.
2. **Surface it on the Recordings page too**, not only under SD Card: "recording is on but the card is read-only" is the answer to *why is there nothing here*, and that is the page people will be looking at.
3. **Offer the repair that already exists.** The page has a `fsck` action; it is the right next step and should be suggested rather than waiting to be found.
4. Optionally flag recent filesystem errors from `dmesg` (`FAT-fs.*error`, `I/O error`, `EXT4-fs error`) as a corroborating detail.

### Note for whoever picks this up

`fsck.vfat` is **not present** on this firmware — only the generic `fsck` wrapper, which needs the fs-specific helper. So the existing "Check" action cannot repair a vfat card on a build like this one, and the UI should not imply it can. Worth checking what `sbin`/busybox actually ships before offering the button, and saying "reformat" when repair is unavailable.

openipc-ai:
Fixes #208.

## The problem

A card whose filesystem has gone bad reads as perfectly healthy in this UI right up until you notice nothing has been recorded for hours. The kernel mounts a vfat card with `errors=remount-ro` and drops it to read-only the moment the FAT stops making sense; `df` then goes on reporting the space that was free *at that instant*, so capacity, the storage bar and the clip list all look exactly as they did before recording stopped.

The only place the truth was written down was `/proc/mounts` — which `j/sdcard.cgi` already reads, for the mountpoint and the filesystem type, three fields away from the answer.

## What changed

**One judgement, made server-side.** `j/sdcard.cgi` gains a `health` field — `ok`, `readonly`, `unmounted`, `unreadable`, `unformatted`, `absent` — decided once and worded by whichever page is asking, rather than each page re-deriving it from `mounted`/`fs`/`partitioned` and drifting apart. `ro` is matched as a whole comma-separated word; a substring test finds the `ro` in `errors=remount-ro` on every healthy mount.

**The SD card page** badges the state, says plainly that nothing can be written however much space is free, marks the mount read-only, and stops claiming "✓ Recording to this card" when nothing is being written to it.

**The Recordings page** — the page people actually open when they wonder where the footage went — carries the same warning above the timeline in its own banner (`#rec-health`, separate from the transient per-clip `#rec-note`, which the player clears on every seek), folds the card's condition into the empty state so "Nothing recorded yet" is replaced by the real reason, stops labelling the truncated last clip as still "recording", and no longer paints a green **Recording** badge next to an archive that stopped growing hours ago.

**The repair is offered, and only when it exists.** busybox ships the generic `fsck` wrapper on every build, but it only execs `fsck.<fs>` — and a build without dosfstools has no `fsck.vfat` for it to find, which is the case on the lab hi3516av300 the issue was reported from. Where the helper exists the alert offers **Check and repair**; where it does not it says so and offers **Format**. `do_fsck` now remounts the card where it found it: taking a recording camera offline to repair it and then leaving it offline is a second outage.

**`fsErrors`** carries up to three matching `dmesg` lines as corroboration, gathered only in the unhealthy states — the endpoint is polled every 5 s and `dmesg` is 80 KB on a running camera. It is never presented as reassurance: on hi3516av300 the CMA allocator alone pushes a card error out of the ring buffer within minutes, so an empty list means nothing was found, not that nothing happened. The UI renders nothing rather than "no errors".

**Incidentally**, since these paths were being touched: `mount`, `unmount` and `fsck` now report what the tool actually said (`mount: mounting /dev/mmcblk0p1 on /mnt/mmcblk0p1 failed: Invalid argument` is the diagnosis; "mount failed" is not), and the op log renders with real newlines instead of a literal `\n` between every line (`L="${L}x\n"` does not interpret the escape in POSIX sh). The SD page's event handlers now delegate once from the container the page never replaces, instead of being re-bound on every 5-second render.

## Verified on hardware

An hi3516av300 with a 32 GB card, over every state the code distinguishes:

| state | result |
|---|---|
| healthy `rw` mount, options containing `errors=remount-ro` | `health:"ok"` — the substring trap avoided |
| remounted `ro` with footage already on the card | `health:"readonly"`, `availKb` still 30 GB — the misleading number the issue is about |
| superblock wiped, nothing will mount | `health:"unreadable"` |
| unmounted with an intact filesystem | `health:"unmounted"` |
| kernel messages injected via `/dev/kmsg` | both FAT-fs lines and the I/O error picked up |
| stand-in `fsck.vfat` installed | Check offered, runs, and the card comes back mounted **rw** |
| `POST op=mount` on a card with no superblock | error is `mount: … failed: Invalid argument` |
| SD page, 12 s of re-renders then one toggle | exactly one `POST /api/v1/config` — handlers bound once |

`npm test` passes. The camera was returned to its starting configuration afterwards (the stub removed from the overlay's upper dir, not through the merged path).

### What the read-only card looks like

**SD Card page** — badge reads **Read-only**; a danger alert states that nothing can be written, names `errors=remount-ro` as the cause, quotes the kernel lines it found, and offers Format (this build has no `fsck.vfat`); the Mount row reads `/mnt/mmcblk0p1 — read-only`; the storage line reads *29.1 GB free, but nothing can be written*; the Recording card reads *Configured to record here, but the card is read-only — nothing is being written.*

**Recordings page** — the same warning above the day picker with an **Open the SD card page** button, and the day badge reads **Cannot record** in red instead of a green **Recording**. Clips written before the failure still play.

Screenshots of all four states are attached in the issue thread.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -162,6 +162,10 @@ When you'd touch this code:
 
 Small `#!/bin/sh` scripts that emit JSON for the front-end. `pulse.cgi` is polled every 2s by `main.js:heartbeat` and fills the top bar (SoC temp, memory, overlay, uptime, day/night). `run.cgi` streams the output of a base64-encoded shell command (`cmd=` for trusted local, `web=` adds `timeout 3` for the console page).
 
+`sdcard.cgi` reports the card and runs its management ops, and it is read by two pages — `a/sdcard.js` and `a/recordings.js`. The judgement of whether a card is usable is made **once, server-side**, in its `health` field (`ok`, `readonly`, `unmounted`, `unreadable`, `unformatted`, `absent`); the pages only choose wording. `readonly` is the one that matters and the one nothing else can show: a card the kernel dropped to read-only (`errors=remount-ro`) still reports its old free space through `df`, so capacity, the storage bar and the clip list all read exactly as they did before recording stopped. `fsErrors` carries up to three matching `dmesg` lines and is gathered only in the unhealthy states — the endpoint is polled every 5 s and `dmesg` is ~80 KB on a running camera. It is corroboration only: the ring buffer is small and chatty enough that an error that stopped recording hours ago has usually scrolled out of it, so an empty list must never be rendered as a clean bill of health.
+
+`canFsck` exists because busybox ships the generic `fsck` wrapper on every build but it only execs `fsck.<fs>`, and a build without dosfstools has no `fsck.vfat` for it to find. The pages hide the Check action and say "reformat" instead when it is false, rather than offering a repair the firmware cannot perform.
+
 ### Front-end — `www/a/`
 
 - Pure JS, no framework. `$`/`$$` are `querySelector` wrappers. Don't introduce jQuery or any bundler — the README is explicit about keeping this small.
--- a/www/a/recordings.js
+++ b/www/a/recordings.js
@@ -405,14 +405,34 @@
 	// The second is why this is loaded before the page decides it has nothing
 	// to show: "nothing recorded yet" and "the card went read-only at 12:04"
 	// look identical from the clip list alone.
-	function loadCard() {
+	// `rec` asks the endpoint to du(1) the whole archive for the storage bar,
+	// which is not free on a card holding a month of clips — so the periodic
+	// refresh, which only needs the health verdict and df's numbers, leaves it
+	// off and carries the measured recBytes forward.
+	function loadCard(withRec) {
 		if (!state.prefix) return Promise.resolve();
-		return apiFetch('/cgi-bin/j/sdcard.cgi?rec=' + encodeURIComponent(state.prefix),
-			{ credentials: 'same-origin' })
+		const q = withRec === false ? '' : '?rec=' + encodeURIComponent(state.prefix);
+		return apiFetch('/cgi-bin/j/sdcard.cgi' + q, { credentials: 'same-origin' })
 			.then(function (r) { return r.json(); })
-			.then(function (d) { state.card = d; })
-			.catch(function () { /* the page is useful without it */ });
+			.then(function (d) {
+				// Asked without ?rec= the endpoint reports recBytes as 0 rather
+				// than leaving it out, so the carry-forward has to key off the
+				// question asked, not the answer given, or the storage bar's
+				// Recordings segment collapses on the first refresh.
+				if (d && withRec === false && state.card) d.recBytes = state.card.recBytes;
+				state.card = d;
+			})
+			// The page still browses the archive without this, but it must not
+			// go on claiming the camera is recording — see cardWritable().
+			.catch(function () { state.card = null; });
 	}
+
+	// Positive claims need positive evidence. A card is only known writable
+	// when the endpoint said so; a request that failed, or an answer this
+	// release does not understand, is an unknown card, and an unknown card
+	// must not be painted green — that false reassurance is the whole bug
+	// this page is here to stop telling.
+	function cardWritable() { return !!state.card && state.card.health === 'ok'; }
 
 	// Why there is no new footage, said on the page people actually arrive at.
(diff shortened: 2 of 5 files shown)


openipc-ai:
Both Qodo findings were real; verified against the code before touching it, and both are fixed in d083723.

**1. Unknown status appeared writable.** Correct — `loadCard()`'s catch left `state.card` null and `cardTrouble()` read null as "no trouble", so a 500 from the endpoint put the green **Recording** badge and the live-clip label straight back. That is the original bug wearing a different hat. Positive claims now require positive evidence: `cardWritable()` is true only when the endpoint actually answered `health: "ok"`, so an unrecognised answer counts as unknown too. Unknown gets a fourth badge state — grey, `Recording — card unknown`, with a title saying why — rather than a colour it has not earned. Playback is untouched, which is what the finding asked for.

**2. Card health never refreshed.** Also correct, and I had talked myself out of it on the grounds that this page polls nothing else. That was the wrong call: a card going read-only mid-session is precisely the failure the page exists to report, and reporting it only to whoever loads the page afterwards misses the person already watching it. It now re-reads the verdict every 30 s — slow on purpose, since this is a viewer that gets left open on a wall display — and **without `?rec=`**, so the refresh does not `du(1)` a month of clips every half minute. The measured `recBytes` is carried forward across refreshes; note that asked without `rec` the endpoint reports `0` rather than omitting the field, so the carry-forward keys off the question asked, not the answer given, or the storage bar's Recordings segment collapses on the first tick. Only the storage numbers repaint on every tick; the banner, day nav and clip list re-render only when the verdict itself changes.

While in there: the storage card now hides when there is nothing to show, instead of leaving a card's last known free space on screen after it was unmounted.

### Verified on the hi3516av300, both live

**Card goes read-only while the page is open** (probe sampled at load, then 46 s later, with the remount happening in between — no reload):

```
before: badge "Recording",     health hidden, 1 live label,
        storage "15 MB of 29.1 GB used · about 87 h 17 min of footage left"
after:  badge "Cannot record", health shown,  0 live labels,
        storage "17 MB of 29.1 GB used | Recordings 15 MB | Other 3 MB | Free 29.1 GB"
```

The projection withdrew (it is a promise about future writes), the Recordings segment survived the `rec`-less refresh, and the banner text is the read-only one.

**Endpoint forced to 500** (injected in the test proxy, camera untouched):

```
badge:      "Recording — card unknown"   class: badge text-bg-secondary
title:      "Recording is switched on, but the SD card's state could not be read"
liveLabels: 0     storageCardHidden: true     clipsRendered: 5
```

`npm test` still passes. The camera is back to its starting configuration.

Issue #114, opened 2026-06-15
Asset-size: custom Bootstrap rebuild (tree-shake the JS bundle / SCSS-only CSS)


widgetii:
Parking-lot of **next-step** asset-size optimizations to pull in **when a board runs out of rootfs space**. Not urgent — the current purge (#112) + minified-dist pipeline (#113) already restored comfortable headroom (e.g. the tightest 5120 KB-rootfs board sits at ~32 KB free). This issue is the head-start for the *next* squeeze.

## Why a minifier can't shrink `bootstrap.bundle.min.js`

Measured this cycle: re-minifying `bootstrap.bundle.min.js` with terser (even `unused,dead_code,toplevel,passes=3`) saved **312 B of 80,673** (~0; xz 22,100 → 22,068). It's already minified, and a minifier can't **tree-shake** a bundled IIFE — every component is internally referenced, so nothing looks unreachable. Cutting it needs a **bundler build from Bootstrap's ES-module source**, not a minifier.

The #113 dist pipeline (`package.json` + `tools/build-dist.sh`, terser/clean-css) is the right place to add this.

## Idea 1 — custom Bootstrap **JS** build (tree-shake)

Build a bundle that imports only the components the UI uses, via Rollup/esbuild from `bootstrap/js/src/*` + `@popperjs/core`.

Components actually used (grepped from `data-bs-*` usage):

| used | `data-bs-toggle`/attr |
|---|---|
| Dropdown | `dropdown` ×13 |
| Modal | `modal` (dismiss ×6, toggle ×1) |
| Collapse | `collapse` ×3 |
| Tab | `tab`/`pill`/`list` ×2 each |
| Tooltip | `tooltip` ×1 |
| Offcanvas | `offcanvas` ×1 |
| Button | `button` ×1 |
| ScrollSpy | `spy` ×1 |
| Carousel | `ride` ×1 |

**Not used:** Popover, Toast. Popper stays (Dropdown/Tooltip need it) — but a custom build can also tree-shake Popper to just the modifiers those two use.

Expected win is **bounded** (most components *are* used) — likely single-digit KB on flash from dropping Popover/Toast + trimming Popper. **Measure with a spike before committing** (`du`/`mksquashfs -b 128K -comp xz` delta is the real number; this session reproduced shipped squashfs sizes exactly that way). Confirm `carousel`/`scrollspy` are genuinely used and not stray matches.

## Idea 2 — Bootstrap **CSS** from SCSS modules

[...]

PR #217, opened 2026-08-29
Drop bootstrap.bundle.min.js: four behaviours in main.js, modals go native

openipc-ai:
The 80 KB bundle (with Popper inside) served exactly four behaviours:
four modals, seven nav dropdowns plus the file-manager row menu, the
navbar toggler, and dismissable alerts. Those are now ~90 lines in
main.js against the same markup and classes, so the purged Bootstrap
CSS keeps styling everything:

- Modals are native <dialog> elements. Escape, focus containment and
  centring come from the platform; a bootstrap.Modal-compatible shim
  keeps files.js and sdcard.js call sites unchanged, re-broadcasting
  the dialog's 'close' as the 'hidden.bs.modal' they listen for.
  Backdrop click closes only when the press started on the backdrop,
  so selecting text in an input and releasing outside does not throw
  the modal away. The .modal-header/-body/-footer classes stay; the
  --bs-modal-* tokens their rules read are declared on dialog.mj-modal
  in the override, since the .modal wrapper that used to define them
  is gone — including the padding-x/y pair whose absence silently
  discards the close button's push-right margin.
- Dropdowns are one delegated handler. data-bs-popper="static" turns
  on the pure-CSS placement rules Bootstrap itself uses when it skips
  Popper in navbars — which is every dropdown this UI has, so Popper
  was dead weight.
- The navbar toggler toggles .show; the slide animation is gone with
  the bundle.
- data-bs-dismiss handles alerts and modals from the same listener.

The one Tooltip lived on the unlinked FPV page; its badges keep their
title attributes, so the browser's own tooltip covers them.

Net: -80,673 B raw (-22 KB on the xz squashfs), main.js +3.7 KB,
override +2 KB.

Validated on hi3516av300 through an in-page probe driven by headless
chromium: 18 checks across open/close/exclusive/outside-click/Escape
for dropdowns, collapse, alert dismiss, and the full modal lifecycle
including hidden.bs.modal relay and the page's own New-folder flow;
screenshots of the open modal, open nav dropdown and expanded mobile
navbar match the bundle's rendering.

Part of #114.


openipc-ai:
Both review findings triaged:

1. **Arrow navigation** — real gap, fixed in the follow-up commit: ArrowDown/ArrowUp walk the enabled `.dropdown-item`s (entering at top/bottom from the toggle, clamping at the ends like Bootstrap), and Escape now restores focus to the toggle. Verified on hardware via the in-page probe (5 keyboard checks pass).

2. **`hide()` on a closed dialog throws `InvalidStateError`** — does not reproduce. The HTML spec defines `dialog.close()` as a no-op when the `open` attribute is absent (only `show()`/`showModal()` can throw `InvalidStateError`), and a headless-chromium test of `close()` on a closed dialog and double-`close()` confirms: no throw. The SD-format success path (`modal.hide(); load()`) is safe even if the user dismissed the dialog mid-format. No change made.

PR #220, opened 2026-08-30
SD card: make the format actually verify itself

openipc-ai:
## The report

From the lab camera after a nightly update: the SD Card page shows

```
Filesystem   unformatted
Mount        not mounted
```

Format → confirm → "Formatting… do not power off" → ten seconds → the fdisk banner dumped into the log, ending at `Calling ioctl() to re-read partition table`, and then:

```
Failed: partition node mmcblk0p1 was not created
```

Which names a symptom and offers nothing to do about it.

## What is actually wrong with that camera

That card has stopped storing data, in the quiet way. **No I/O error of any kind** — no `dmesg` line, no error counter, `/sys/block/mmcblk0/ro` is 0. It acknowledges every write, and every read of the same LBA returns different bytes:

```
# sector 100, three cold reads
75b14fc2e2a1eeeca92ef74fb054cfa9
3476cc7886b4bd4fcbd0ae4404a180a8
655e71bfda4a013fb403a9b508703e4e
# write zeros there, sync, drop caches, read back
66f322dfe1e90340e1f7fe72f6c74673   (all-zero 512B would be bf619eac0cdf…)
```

`fdisk` writes a table and reports success, the kernel re-reads it and finds nothing, `mmcblk0p1` never appears. The driver logged no CRC errors, so the garbage is what the card genuinely sent rather than a corrupted transfer.

## What is wrong with the WebUI

The message was the smaller half. **Nothing `do_format` did afterwards was verification.**

- `blkid "${DEV}p1" >/dev/null || err="format failed"` **could never fire** — busybox `blkid` exits 0 for a blank partition and for a device that does not exist at all. Dead code from the day it was written.
- **`mkfs`'s exit status was discarded**, so a genuine mkfs failure surfaced as whatever broke after it.
- **The partition table was never read back**, and the branch that would have written one is skipped whenever the kernel already lists `mmcblk0p1` — which it does while holding a partition from an earlier boot over a card that no longer has the table.

That third one is not hypothetical. It is the exact path on which this dead card formatted **successfully** during #211's testing last week and reported a fresh UUID it did not have.

## The change

**The decisive check is a write probe.** A stamp is put on the partition's first sector and read back cold *before* `mkfs` runs:

```sh
probe_write "${DEV}p1"
uncache
probe_seen "${DEV}p1" || { err="$UNSTORED"; return; }
```

That is a fact measured about the card rather than an inference from the exit status of whatever wrote to it. It is what catches a card gone **read-only in hardware** — the ordinary way an SD card ends its life, dropping the write without complaint while the filesystem already there reads back intact, so every test that only asks "is there a filesystem?" says yes.

Past the probe the card is known to keep writes, so **nothing after it blames the hardware**: an `mkfs` that exits 0 leaving no filesystem is reported as exactly that, and `blkid` is judged on its output rather than its exit status.

**A card out of step with the kernel is not a failing card.** A stale `mmcblk0p1` over a blank sector 0 is a reason to write a table, not a verdict — a healthy card whose table went out from under a mounted kernel needs it put back. So `do_format` looks, writes if what it found is not a table, looks again, and reaches the failing-card message only after a table has genuinely been written on this run and has not come back.

`mbr_ok` wants the `0x55AA` signature **and** a non-empty type byte in the first partition entry — a vfat boot sector ends in `0x55AA` too, so a bare filesystem would otherwise read as a partition table. It has three outcomes rather than two: table, no table, and *could not look*; empty reads go into the op log and the format carries on, because a card is never convicted on a test that did not happen.

Failure is also immediate instead of after `ensure_node`'s ten one-second retries. The verdict a genuinely failing card gets is:

> Failed: the card did not keep what was written to it, so it cannot be formatted here. The card (or the slot) is failing and needs replacing.

`uncache()` precedes each read-back **as defence in depth, not as the mechanism** — measured on target, this kernel invalidates a block device's page cache when its last opener closes it, so a fresh `dd` or `blkid` is cold already; that stops holding while something such as the automount keeps the device open. It drops pagecache only (`1`, not `3`) and notes in the op log when it cannot.

## Verified on hi3516av300

**Passing**

| case | result |
|---|---|
| fresh format of a blank device | `ok`, status `unformatted` → `ok` |
| two formats back to back, no delay | both `ok` — an earlier revision compared volume IDs and would have called the second a failing card |
| healthy device, stale kernel `p1` over a wiped sector 0 | recovers to `health: ok` — an earlier revision condemned it |

**Caught**

| case | reported as |
|---|---|
| the dead card | card is failing — in 1.5 s, where it was a 10 s dead end |
| card that silently drops the probe | card is failing |
| `mkfs` exits 1 with a message | `mkfs.vfat: device is too small` — its own words |
| `mkfs` exits 0 without writing, blank partition | `mkfs.vfat reported success but no filesystem was written` |
| `mkfs` exits 0 without writing, partition already formatted | same — the case a bare "is there a filesystem" test passes |

`mbr_ok`'s three states were checked separately, including the vfat-superfloppy shape that carries `0x55AA` with no partition entry behind it.

`npm test` passes; `dash -n` clean. No front-end change — `sdcard.js` already renders `r.error` verbatim.

The lab camera is running this build. Its card needs replacing.


openipc-ai:
Thanks — taking the three findings in turn.

**1. \`blkid\` accepts stale filesystem — real, fixed in \`abe8f81\`.** This is the better version of the bug the PR was already about. A non-empty \`blkid\` only proves *a* filesystem is there, and the failure mode that matters most in the field — a card that has gone read-only in hardware — drops the write silently and hands back the filesystem that was already present. \`do_format\` now captures the identity before \`mkfs\` and requires the cold read-back to be both non-empty **and different**:

```sh
id_was=$(blkid "${DEV}p1" 2>/dev/null)
...
id_now=$(blkid "${DEV}p1" 2>/dev/null)
[ -n "$id_now" ] && [ "$id_now" != "$id_was" ] || { err="$UNSTORED"; return; }
```

Verified both ways on hardware: a genuine reformat of an already-formatted partition moves the volume ID (`6A93-E29A` → `6A93-E29C`) and still passes, while an \`mkfs\` stubbed to exit 0 and leave the old filesystem in place is now caught.

I did not add a \`TYPE\` match against \`POST_fs\` as the prompt suggests, because busybox \`blkid\` does not report \`TYPE\` for vfat on this firmware — it prints \`UUID=\` alone. A check keyed on it would be inert here, which is exactly the failure this PR exists to remove.

**2 and 3 were already addressed in \`c457be7\`** and look like carryover from the review of \`69bd8f3\`:

- *mbr_ok skips existing partitions* — \`uncache\` and \`mbr_ok\` moved out of the \`if\` block. The branch now closes at line 180 and the check runs unconditionally at 185–186. This was a good catch: it is the specific path on which the dead card in the lab formatted "successfully" last week.
- *Cache eviction failures ignored* — \`uncache\` is now \`sync && echo 1 > … || logln "# note: could not flush caches, …"\`, so a failure to go cold is written into the op log the user sees rather than swallowed. It also drops pagecache only rather than dentries and inodes too.

One correction to the framing in my original description, which the alternatives section prompted me to measure: on this kernel the page cache is **not** what was hiding the failure. A block device's cache is invalidated when its last opener closes it, so a fresh \`dd\` or \`blkid\` after \`mkfs\` exits is already cold, and the dead card is caught with or without \`drop_caches\`. \`uncache\` earns its place only for the window where something (the automount) holds the device open across a check. The actual hole was that none of the three checks was a check.

openipc-ai:
All three new findings were real, and taking them together pointed at a better check than the one I had. `37fd002`.

**6 — Volume collision rejects format.** Correct, and this is the one that mattered most: a false "your card is failing" is the worst thing this endpoint can say, and my own comment admitted the hole rather than closing it. Comparing volume IDs made a good card's verdict depend on two `mkfs` runs not landing in the same second.

So the identity comparison is gone, replaced by a **write probe** — a stamp put on the partition's first sector and read back cold *before* `mkfs` runs:

```sh
probe_write "${DEV}p1"
uncache
probe_seen "${DEV}p1" || { err="$UNSTORED"; return; }
```

This is a fact measured about the card rather than an inference from the exit status of whatever wrote to it, so it has no timing dependency at all. It also catches the hardware-read-only case (finding 3, previous round) *earlier and more directly* than the identity check did — before a pointless `mkfs`, rather than by hoping two UUIDs differ. Two formats back to back with no delay now both pass.

Past the probe the card is known to keep writes, so nothing after it blames the hardware: `mkfs` exiting 0 without leaving a filesystem now reports `mkfs.vfat reported success but no filesystem was written`, not a verdict on the card.

**2 — Signature does not verify table.** Correct, and more reachable than the description suggests: a **vfat boot sector also ends in `0x55AA`**, so a card carrying a bare filesystem and no table at all passed a signature-only test. `mbr_ok` now requires the signature *and* a non-empty type byte in the first partition entry. Checked directly against a synthesised superfloppy shape (`55aa` at 510, zero at 450) — correctly rejected.

**1 — mbr_ok accepts unreadable MBR.** Fair; treating "could not look" as "looked and found nothing wrong" is the same class of error this PR exists to remove. `mbr_ok` is three-state now — 0 table, 1 no table, 2 the reads came back empty — and the caller logs the third into the op log the user sees and continues, rather than either passing silently or convicting a card on a test that did not run. The probe is what decides, and it still has to pass.

**3, 4, 5 are carryover** from the earlier passes and are fixed: 4 in `c457be7` (the check moved out of the conditional branch), 5 in `c457be7` (`uncache` now reports its own failure into the log), 3 as described above.

Nine cases verified on hi3516av300 — passing: fresh format, and two back-to-back formats with no delay. Caught: the dead card, a stale kernel partition over a wiped table, a card that silently drops the probe, `mkfs` exiting 1 (reports its own message), and `mkfs` exiting 0 without writing, on both a blank and an already-formatted partition.

openipc-ai:
**Missing MBR falsely condemns card — correct, and it caught me mis-scoring my own test.** Fixed in `35df52e`.

I had a case in the verification table reading *"stale kernel partition over a wiped table → caught"* and counted it as a win. It was run against a RAM disk — a perfectly healthy, writable device — and the right answer there is to put the partition table back and format it, not to tell the user their hardware needs replacing. Nothing had been written to the card on that path, so the code had established nothing about it.

`do_format` now looks before it writes, and only condemns after it has written:

```sh
uncache
mbr_ok "$DEV"; st=$?
if [ "$st" != 0 ] || ! grep -q 'mmcblk0p1$' /proc/partitions; then
    ...fdisk + partprobe...
    uncache
    mbr_ok "$DEV"; st=$?
fi
case "$st" in
    1) err="$UNSTORED"; return;;
    2) logln "# note: could not read the partition table back to check it";;
esac
```

A stale `mmcblk0p1` over a blank sector 0 is now a reason to write a table, not a verdict. Re-tested: that case recovers to `health: ok`, and the real dead card is still condemned — after a table has genuinely been written to it and failed to come back.

This also closes an edge I had not considered: on a card returning random bytes, sector 0 has a ~1/65536 chance of happening to read `55aa` at offset 510. The `|| ! grep -q 'mmcblk0p1$'` arm means a missing partition node forces the table write regardless, so that fluke can no longer skip `fdisk` and land in the "kernel never created the node" message.

Findings 2–7 are carryover from the earlier passes: 2 and 3 fixed in `37fd002` (three-state `mbr_ok`; signature *and* first-entry type byte, since a vfat boot sector also ends in `0x55AA`), 4 and 7 fixed in `37fd002` by replacing the volume-ID comparison with the write probe, 5 and 6 in `c457be7`.

Nine cases on hi3516av300. **Passing:** fresh format; two formats back to back with no delay; healthy device with stale kernel partition over a wiped sector 0 (recovers). **Caught:** the dead card; a card that silently drops the probe; `mkfs` exiting 1 (reports its own message); `mkfs` exiting 0 without writing, on a blank partition and on an already-formatted one. `mbr_ok`'s three states checked separately, including the vfat-superfloppy shape.

openipc-ai:
**Fixed PROBE can preexist — correct.** `c2940ea`.

A format that aborts between writing the stamp and finishing leaves the stamp on the card. With a constant string, the next run finds it there without having written it, so a card that had stopped taking writes in between reads as one that took this one. The failure is still caught downstream, but attributed to `mkfs` rather than to the card — the wrong thing to tell someone whose card is dying.

The stamp is minted per run now:

```sh
probe_new() {
    PROBE="MJ-PROBE $$ $(date +%s) $(dd if=/dev/urandom bs=6 count=1 2>/dev/null |
        od -An -tx1 | tr -d ' \n')"
}
```

Tested directly: write the old fixed stamp onto the partition, then format a device that drops writes. It now reports the card as failing, where before this change it reported `mkfs.vfat reported success but no filesystem was written`.

Findings 2–8 are carryover from earlier passes, all fixed: 2 in `35df52e` (a table is written and re-checked before any verdict), 3 and 4 in `37fd002` (three-state `mbr_ok`; signature *and* first-entry type byte), 5 and 8 in `37fd002` (the write probe replaced the volume-ID comparison), 6 and 7 in `c457be7`.

Ten cases on hi3516av300. **Passing:** fresh format; two formats back to back with no delay; healthy device with a stale kernel partition over a wiped sector 0 (recovers to `health: ok`). **Caught:** the dead card; a card that silently drops the probe; a stale stamp from an earlier run over a device that no longer takes writes (now attributed to the card); `mkfs` exiting 1 (its own message); `mkfs` exiting 0 without writing, on a blank partition and on an already-formatted one. `mbr_ok`'s three states checked separately, including the vfat-superfloppy shape.

PR #221, opened 2026-08-30
Nav: group the menu by task, not by which subsystem owns the page

openipc-ai:
## Why

Newcomers keep asking what **"Majestic"** in the top menu means — and the answer is "the settings menu": the most-used page in the UI sits behind a daemon codename. Around it, the bar is grouped by *which subsystem implements a page*, not by what the user is trying to do:

- two different **Settings** entries (Majestic→Settings vs Firmware→Settings, which is actually backup/maintenance);
- **Interface**, which is actually admin password + theme;
- **Configuration**, which is a raw YAML view;
- **Status** — the landing page — buried inside a dropdown;
- **Tools** a grab-bag, **Extensions** mixing notification pushers with VPNs;
- and no indication anywhere of which page you are on.

## What changed

The bar now reads by task:

```
Dashboard | Live | Recordings† | Camera ▾ | System ▾ | Services ▾ | Sign out

Camera ▾           System ▾                  Services ▾
  Settings           — Setup —                 — Notifications —
  Stream URLs        Network                   OpenWall
  Config file        Time                      Telegram
                     Access                    Ntfy
                     — Maintenance —           — Networking —
                     Update                    P2P network ↗
                     Backup & Restore          VTun
                     — Diagnostics —           WireGuard
                     Logs                      Proxy
                     Console
                     File Manager
                     SD Card†
```

† behind the existing `/dev/mmcblk0` guard, as before. Group labels are `.dropdown-header`s instead of bare dividers. **Every URL is unchanged** — bookmarks, `?tab=` deep links and the default-password redirect all still land.

- **Active-page highlighting, for the first time**: `main.js` marks the current page's entry (`.active` + `aria-current="page"`) and lights the parent dropdown toggle. The marker is an underline, not Bootstrap's active colour — that resolves to emphasis-black, on a light-theme bar painted brand indigo, where every link is already forced white.
- Page titles follow the new labels: Camera Settings, Stream URLs, Configuration File, Backup & Restore, Access Settings, Live View.
- `bootstrap.min.css` regenerated: `.dropdown-header` was purged before this markup existed; `dropdown-menu-lg-end` is purged now that nothing uses it; `.dropdown-divider` survives because the camera switcher still builds dividers dynamically.
- The shorter bar (43 chars of labels vs 59, two fewer carets) eases the documented ~1300px overflow; the stale "seven menus" comment updated.

## Verified on hardware

On the hi3516av300 lab camera (nightly 2026-08-29): rendered HTML via curl; headless Chromium probes confirming the active underline + `aria-current` on every page type and the lit parent toggle; both themes (the active-colour fix is specifically a light-theme issue); 1400px / 1000px / 390×844 portrait / 844×390 landscape with the tall System menu open — no horizontal overflow at any of them; keyboard navigation (arrows skip the headers, Escape closes) and the mobile collapse stack. `tools/lint-templates.sh` and `npm test` pass.


openipc-ai:
Re the review finding ("Submenu active style remains blue"): that is the intended design, not a missed selector.

The two levels signal differently on purpose. Top-level `.nav-link`s sit on the navbar, where the override already forces every link white (and the light-theme bar is brand indigo, so Bootstrap's emphasis-color active is unreadable) — colour cannot mark the current page there, hence the underline. Dropdown items sit on the menu surface, where Bootstrap's stock `.dropdown-item.active` (white on `--bs-dropdown-link-active-bg`) is the conventional, high-contrast current-item treatment — verified legible in both themes on the hi3516av300. Giving submenu items an underline instead would replace a stronger signal with a weaker one for consistency's sake.

So: underline = "this top-level entry is where you are", blue row = "this menu item is the page you're on", and the parent toggle's underline ties the two together when the menu is closed.

PR #223, opened 2026-08-30
Live View: put everything on the video, and fit the video to the screen

openipc-ai:
## Why

The Live page was assembled control-by-control, and it showed:

- **"WebRTC" rendered twice** — a checkbox-styled button and a ` · WebRTC` badge suffix, ~2 cm apart; the suffix vanished on MSE, so the transport was named exactly when redundant and absent when informative.
- **Stats displaced the picture** — the panel was an in-flow block *above* the `<video>`; opening it pushed a 4K frame below the fold.
- **Resolution without fps** — the WebRTC player computed fps every second and dropped it one function short of the badge.
- **A 90s control column** — a quarter of the row spent on three full-width text buttons, and on PTZ cameras a 3×3 emoji pad with an inline script: no keyboard, no accessible names, `mouseleave` stopped a pan mid-move, raw `fetch` that 401s silently forever after session expiry, unvalidated query params reaching a motor binary's argv.
- No fullscreen, no snapshot, layout jump at load, and a full-width 16:9 stage would have scrolled on a laptop anyway.

## What changed

The player is now a **stage** — the picture is the page, and everything lives on it:

- **Status chip** (top-right): `H264 3840×2160 · 25 fps`. fps is measured per-second over WebRTC (page-side `onStats` gate loosened; no player changes), configured `videoN.fps` on MSE, which measures nothing. Carries the connecting…/no signal/MJPEG states the badge held.
- **Segmented [WebRTC | MSE] picker** in the bar — the transport named exactly once, the alternative finally visible. Auto-fallback lights the MSE side; the failure reason still lands on the WebRTC label's tooltip. `preview-transport.js` semantics (pick/demote/expiry) untouched.
- **Stats as translucent overlay** (top-left) — stage geometry is bit-identical with the panel open (verified with `getBoundingClientRect`).
- **Auto-hiding control bar** — hover / focus / tap-toggle on touch; single row scrolled sideways on phones (a wrapped bar stacked three rows over a 220px stage). The `:focus-within` rule lives on the *bar*, not the stage: the stage is focusable to scope PTZ arrow keys, and a touch tap focuses it — a stage-level rule pinned the bar open.
- **Snapshot** (`/image.jpg` → timestamped download; hidden when the JPEG channel is off) and **fullscreen** (on the stage so all overlays come along; inline SVG glyph — U+26F6 is tofu in too many fonts; hidden on iOS Safari which lacks the API).
- **Viewport fit**: the stage reserves the stream's aspect ratio and clamps to `min(100%, calc((100dvh − 12.5rem) × aspect))`, centered — at 1470×830 (MacBook Air) the picture, bar and chips end 34px above the fold. No page `<h2>`: the nav underlines **Live**, and the caption's row is the row the player needed.
- **Night mode / IR filter / Light** as labelled chips under the stage (same ids, same wiring).

**PTZ, rewritten** (`p/motor.cgi` markup-only + new `preview-ptz.js` + hardened `j/ptz.cgi`): real `type=button`s with aria-labels; Pointer Events with `setPointerCapture` so a hold survives drifting off the button (verified: kept firing through a 120px drift, stopped on release); arrow-key panning while the stage itself has focus — the volume slider and radio groups keep their arrows (verified no leakage); `apiFetch` so a lapsed session redirects instead of 401ing at 4 Hz; `h`/`v` validated as small signed integers before reaching `gpio-motors`/`motor` argv. Same `?h=&v=` protocol.

## Structure

New `preview-hero.js` holds the DOM chrome because three test suites execute `preview-page.js` in a bare `vm`. `tests/staging.test.js` — which exists because this page has twice shipped a fallback burying a working player — is adapted to the segmented radios and still guards that invariant; `tests/auto-source.test.js` learns that Auto now measures the stage (the viewport clamp can narrow it below the column). `mj-settings`' live tab builds its own markup and is regression-checked untouched; `preview()` has exactly one caller, and CLAUDE.md's stale claim to the contrary is fixed.

## Verified on hardware

hi3516av300, live 4K: chip fps on both transports and across the picker; stats overlay displacing nothing; bar hover/focus/tap states; hold-with-drift and keyboard pan scoping (PTZ enabled via a transient sysinfo override); fit at 1470×830 and 1280×720; portrait + landscape phones with touch; both themes; talkback/audio controls; `j/ptz.cgi?h=abc&v=;reboot` neutralized to 0. `npm test` (4 suites) and template lint green.


openipc-ai:
All five findings addressed in d53c7bb — each one was real:

1. **PTZ range** — bound tightened to ±99 (`^-?[0-9]{1,2}$`). The pad sends ±5; two digits leave room for a coarser caller without letting one request command a four-digit sweep.
2. **Staging IDS** — `mj-stage`, `mj-note-close` (and `mj-player`) added to the staging suite's stub list, matching auto-source.
3. **MSE metadata lost on promotion** — the sharp one, and a latent flaw in the old badge too (the fps just made it visible). Each attachment now holds media reported before it was live and applies it the moment the swap promotes it, via the wrapped `onState`. Verified on hardware: chip reads `H264 3840×2160 · 20 fps` (configured rate) immediately after a WebRTC→MSE switch, stage ratio intact.
4. **Stats black-on-black in light theme** — correct: Bootstrap cells read `--bs-table-color`, not inherited colour. The overlay now sets the variable, and the muted key rule got the specificity to beat the cell rule that was silently repainting it. Verified: white values, `rgba(255,255,255,.7)` keys, light theme.
5. **Auto misses height-only resizes** — correct consequence of the viewport clamp: the stage narrows when only the viewport height changes, and `#mj-player` never sees it. The ResizeObserver now watches the stage; the auto-source suite asserts the new target.

All four test suites and lint green; re-verified live on the hi3516av300.

PR #225, opened 2026-08-30
Live View: when WebRTC serves the other channel, the chip says so

openipc-ai:
Follow-up to the report in #184 (and majestic#299): with WebRTC enabled, the reporter selected **Main** and got the **Sub** stream — and nothing on the page admitted it.

The stream request itself is correct and was verified on hardware (CDP WebSocket capture: `?stream=0`/`?stream=1` follow the radios, and the av300 serves the requested channel both ways — the wrong-channel behaviour is daemon-side, tracked in majestic#299). But WebRTC's `?stream=` is a preference, not an order, and the UI should not pretend otherwise: the chip now appends the served channel's name — `H264 800×600 · 15 fps · Sub stream` — whenever the frame recognisably belongs to the channel that was *not* selected.

Deliberately conservative: only when both channels' configured sizes are known and the frame matches the other one exactly; a camera with unset (sensor-native) sizes claims nothing, and MSE is exempt because `/ws/video` serves the number it is given or nothing. Covered in the vm suite (which gains a `transport` knob); verified live that an honestly-serving camera gets no note in either direction.

### DIFF
--- a/tests/auto-source.test.js
+++ b/tests/auto-source.test.js
@@ -319,9 +319,12 @@ const tick = (n) => new Promise((r) => setTimeout(r, n || 60));
 	{
 		// WebRTC takes ?stream= as a preference, and majestic#299 arrived as
 		// "Main selected, Sub displayed" with nothing on the page admitting
-		// it. With no remembered choice the page opens on Sub.
+		// it. With no remembered choice the page opens on Sub. Main's size is
+		// deliberately non-canonical: the settings page accepts and persists
+		// this form, and a size the preview refused to read would silence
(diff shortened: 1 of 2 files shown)


openipc-ai:
The review's finding is correct and fixed in the follow-up commit: the preview now parses `videoN.size` with the same case-insensitive, whitespace-tolerant grammar as mj-settings' `parseWH` — the settings page accepts and persists `1920 X 1080`, and the strict parser would have silenced both Auto and the served-channel note for exactly such a value. The vm test's mismatch case now feeds the non-canonical form on purpose.

PR #226, opened 2026-08-30
PTZ: one pad, three backends — and Pelco-D works out of the box

openipc-ai:
Addresses maintainer feedback (@flyrouter): two PTZ implementations cover 70–80% of hardware in the field — GPIO steppers via `gpio-motors`, and the community **Pelco-D** serial mod — and they should both be usable without hand-installing scripts over the WebUI's files.

## What was wrong

The Pelco-D mod lived as a curl-installed replacement of `p/motor.cgi`, and on current master that arrangement was broken three ways: `fw_setenv ptz true` alone never set `ptz_support` (the detection also demands `/usr/bin/motor`), so the pad gated itself off; the replaced pad leaned on `.btn-motor` CSS this tree no longer ships; and the next `updatewebui` would archive the local edit and reinstall ours, which speaks a protocol (`?h=&v=`) the camera has no binary for.

## What changed

- **Detection** (`common.cgi`) now names the backend — `gpio`, `motor`, or `pelco` (`ptz=` + `/usr/bin/btzoom`) — cached in sysinfo as `ptz_backend`. The stepped backends win when several are installed, since their protocol carries magnitudes the Pelco pulses cannot.
- **`p/motor.cgi`** draws the pad the backend deserves. Stepped: the eight-way pad as before. Pelco: what its users already know from the sandbox pad — four directions in a plus shape, a stop, zoom −/+ and the orange/blue focus diamonds — as real labelled buttons on the video stage.
- **`j/ptz.cgi`** gains `?act=<verb>` for Pelco next to the existing `?h=&v=`, matched against a closed whitelist (`up down left right stop wide tele near far`) — `btzoom` execs `pelcoD_$1`, so the verb must never pass through raw; `start`/`day`/`night` (lens maintenance) stay unreachable from HTTP.
- **`bin/btzoom` ships in the tree** (installs to `/usr/bin` like `ntfy.sh`), adopted from OpenIPC/sandbox `scripts/pelcoD` by @sansarus — so a Pelco camera needs `fw_setenv ptz true` and nothing else. Two changes from the original: POSIX function syntax (busybox ash does not promise the `function` keyword), and the serial port comes from U-Boot `ptz_port` (default `/dev/ttyAMA0`, which the original hardcoded — including inside `start`'s `stty`).
- **`preview-ptz.js`** reads what each button says (`data-dir` steps vs `data-act` verbs). A held Pelco button strings pulses end-to-end through the existing one-request-in-flight guard — `btzoom` answers only after its pulse ends, so pacing comes free. Stage-focused arrow keys resolve to whichever button the pad has for that direction, driving both protocols with the same keys.

## Verified on hardware

On the hi3516av300, using the maintainer's own fake-enable recipes (env vars + stub binaries logging their argv; no reboot — dropping `/tmp/webui/sysinfo.txt` suffices):

- gpio variant: eight-way pad renders, 550 ms hold = exactly three paced `gpio-motors 0 5 10` calls;
- pelco variant: both clusters render in the stage mount, clicks land `tele` / `near` / `stop`, ArrowUp lands `up`, and `act=day` and `act=;reboot` both bounce off the whitelist;
- `bin/btzoom` passes `sh -n` and `busybox ash -n`; templates lint clean, all test suites green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -222,11 +222,24 @@ The decisive check is `probe_write`/`probe_seen`: a stamp written to the partiti
   `preview-page.js` in a bare `vm` with a stubbed `$` over an `IDS` list: any
   new element preview-page.js touches must be `$`-guarded and, for coverage,
   added to both `IDS` lists. PTZ is `p/motor.cgi` (markup only, hidden) +
-  `preview-ptz.js`, which relocates the pad into the stage's `#mj-ptz` mount:
-  Pointer Events with capture for press-and-hold, arrow keys only while the
-  stage itself has focus (the bar's slider and radios own them otherwise),
-  `apiFetch` to `j/ptz.cgi?h=&v=` (which validates both as small signed
-  integers before they reach the motor binary's argv).
+  `preview-ptz.js`, which relocates the pad(s) into the stage's `#mj-ptz`
+  mount: Pointer Events with capture for press-and-hold, arrow keys only
+  while the stage itself has focus (the bar's slider and radios own them
+  otherwise), `apiFetch` to `j/ptz.cgi`. **Three PTZ backends**, detected in
+  `common.cgi:update_caminfo` into `ptz_backend` (cached in sysinfo like
+  `ptz_support`): `gpio` (U-Boot `gpio_motors` + `gpio-motors` binary) and
+  `motor` (`ptz` + `/usr/bin/motor`) are stepped eight-way pads speaking
+  `j/ptz.cgi?h=&v=` (validated as small signed ints); `pelco` (`ptz` +
+  `/usr/bin/btzoom`) is the community Pelco-D serial mod — four directions,
+  zoom and focus, each a fixed timed pulse — speaking `j/ptz.cgi?act=<verb>`
+  against a closed whitelist (btzoom execs `pelcoD_$1`, so the verb must
+  never pass through raw). `bin/btzoom` ships in this repo (adopted from
+  OpenIPC/sandbox `scripts/pelcoD`, port from U-Boot `ptz_port`, default
+  `/dev/ttyAMA0`) so a Pelco camera needs only `fw_setenv ptz true`. A held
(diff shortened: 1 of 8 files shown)


openipc-ai:
All four findings were real — fixed in the follow-up commit, each verified on the hi3516av300:

1. **Stop dropped mid-pulse** — worse than reported, in fact: the in-flight guard it was being dropped by wasn't even pacing anything, because `j/ptz.cgi` answers 200 *before* exec'ing and `fetch` resolves on headers, so a held button stacked pulses at 4 Hz behind the port. Two fixes: the request now awaits the response **body** (closes when the CGI and its pulse end — measured: a held button pulses at ~750 ms end-to-end), and stop queues past the guard, dispatching the moment the current request answers, after killing any live hold. Measured: a stop clicked mid-pulse lands last, nothing follows it.
2. **UART never configured** — `btzoom` now runs the cheap idempotent `stty` before every command; the long lens-wakeup/ICR sequence stays in `start` as a manual maintenance action, still unreachable over HTTP.
3. **Stale sysinfo after upgrade** — `updatewebui` drops `/tmp/webui/sysinfo.txt` after an install; the cache was written by the release just replaced, and a camera that gained `btzoom` from that very install would have hidden its pad until reboot.
4. **Concurrent serial writes** — `btzoom` takes a device-wide `mkdir` lock (bounded ~2 s wait, stale locks broken after a minute). Measured: three concurrent moves serialized (~2 s vs ~1 s raced); a lock planted with a 2020 timestamp was broken and the command ran.

openipc-ai:
For the record on the re-review's two unmarked findings: both fixes are in head 0043e5f — the pre-dispatch `stty` runs before every command (bin/btzoom:51; the hunk the re-review quotes is the untouched `pelcoD_start` copy), and the device-wide mkdir lock with bounded wait, stale-break and EXIT trap is at bin/btzoom:17-44. Both measured on hardware: three concurrent moves serialize (~2 s vs ~1 s raced), a lock planted with a 2020 timestamp is broken open and the command runs.

PR #228, opened 2026-08-30
Live View: bring back the side rail — motor.cgi is a plugin point

openipc-ai:
Follow-up to the report in #184 ([comment](https://github.com/OpenIPC/majestic-webui/issues/184#issuecomment-5468231569)): the hero redesign broke custom `p/motor.cgi` replacements — a de-facto plugin point users fill with their own control sets built for the old ~300px Controls column. Their content landed full-width under the video, and the `.btn-motor` class it leans on was gone.

The reporter also pointed at the space to pay for the fix: the stage's viewport clamp usually leaves the rail's width free. So:

- **The side rail is back**: Night mode / IR filter / Light stacked as the old Controls card had them, then whatever `motor.cgi` renders, then Stream URLs. The stage yields width to the rail like the old `col-9`/`col-3` split; the viewport clamp still caps its height; on narrow screens the rail wraps below **at its own width** (max 19rem) instead of stretching stacked buttons across the page.
- **`.btn-motor` restored** for the replacements in the field.
- **Stock pads unaffected** — they're hidden markup that `preview-ptz.js` lifts onto the video, so with stock files the rail holds three toggles and a link.

Verified on the hi3516av300 at the reporter's viewport (1340×708): stock chips beside the stage; **the reporter's actual `motor.cgi`** (taken from his report) rendering in the rail at its old width — full-rail lighting buttons, 2.3rem digit row, its status-bar injection working, zero console errors; the stock Pelco pads still relocating onto the stage with the rail present; mobile wrap at the rail's own width. Purged CSS regenerated (`d-grid` returns to the subset).


openipc-ai:
Rejected by the maintainer: the wide full-bleed design is the design of record for the Live page, and reintroducing a side column — even for motor.cgi plugin compatibility — is not the direction. Custom motor.cgi content should adapt to the wide layout instead; guidance going to #184.

PR #229, opened 2026-08-30
SD card: land a blank card in the archive, in one press

widgetii:
Walking the "insert an empty SD card, get recordings" journey on a t31 nightly found two separate things wrong. Both are fixed here, and both were reproduced and verified on that camera.

## The format did not survive a blank card

#220 made the format verify itself; it did not make it survive the hotplug path. On a genuinely blank card the format failed on roughly **one attempt in four**, with a different message each time and every one of them wrong about the cause.

One cause underneath all of them: `/proc/sys/kernel/hotplug` is `/sbin/mdev`, and `mdev.conf` runs `/lib/mdev/automount.sh` for `mmcblk[0-9]p[0-9]`. That script's `my_umount` unmounts the card **and `rmdir`s the mountpoint**. Every partition-table re-read — `fdisk`'s, and again when `mkfs` closes the device — emits a remove/add pair, so throughout a format both `/dev/mmcblk0p1` and `/mnt/mmcblk0p1` blink out and back. `dmesg` shows the tell: `mmcblk0: p1` twice, 23 ms apart.

Each step failed in its own voice, and each blamed the card:

| what the user saw | what was actually happening |
|---|---|
| `mkfs.vfat: image is too small` | about a 59 GB card — mkfs opened a partition momentarily at length 0 |
| `mount … failed: No such file or directory` | ENOENT on the mountpoint automount had just removed |
| `mkfs reported success but no filesystem was written` | the post-mkfs `blkid` landed in the gap; busybox blkid prints nothing for an absent device |
| **"the card did not keep what was written to it … needs replacing"** | the write probe reading back empty, or reading the card's own pre-existing random bytes after a node swap |

That last one is the harshest thing this endpoint can say, and it was saying it to people holding good cards.

Worst of all, `ensure_node` — the function meant to recover from this — called `mdev -s`, which rescans every device in `/sys` and re-runs the hotplug rules, i.e. re-runs `automount.sh`. **The recovery was a generator of the churn**, and each retry made the next step likelier to fail.

The fixes, in one sentence each:

- `ensure_node` makes the node itself with `mknod` from the major:minor already in `/sys`, never `mdev -s`; when `/sys` has no partition at all it lets go of the card and re-runs `partprobe`, which can have failed with EBUSY while the automount held the disk.
- `node_ready` replaces `[ -b … ]` — a filename existing is not a partition being usable, so it reads the partition's **last sector** to prove the kernel has it at full length (no `blockdev` on OpenIPC).
- `probe_card` gets `mbr_ok`'s three outcomes and needs **three** mismatches before convicting: an empty read is never evidence, and one mismatch can be a read across a node swap, while a card that truly refuses writes fails every attempt.
- `mkfs` and `mount` retry, matching only transient complaints; anything else is a real refusal reported first time.
- the partition table is written **unconditionally** — format means erase, and the guard that skipped it read sector 0, which can come from the page cache.
- **mounting is the proof** a filesystem was written; the raw-device checks are demoted to diagnosing a mount that failed, once the retries have spent themselves.

## Four pages and six clicks to switch recording on

Inserting an empty card and wanting recordings meant: Recordings (told the card is not formatted) → SD card page → Format → back to Recordings for the switch → settings if the path did not match where the card mounted. None of those steps is a choice the person makes.

The empty state now carries the action:

> The SD card is not formatted — nothing is being recorded.
> **[ Format the card and start recording ]**

One press does the lot: format, mount, point `records.path` at **where the card actually mounted**, switch recording on. That last part is easy to miss — a card with no partition table mounts at `/mnt/mmcblk0` while `records.path` still reads `/mnt/mmcblk0p1`, so setting `enabled` alone arms a recorder aimed at the overlay, which majestic refuses outright, leaving recording "on" and nothing written.

Offered only where pressing it can work: `unformatted`/`unreadable` get the destructive version behind a confirm, `unmounted` gets a plain Mount with no confirm, and `readonly`/`absent` get no button at all — neither is mended by pressing anything here. The filesystem is not asked for either, since `mkfs` is `["vfat"]` on every build seen and choosing between one option is not a choice.

Also: the wait for the first clip now admits when it has stopped being a wait. "Nothing recorded yet" was repeated for as long as the tab stayed open. Once the backoff is spent it says so and names the two things that produce it, without picking one — from the page they look identical, and the card has already answered healthy. Polling continues underneath, so a merely-slow recorder still lands its clip and the page moves on.

## Verification

All on the reported t31 (`nightly-20260830-2b619b9`), card filled with 32 MiB of `/dev/urandom` and rebooted, so the kernel reports `unknown partition table`, no `mmcblk0p1`, nothing mounted.

- **Format: 15/15 clean**, against 11/15 before, and no run reaches the failing-card verdict. Each of the four messages above was reproduced first and is gone.
- **Journey**: one press, one confirm, page showing clips **25 seconds later**, majestic's pid unchanged from boot.
- `npm test` passes; `tools/lint-templates.sh` clean; `bootstrap.min.css` regenerated and unchanged (every class used is already in the purged set).

## Note

The final step — the clip actually appearing — also needs the runtime `records.enabled` fix on the majestic side, without which enabling recording via the API writes the config and starts no recorder on Ingenic. That is a separate change in the daemon; this PR stands on its own for the format and the journey, and the "stopped being a wait" note above is what an older daemon looks like from here.

### DIFF
--- a/www/a/recordings.js
+++ b/www/a/recordings.js
@@ -1362,15 +1362,34 @@
 	// still reads /mnt/mmcblk0p1, so setting `enabled` alone would arm a
 	// recorder aimed at a directory on the overlay — which majestic refuses
 	// outright — leaving recording on and nothing written.
+	//
+	// Which is why the mountpoint is demanded rather than used if present.
+	// loadCard() swallows its own failure and resolves with state.card null, so
+	// "no mountpoint" here means the card was not asked about successfully — and
+	// enabling anyway would leave records.path at whatever it was, recreating
+	// precisely the state this exists to prevent.
+	//
+	// Resolves to '' when recording is really on, and to a reason when it is
+	// not; a write that was refused must never be reported as one that landed.
 	function useCardForRecording() {
 		return loadCard().then(function () {
-			const mp = state.card && state.card.mountpoint;
-			const rec = { enabled: 'true' };
-			if (mp) rec.path = mp + '/%F';
+			const d = state.card;
+			if (!d || !d.mounted || !d.mountpoint)
+				return 'The card was prepared, but the camera did not report where it ' +
+					'mounted, so recording was left off.';
 			return apiFetch('/api/v1/config', {
 				method: 'POST', credentials: 'same-origin',
 				headers: { 'Content-Type': 'application/json' },
-				body: JSON.stringify({ records: rec }),
+				body: JSON.stringify({
+					records: { enabled: 'true', path: d.mountpoint + '/%F' },
+				}),
+			}).then(function (r) {
+				if (r.ok) return '';
+				// Plain text, not markup: this is read out through alert().
+				return r.text().then(function (t) {
+					return 'The card is ready, but the camera refused to start recording' +
+						(t ? ': ' + t.slice(0, 160) : ' (HTTP ' + r.status + ').');
+				});
 			});
 		});
 	}
@@ -1398,10 +1417,19 @@
 						return;
 					}
 					say('<span class="text-secondary">Starting recording…</span>');
-					return useCardForRecording().then(function () {
+					return useCardForRecording().then(function (why) {
+						// The card really was prepared, so the page has to re-read
+						// itself either way: everything it currently says about the
+						// card is now out of date. Which is also why the failure
+						// cannot be reported in place — the re-read owns #rec-note
+						// and watchCard() refreshes it every few seconds, so a line
+						// written there is wiped almost immediately. alert() is
+						// what sdcard.js already uses to report a failed op, and it
+						// is the one surface a re-render cannot take away.
+						if (why) alert(why);
 						waitStep = 0;
 						stalled = false;
-						boot();
+						return boot();
 					});
 				})
 				.catch(function () {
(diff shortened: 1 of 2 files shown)


widgetii:
All three addressed in 7baeab5, each reproduced first.

**1. Config failures treated as success** — `useCardForRecording()` now resolves to `''` or a reason, like `setRecording()` beside it. Proved by stubbing `/api/v1/config` to 500 on the t31 with the card unmounted: the user is told *"The card is ready, but the camera refused to start recording: …"* and `config.json` still reads `records.enabled: false`. Before the fix the page reported recording as on and polled for clips that were never coming.

**2. Card refresh failure enables stale path** — the mountpoint is demanded rather than used-if-present, since `loadCard()` swallows its own failure and resolves with `state.card` null.

Reporting that failure took two attempts, which is worth recording: the first version appended to `#rec-note`, and the message survived under three seconds — the page must re-read itself after the prepare, and `watchCard()` refreshes that element every five seconds. `alert()` is what `sdcard.js` already uses for a failed op, and it is the one surface a re-render cannot take away.

**3. Nested retries share counter** — correct, and it mattered: `ensure_node` resets `i` to 0 inside each caller's loop, so the retries added to ride out the hotplug churn could be spent by the settling they were waiting on. Every loop carries its own counter now. The `i` in `json_errs` is left alone — pipe subshell, not on the format path.

Re-verified on the reported t31: format 15/15 from blank after the counter fix, and the full blank-card journey still reaches clips in 20 s.

Issue #227, opened 2026-08-30
Feedback on testing the integrated PTZ function and some suggestions for improvement.


flyrouter:
@openipc-ai 

Feedback and suggestions are basically based on the results from https://github.com/OpenIPC/majestic-webui/commit/08a2909f664d19c05541040cf11d7e9b02612ecd

1. In the script /usr/bin/btzoom the port speed is hardcoded to 115200. However, there are many devices with lower speeds. It's suggested to add a ptz_speed variable so you can set popular speeds like 115200, 9600 and others. By default, if nothing is specified, you can leave it at 115200.

2. Maybe we should reconsider the whole naming structure for control options related to PTZ that are stored in ENV. First, there’s the method we’ll use to operate the device. In the current implementation, we have direct "gpio" (through the gpio-motors utility) and via some microcontroller, for example "pelco-d". Since there could be various other options like "visca", "i2c", "spi" and others, or named after specific chips, it makes sense to have an ENV option, for example called "ptz_control", which would indicate the specific method and way of controlling. Currently it’s either "gpio" (direct control through the motor driver) or "pelco-d" via a microcontroller. If ptz_control (or any other name) is not defined, or set to "none", the PTZ buttons won’t show and none of the procedures are called. Also, the proposed "ptz_control" could replace the current check for the availability of the ENV option "ptz true" which doesn’t give specifics.

3. Replace the 'gpio_motors' option used in ENV with 'ptz_gpio'. This is just to tidy up the naming.

The final options might look something like this (discussions welcome):

```
fw_setenv ptz_control {gpio, pelco-d, none, and other}
fw_setenv ptz_port {/dev/ttyAMA0, /dev/ttyAMA1, /dev/ttyS2, and other}   # for pelco-d method
fw_setenv ptz_speed {115200, 9600, and other}                           # for pelco-d method
fw_setenv ptz_gpio H1 H2 H3 H4 V1 V2 V3 V4             # for gpio control method, replaces the gpio_motors option
```

openipc-ai:
@flyrouter — all three points implemented and merged (#232), each verified on an hi3516av300 with your no-hardware recipes.

## The scheme, as landed

```
fw_setenv ptz_control {gpio | pelco-d | motor | none}
fw_setenv ptz_port  /dev/ttyAMA0      # pelco-d; default /dev/ttyAMA0
fw_setenv ptz_speed 9600              # pelco-d; standard rates, default 115200
[...]

PR #232, opened 2026-08-30
PTZ: ptz_control names the method, ptz_speed sets the rate (#227)

openipc-ai:
Implements all three points of #227:

1. **`ptz_speed`** — feeds both `stty` calls in `btzoom`, validated against the standard rates (1200–115200), defaulting to the old 115200; a bogus value falls back rather than killing the command.
2. **`ptz_control`** — the explicit method switch: `gpio`, `pelco-d`, `motor`, `none`. An explicit method is trusted but still requires its binary; `none` disables PTZ regardless of any legacy configuration; **unset falls back to the pre-#227 detection unchanged**, so no field camera loses its pad. `j/ptz.cgi` honours the same switch, so pad and endpoint cannot disagree.
3. **`ptz_gpio`** — accepted as the tidier name, with one caveat the WebUI cannot fix alone: `gpio-motors.c:120` runs `fw_printenv -n gpio_motors` itself for the pin list, so that variable still has to be set until the firmware utility learns the new name (one-hunk patch proposed in the issue).

Also: the motor-profile backend (not covered in the issue) slots in as `ptz_control=motor` with the profile in `ptz_profile`, falling back to the legacy `ptz` value where it doubled as the profile name.

Verified on the hi3516av300: `pelco-d` enabled by `ptz_control` alone (no legacy `ptz`); `stty` observed at `ptz_speed=9600` and back at 115200 for a bogus rate; `gpio` enabled by `ptz_control` alone (no legacy `gpio_motors`); `ptz_control=none` showing no pad and refusing the API with every legacy variable and binary present; legacy-only configuration untouched. Tests and template lint green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -234,15 +234,22 @@ The decisive check is `probe_write`/`probe_seen`: a stamp written to the partiti
   while the stage itself has focus (the bar's slider and radios own them
   otherwise), `apiFetch` to `j/ptz.cgi`. **Three PTZ backends**, detected in
   `common.cgi:update_caminfo` into `ptz_backend` (cached in sysinfo like
-  `ptz_support`): `gpio` (U-Boot `gpio_motors` + `gpio-motors` binary) and
-  `motor` (`ptz` + `/usr/bin/motor`) are stepped eight-way pads speaking
-  `j/ptz.cgi?h=&v=` (validated as small signed ints); `pelco` (`ptz` +
-  `/usr/bin/btzoom`) is the community Pelco-D serial mod — four directions,
-  zoom and focus, each a fixed timed pulse — speaking `j/ptz.cgi?act=<verb>`
-  against a closed whitelist (btzoom execs `pelcoD_$1`, so the verb must
-  never pass through raw). `bin/btzoom` ships in this repo (adopted from
-  OpenIPC/sandbox `scripts/pelcoD`, port from U-Boot `ptz_port`, default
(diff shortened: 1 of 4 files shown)


openipc-ai:
Findings 2 and 3 were real and are fixed in the follow-up commit: an explicit `ptz_control` method now requires its configuration, not just its binary — `gpio` needs a pin list (`ptz_gpio`, or the legacy `gpio_motors` the binary itself still reads; this also makes `ptz_gpio` genuinely read, as documented), `motor` needs a profile (`ptz_profile`, legacy `ptz` fallback). Verified on hardware in both directions for both methods.

Finding 1 is declined as a deliberate, documented decision rather than a gap: an unset `ptz_control` keeps the pre-#227 legacy detection so no field camera — including ones set up from the maintainer's own published recipes — loses its pad on upgrade. The issue's literal "not defined → no buttons" would be a silent hard cutover; whether to make that cutover is put to the maintainer explicitly in the #227 reply, where that decision belongs.

openipc-ai:
Re-review decisions: **Finding 2** is stale against its own head — `ptz_gpio` is read executably at 70817a21b in both `common.cgi` (gpio detection) and `j/ptz.cgi` (gpio dispatch); the residual — the `gpio-motors` binary itself reading the legacy name for its pin list — is cross-repo (OpenIPC/firmware, `gpio-motors.c:120`) and the one-hunk patch is proposed to the maintainer in #227. **Finding 1** stands declined as the documented compatibility decision, explicitly delegated to the maintainer in the same reply. Finding 3 from the first round is confirmed fixed by the re-review.

PR #233, opened 2026-08-30
Live View: the toast says when the shared encoder moved, and for whom

openipc-ai:
WebRTC's bitrate adaptation drives one encoder per channel, and majestic moves it to the lowest estimate among that channel's viewers — so a drop on your screen is not necessarily your link, and a drop you caused is not only your problem. Until now nothing on the page could tell those apart: a standing note disclosed that adaptation happens, always, in the same words, whether or not anything was happening.

**The toast replaces the note with the event itself.** `preview-adapt.js` keys off the `enc=` counter majestic now sends in the per-second `/ws/webrtc` stats line — where the arbitration actually holds the encoder, `0` meaning "at the configured rate" — rather than inferring from received throughput, which breathes with scene complexity and would cry wolf on every busy frame. When `enc=` steps, the toast names both rates, the direction (green up, amber down), and whose connection did it:

- our own `remb=` at or near the new rate → *"Matched to your connection — every viewer of this stream now receives this rate."*
- our estimate well above it → *"Another viewer's connection can't keep up — the stream follows the slowest link watching it."*
- up-steps → *"Raised — the slowest connection watching improved."* / *"Back at the configured rate."*

The baseline is adopted silently on join, channel change and transport return — a state you arrive into is not an event you witnessed. Hover pins the toast, click dismisses it, a new step replaces it. The stats panel gains an **encoder** row from the same counter. On a majestic without `enc=` the toast stays dormant and the row shows `-`: nothing here guesses.

**The standing disclosure note goes** (`#mj-transport-note`, its dismissal wiring, the `adapts[]` config read): it fired whether or not anything was adapting — the kind of disclosure people learn to ignore — and said "your connection" even when the encoder was following someone else's. The one thing it did besides disclose, restoring the transport toggle's tooltip after a failure, moves to `settle()`.

Verified on a hi3516av300 against two live Chromium viewers, one behind a 600 kbit/s policed link. The clean viewer's timeline:

```
TOAST: 4 Mbit/s ↘ 1 Mbit/s    Another viewer's connection can't keep up —
                              the stream follows the slowest link watching it.
  (throttled viewer leaves)
TOAST: 1 Mbit/s ↗ 4.1 Mbit/s  Raised — the slowest connection watching improved.
```

while the throttled viewer itself saw its own arrival as *"Matched to your connection"*. `npm test` passes; the vm-driven page tests needed no new stubs (the module is its own file, every hook in `preview-page.js` is guarded), and the retired note's ids left the `IDS` lists.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -219,8 +219,14 @@ The decisive check is `probe_write`/`probe_seen`: a stamp written to the partiti
   nothing ever displaces the picture: the status chip `#mj-badge` top-right
   (`CODEC W×H · fps`; fps is live WebRTC stats via a loosened `onStats` gate,
   or the configured `videoN.fps` on MSE, which measures nothing), the stats
-  panel top-left as translucent glass, the adaptation note (dismissible)
-  above the bar, and the auto-hiding control bar (`.mj-bar`: shown on hover /
+  panel top-left as translucent glass, the adaptation toast `#mj-adapt`
+  under the chip (`preview-adapt.js` — event-driven off the camera's `enc=`
+  stats counter, it reports the moment the shared encoder's bitrate moved,
+  in which direction, and whose connection moved it; it replaced a standing
+  "the camera is adapting" disclosure note that fired whether or not
+  anything was happening and could not name the responsible link; on a
+  majestic without the counter it stays dormant), and the auto-hiding
+  control bar (`.mj-bar`: shown on hover /
   `:focus-within` / a JS `.mj-show` tap-toggle). `preview-hero.js` owns the
   stage chrome — bar visibility, fullscreen (on the stage; hidden on iOS
   Safari which lacks the API), snapshot (`/image.jpg` → blob download, shown
--- a/tests/auto-source.test.js
+++ b/tests/auto-source.test.js
@@ -21,7 +21,7 @@ const IDS = [
 	'mj-stats-btn', 'mj-stats-ctl', 'mj-stream-0', 'mj-stream-1',
 	'mj-stream-auto', 'mj-auto', 'mj-sub', 'mj-talk', 'mj-talk-ctl',
 	'mj-talk-lbl', 'mj-transport-w', 'mj-transport-m', 'mj-transport-ctl',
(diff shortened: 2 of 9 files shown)


openipc-ai:
Review addressed, point by point:

**2 — wrong fallback-channel bitrate: fixed.** The served-channel detection the chip already used is now a shared `servedStream()`, and the toast's tick carries both the served channel's `cfgKbps` and the channel itself. Confirmed live on the lab camera in the served-mismatch state (chip reading "· Main stream" against a Sub request): the toast named Main's configured 4.1 Mbit/s as the endpoint, not Sub's 1 Mbit/s.

**3 — stale baseline across reconnects: fixed where it was wrong, kept where it was right.** The baseline is now keyed on the channel: a reconnect (or served-channel flip) that lands the ticks on the *other* encoder re-adopts silently instead of comparing across two encoders. A reconnect that stays on the *same* channel keeps its baseline deliberately — the encoder is the same encoder, and a step it took during a two-second renegotiation gap is a real step the viewer effectively watched happen; swallowing it would hide exactly the events this toast exists for. From/to are accurate in that case by construction (last seen → current).

**4 — 0↔configured step consumed before the config lands: fixed.** `tick()` now holds the baseline (returns without advancing) when a step touches 0 and `cfgKbps` is still unknown — the step stays pending and is announced on the tick after the config arrives, instead of being silently consumed by the baseline advance.

**5 — accessibility: fixed.** The toast is `role="status"` so the announcement reaches screen readers when the text lands; a real `×` button (corner-anchored, `aria-label="Dismiss"`) gives the keyboard the dismissal that click-anywhere only gave the pointer; focusing into the toast pins it exactly as hovering does; and if the toast hides while the × holds focus, focus returns to the stage rather than falling to the body.

**1 — toast ids in the vm-test `IDS` lists: declined, with reasons.** The rule covers elements *preview-page.js* touches, because those tests execute preview-page.js and an unstubbed lookup would crash them. The toast elements are owned by `preview-adapt.js`, which the vm tests do not load — same pattern as the stats-panel row ids (`mj-st-*`), which have never been in the lists; preview-page.js's only new lookups are behind the same guards the tests already exercise, and the suite passes.

All 247 checks pass; re-verified against the live camera after the fixes.

Issue #235, opened 2026-08-30
Live adjustments wastes ~60% of the page, and gives runtime state and saved settings the same weight


widgetii:
`mj-settings.cgi?tab=live` is the panel you tune a camera's picture from, and it spends most of the screen it is given on nothing.

Measured on an hi3516av300 (imx415, 2.6.08.29-lite) at a 1440px viewport, with `getBoundingClientRect` / `getComputedStyle` rather than by eye:

- The form column is **990px** wide. The preview renders at **520×292**, because `.mj-live-video` carries `max-width: 640px` and a hard-coded `aspect-ratio: 16/9`. Below y≈600 the column is empty — roughly **60% of the content area holds nothing**.
- A sensor that is not 16:9 gets squeezed into that box rather than letterboxed.
- Six emoji label the knobs (`☀ 🌗 💧 🌈 ⇄ ⇅`) and three more the toggles (`🌙 👁 💡`). Emoji are not a typeface we control: **the eye on the IR filter button has no glyph in the stack the camera ships and renders as an empty box.**
- The three runtime toggles use Bootstrap's default `#0d6efd` rather than the brand indigo, and they are the same shape and weight as the sliders beside them.

That last one is the substantive problem, not a cosmetic one. **Night mode, IR-cut and the lamp change the camera for every viewer the instant they are pressed. Brightness stages a value that Save writes.** They are two different kinds of thing and the panel presents them identically, so there is nothing on screen to tell you which of the controls in front of you has already taken effect.

Two things the panel also cannot answer today:

- **Is this camera tuned, or is it stock?** Four sliders sitting at 50 look exactly like four sliders sitting at 50 because somebody set them there. Whoever inherits the camera cannot tell.
- **Am I blowing the sky or crushing the shadows?** The commonest install fault, and the one thing a 520×292 preview will not show you. On the lab camera the stock picture is 13% crushed; nothing in the UI says so.

### What I would like to do

Make the picture the hero and let it own the column, sized from the viewport height so the sliders stay on screen while you drag them, with `object-fit: contain` so any sensor shape is letterboxed in a box that never moves.

Move the runtime controls **onto** the picture as lit indicators, and leave configuration in the form below. Where a control lives becomes the answer to "does this take effect now or on Save".

[...]

Issue #222, opened 2026-08-30
Blue border around section titles after opening Camera → Settings


usa-:
Starting with the latest version, immediately after opening **Camera → Settings**, each section has a blue border around its title.

The border appears without any user interaction.

<img width="1340" height="371" alt="Image" src="https://github.com/user-attachments/assets/1c8f3979-9132-4e94-a59b-84a71ac6fefb" />

**SoC**: gk7205v300 (gk7205v200)
**Sensor**: imx335
**Firmware**: 2.6.08.29-lite
**Build**: master+eca39ee, 2026-08-29
**Majestic**: Lite HiSilicon (gk7205v200), master+a63821f, 2026-08-29 19:59

openipc-ai:
@usa- thanks for the report — fixed on master (#224).

What you were seeing was the browser's own focus ring. When you pick a section, the page moves keyboard focus onto the section's title: that's what makes the pick announce itself to a screen reader, since choosing a section doesn't navigate to a new page. The title isn't a control, though, so for a mouse pick the ring was supposed to be invisible — and the page relied on the browser to make that call, via the CSS `:focus-visible` distinction.

It turns out browsers genuinely disagree on that call when a script moves focus. Chrome and Firefox decide "the user was clicking, no ring needed"; your browser decides "a script moved focus, better show where it went" and paints its default ring around the title — every time, for every section you open, which is exactly what your screenshot shows.

The fix stops asking the browser to guess. The page itself knows whether you picked the section with the mouse or the keyboard, and now says so outright: a mouse pick draws no ring in any browser, while a keyboard pick keeps it (there it's earning its keep — it shows where you are).

Nothing else about the page changes — the focus still moves, screen readers still hear the section name. You can pick it up with `updatewebui` on the camera; anything from master today has it.

usa-:
Thanks, the fix works.

usa-:
The blue border still sometimes appears when switching to another browser tab and then returning to the Camera Settings tab:

<img width="1410" height="678" alt="Image" src="https://github.com/user-attachments/assets/0d0e28c9-b395-4b87-b224-09ec34827369" />

widgetii:
Found it, and "sometimes" was the clue that identified it.

The ring is muted two ways: one rule that asks the browser whether it considers a script-set focus "visible", and a `.mj-focus-quiet` tag for Safari, which answers that question differently. The tag was being dropped on **blur**.

[...]

PR #237, opened 2026-08-30
Live adjustments: the picture is the hero, and runtime is not configuration

widgetii:
First half of #235. Stacked on #236 (a CI unblock, unrelated) — review that first or read this commit alone.

At 1440px the form column is 990px and the preview rendered 520×292 inside it, because `.mj-live-video` capped at `max-width: 640px` with a hard-coded 16:9 box. Everything below y≈600 was empty: about **60% of the content area held nothing**, on the one page whose whole job is to let you watch a picture while you drag a slider.

The picture now owns the column, sized from the viewport height so the first rows of the deck stay on screen, and letterboxed by `object-fit: contain` rather than squeezed when a sensor is not 16:9. Under it a deck: tone on the left, orientation on the right.

**Two things this changes beyond the layout.**

**Runtime state moves onto the picture.** Night, IR-cut and the lamp change the camera for every viewer the moment they are pressed, while a slider stages a value that Save writes — and the old panel gave both the same blue outline and the same row, which is to say it did not distinguish them at all. They are now lit indicators on the stage bar. The `lightMonitor` case replaces the trio with one sentence naming what has the wheel, instead of three dead switches beside a small link.

**The knobs become detent sliders.** The fill runs from the schema's own `default` to the current value rather than from the minimum, so a stock camera shows no fill at all and one look down the column answers the question an installer actually has: *has anyone touched this camera, and in which direction*. A tick marks the default, a signed delta says how far, and the readout is an input because typing an exact value is what a number is for.

Snapping to the detent happens **only under a pointer**. Doing it on every input event would trap the arrow keys at the default (49 → 50, 51 → 50) and put both values permanently out of reach.

**The per-row `↺` is now local**: it sets the control to its default and leaves the row dirty, like every other control here. It used to call `/api/v1/reset`, which wrote the camera the moment it was pressed while the sliders beside it did not — on a panel whose claim is that nothing is written until Save, it was the one control that broke the rule. The rest of `mj-settings` keeps the server-side reset, where every row behaves that way.

**Orientation.** `mirror` and `flip` stay two booleans in the config and become four states to a person — Normal / Mirror / Flip / 180° — which is how cameras present them. The pad only replaces the switches when both halves are marked `x-live`; any other `x-live` boolean keeps the switch it has always had, and a knob with no declared `default` gets no tick, no delta and a disabled reset. A note under the pad points at `image.rotate` for the quarter turns: that enum is deliberately `["0","90","270"]` — 180 is absent because mirror+flip already give it, free and at sensor level — and 90/270 resize the stream, so they are not `x-live` and stay on their own leaf.

**Hold-to-compare** posts the schema defaults while held and the live values back on release, through the same endpoint the sliders already drive. It is the one control that can leave the camera somewhere nobody asked for, so the restore is guarded on pointerup, pointercancel, lost capture, blur and visibilitychange; a browser that dies mid-hold leaves stock on the camera and a dirty form that Save puts right.

**Every emoji is gone.** A word in small caps is denser and less ambiguous than a glyph plus the same word, and emoji are not a typeface we control: the eye on the IR filter button had no glyph in the stack the camera ships and rendered as an empty box. What icons remain are inline SVG on a 20px grid.

`preview-hero.js` now publishes `MajesticHero.wireFullscreen`/`wireSnapshot` before it looks for the Live page's own stage, so this page shares them instead of keeping a second copy of the `/image.jpg` naming, the `jpeg.enabled` gate and the delayed `revokeObjectURL`. The bar's auto-hide stays Live-page-only: here the bar carries the runtime indicators, and state you have to wave a mouse at is state nobody reads.

No majestic changes and no new endpoint. `bootstrap.min.css` is untouched — every Bootstrap class dropped here is still used by another page.

### Verified on an hi3516av300 (imx415, 2.6.08.29-lite)

- each of the four sliders pushes `/api/v1/image` with its own key and all six values; Save persists (`yaml-cli -g .image.contrast` → 57) and survives a reload
- **toggling Night changes the picture and leaves the dirty count alone** — the rule this exists to make visible
- the local `↺` issues no `/api/v1/reset`; orientation 180° posts `mirror=1&flip=1` and reports two pending changes
- hold-to-compare reverts and returns, including when the tab is hidden mid-hold
- `nightMode.lightMonitor: true` hides the trio and shows the chip; IR-cut and the lamp correctly disable with no pins configured; `video1.enabled: false` leaves **Sub** disabled rather than merely hidden
- at 390px the deck stacks with no horizontal overflow and every target is ≥44px (the track is 316×44)
- light theme readable; the Live page still plays, still fullscreens and still auto-hides its bar


widgetii:
All four addressed in 5e35e5d. Three were real bugs, one a consistency slip I'd claimed in the commit message and not kept.

**Compare restore could lose (2)** — the sharpest one. `postLive()` fired each write as an independent fetch, so a short hold could let the defaults land after the restore and strand the camera at stock: the exact state the control exists to undo. Serialised through one promise chain, each link swallowing its own rejection so a failed write cannot wedge the queue. Proof off the wire, five press/release pairs back to back: `50 70 50 70 50 70 50 70 50 70`, ending on the live value.

**Fullscreen listeners leak (3)** — real. The Live page wires it once per load, but this stage is rebuilt on every visit to the leaf. `wireFullscreen` returns a disposer now and `renderLive` registers it in `state.liveCleanup`. Counted over three away-and-back cycles: 1 while mounted, 0 while away, 1 on return.

**Live navigation loses focus (4)** — real, and it was a regression against #222 specifically. The old panel had an `<h3>`; I replaced it with a `<span>` for the micro-caps look and took the `revealSection()` focus target with it. It's an `<h3>` carrying the same class, with the heading margin removed.

**Orientation icons on a 26px grid (1)** — fair. Redrawn inside 20×20, with the frame landscape within the box rather than the box itself, and the four transforms reflecting about (10,10) to match. All fifteen inline SVGs on the leaf now report the same viewBox.

Re-verified on the hi3516av300: 180° still posts `mirror=1&flip=1`, the stage still re-mounts, fullscreen and snapshot still wired, camera left at stock.

PR #238, opened 2026-08-30
Live adjustments: a luma histogram and scene starting points

widgetii:
Second half of #235, stacked on #237.

Two things the panel could not answer before.

### The histogram

*"Am I blowing the sky or crushing the shadows"* is the commonest install fault and the one thing a preview at this size will not tell you by eye. The answer is already in the browser — the decoded picture is sitting in the `<video>` element — so a 160×90 canvas readback four times a second computes it for free and **the camera does no work at all**. No endpoint, no majestic change, nothing per-frame on a 32 MB board.

The two clipping fractions are the point; the shape is context. Anything over 1% of pixels in the first or last bucket lights an amber band on that edge and says how much.

The maths lives in `www/a/mj-luma.js` rather than in `mj-settings.js` because **it fails silently**: a histogram computed with the wrong coefficients, the wrong bucket edges or the wrong denominator still draws a plausible shape over a live picture, and nobody reviewing a screenshot can tell that "2.1% blown" is out by a factor of three. `tests/luma.test.js` drives it with frames whose answer is known by construction — flat black, flat white, a quarter/quarter mix, a 256-step ramp — and pins the two off-by-ones that would otherwise ship looking fine:

- 255 landing one past the last bucket, so **a blown-out sky reports no clipping at all**;
- Rec.601 coefficients on 709 content (each primary lands in a different bucket under 601, 709 and a plain average, so the test can tell them apart).

It buckets the **rounded** 8-bit code value. The 709 coefficients do not sum to exactly 1 in binary, so a grey pixel at 128 computes as 127.99999999999999 and truncates one bucket low; on a greyscale ramp that shows up as buckets of 3 and 5 where every one should hold 4. That is the one case where a reader can check the answer by eye, so it is worth getting exactly right.

The vertical scale is logarithmic. Linearly the shadow and highlight tails — the part worth looking at — flatten into an invisible smear along the axis: a bucket at 1% of the peak would draw 0.94% tall. The test pins that it draws as a bar instead.

### Scene presets

Four coarse starting points above the sliders that refine them. Which one is lit is **derived** by comparing the current values against the table, never stored, so there is no fifth piece of state to keep in step with the four that already exist, and a preset the user has since nudged reports itself as Custom without anyone having to remember to clear a flag.

They set tone only — a preset must never quietly change which way up the picture is — and the panel calls them starting points, because a preset presented as an answer is worse than no preset: every install has its own light. **The values are my judgement and want a pass against real scenes on real hardware**; treat them as a first cut. Neutral reads as "Stock, every value at its factory default" rather than as a preset name, which is what somebody inheriting a camera needs to hear.

The sampler is torn down through `stopLivePreview()` with the rest of the leaf, so leaving the section does not leave a timer reading a detached element four navigations later.

One collision worth recording: `.mj-chip` was already taken by the Live page's on-video status badge, which is `position: absolute` — so the preset row laid out as four buttons stacked in the top-right corner of the viewport. Renamed to `.mj-scene-chip`. Neither `node --check` nor the unit tests can see that; only looking at the running page can.

### Verified on an hi3516av300 against a real stream

- at stock the lab scene reads mean 33 with 13.0% crushed and the amber band on the left; lifting brightness to 85 moves the mean to 128 and clears it
- each slider moves the plot, except hue — a chroma rotation leaves luma unchanged by definition, which is the right answer rather than a miss
- Low light sets 60/42/38/50 and names itself; nudging one value drops to "Custom — 1 value differs from stock"
- leaving the section stops the sampler, with exactly one running again on return

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/reset-watch.test.js"
+    "test": "node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/reset-watch.test.js && node tests/luma.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/luma.test.js
+++ b/tests/luma.test.js
@@ -1,0 +1,133 @@
+// The luma histogram behind the Live adjustments panel.
+//
+// This exists because the subject fails silently. A histogram computed with
+// the wrong coefficients, the wrong bucket edges or the wrong denominator
+// still draws a plausible shape over a live picture, and nobody reviewing a
+// screenshot can tell that "2.1% blown" is off by a factor of three. The two
+// clipping fractions are the only numbers on the panel an installer acts on,
+// so they are the ones worth pinning to frames whose answer is known by
+// construction.
+//
+// It also cannot be reproduced on demand in a browser: it needs a camera, a
+// negotiated stream and a scene that clips. Here the frames are made up, which
+// is the point.
+'use strict';
+
+const path = require('path');
+const { check, group, done } = require('./assert');
+
+const luma = require(path.join(__dirname, '..', 'www', 'a', 'mj-luma.js'));
+
+// An RGBA buffer of w*h pixels, each filled by px(i) -> [r, g, b].
+function frame(w, h, px) {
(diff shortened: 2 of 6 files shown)


widgetii:
Real, and fixed in ec09bcc.

The neutral branch called 50/50/50/50 "Stock — every value at its factory default" on the strength of matching the Neutral row of a table that hard-codes 50. On a build whose declared defaults are not 50, matching Neutral and sitting at the factory setting are different facts, and asserting the second when only the first held is the one lie this line must never tell — the whole point of it is to tell somebody inheriting a camera whether it has been touched.

It now reads `off`, the count of fields differing from their own `schema.default`, which the same function already computed for the Custom branch. A build with unusual defaults reports "Neutral — unmodified preset" instead: true, and still useful.

Worth being explicit that the divergent-schema case is **not** something I can reproduce — it needs a majestic-side schema change, and every camera I can reach declares 50. So this rests on the code reading the schema rather than a table, plus a check that nothing regressed on a real one: Neutral still reads Stock, and nudging one value still reads "Custom — 1 value differs from stock".

Issue #240, opened 2026-08-30
WebRTC served-channel mismatch: reflect it in the controls, not only the chip


openipc-ai:
From #184. When WebRTC cannot serve the selected channel (e.g. a main-stream profile no browser accepts), the camera serves the other channel and the status chip now names it — but the stream radios still show the user's selection, which the reporter rightly calls short of ideal:

> if the video corresponding to the selected combination of buttons is unavailable, the better approach would be to leave the controls unchanged and restore their previous state, ideally with a message explaining why the switch could not be performed.

Doing this honestly needs one prerequisite: today the page *infers* the served channel by matching frame dimensions against the configured sizes, which is silent exactly when the two channels share a size or a size is unset. The clean design is for the camera to state the served channel authoritatively in the WebRTC signalling, and for the page to then:

1. move the radio to the channel actually being served (controls tell the truth),
2. show a dismissible message naming why the requested channel could not be served (the camera log already knows the reason — it belongs in the UI too).

The chip disclosure that shipped first remains as the always-visible summary; this issue is about the controls and the explanation.

Proposed by @usa- in https://github.com/OpenIPC/majestic-webui/issues/184#issuecomment-5469920584

openipc-ai:
@usa- — this landed today, in the shape you proposed on #184.

**What was wrong.** Over WebRTC, `?stream=` has always been a preference the camera may override — a main stream on `profile: high` has nothing a browser's WebRTC offer accepts, so the camera serves the substream. The page could only *infer* that from frame sizes, which proves nothing when both channels share a size, and the radios kept claiming the channel that was not playing.

**What changed.** Two halves:

- The camera now states the served channel outright in the WebRTC signalling, right after it answers the offer — which channel it is serving, which one was requested, and why the request could not be met when the two differ.
- The Live page believes it. When your explicit pick cannot be served, **the radio moves to the channel actually playing**, and a dismissible message under the status chip says why — e.g. *"This browser's WebRTC can't decode the Main stream's format — showing Sub stream instead."* The message stays until you dismiss it or change something; it deliberately does not time out, and an audio toggle or a network reconnect won't clear it either.

A few behaviours worth knowing:

- **Your choice is still yours.** The fallback never overwrites your remembered preference — the next page load asks for your channel again, and if it still can't be served you get the honest reflect and the explanation again. Re-clicking your original channel is a real retry, not a dead button.
[...]

Issue #239, opened 2026-08-30
Live adjustments: the preview is too small to judge the picture by


openipc-ai:
From #184: the preview in **Camera → Settings → Live** is capped small (max-width 640px), and that is too little for judgments the panel exists for — the reporter's example is deciding what lighting level a scene needs right now.

Worth solving with the same tools the Live page already uses: size the preview against the viewport rather than a fixed cap, and consider moving the runtime controls onto the video (the Live page's overlay bar pattern) so the picture gets the column's full width without pushing the adjustment knobs below the fold. The knobs-beside-picture layout is this panel's point, so any change has to keep a dragged slider and its effect visible together.

Reported by @usa- in https://github.com/OpenIPC/majestic-webui/issues/184#issuecomment-5469920584

widgetii:
Done in #242 (on top of #237, which replaced the 640px cap).

The preview now sizes against the viewport, and the co-visibility requirement you called out is measured rather than eyeballed — the picture and **every** knob are on screen together at 1280×800 and up. Stage area vs the old fixed 520×292: 1.1× at 1280×800, 1.9× at 1440×900, 3.5× at 1080p and above.

Worth saying that requirement was the part that failed. #237 alone made the picture bigger but pushed Brightness exactly onto the fold on a 1440×900 laptop, which I'd have shipped if I hadn't measured it. Runtime controls (night, IR-cut, lamp) moved onto the video as you suggested — the Live page's overlay pattern — which is what freed the column width.

cc @usa-

usa-:
I have just updated WebUI with `updatewebui` and noticed that the preview image is still too small, although the controls have been updated. The channel switch button is now in the title bar of the “Live adjustments” settings section.

Is this expected?

<img width="1340" height="708" alt="Image" src="https://github.com/user-attachments/assets/5dc171eb-1fd6-46b8-a0d4-8768d0fd973f" />

usa-:
Has anything changed? I see there’s a commit at the end. Can I test it?

usa-:
Is the video supposed to be covered by the text like this in mobile view?

<img width="540" height="928" alt="Image" src="https://github.com/user-attachments/assets/611a518d-df67-43dc-8231-a819d47cab78" />

widgetii:
Not expected, no — and the `1340 × 708` in your screenshot turned out to be the whole story.

[...]

PR #241, opened 2026-08-30
Live View: in Auto, the chip always names the channel it picked

openipc-ai:
From #184's latest round (@usa-): Auto mode had no indicator of the actual current stream — the radios can't say which channel Auto picked, and the chip's served-channel note spoke only on a WebRTC mismatch. With Auto selected the chip now names the picked channel on every frame (both transports); manual picks keep the mismatch-only rule since the radio already names the intent. Covered in the vm suite. Related follow-ups filed separately: #239 (Live-adjustments preview size), #240 (reflecting a served-channel mismatch in the controls, pending authoritative channel metadata from the camera).


openipc-ai:
Both findings were real and are fixed: Auto's label now claims nothing over WebRTC when frame size cannot distinguish the channels (unset or identical sizes — a fallback there is undetectable, and a guess is worse than silence; #240 stays the authoritative fix), while MSE keeps the label since its subscription is exact; and the Auto handler now redraws the chip itself, covering the pick-unchanged path where nothing else would (and MSE never could). Both cases covered in the vm suite; purged CSS regenerated, which was the failing check.

PR #244, opened 2026-08-30
Live adjustments: balance the deck, and let a missing player cost only the preview

widgetii:
Two fixes — one reported, one found while checking it on more hardware.

### 282px of nothing under "Reset all to stock"

#242 moved Scene out of the left column to keep the sliders next to the picture. That fixed the co-visibility problem and left Tone alone in a column sized by the three groups opposite it. I had the numbers when I made that change and shipped it without looking at the result.

Scene goes back to the left column, but **under** the sliders rather than above them: it sets those same four values so it belongs beside them, Tone stays hard against the picture so #239's guarantee still holds, and the columns end up carrying comparable weight.

|                          | before | after |
|--------------------------|--------|-------|
| dead space, left column  | 282px  | 59px  |
| dead space, right column | 23px   | 23px  |
| deck height              | 489px  | 377px |

Phone stacking becomes Tone → Scene → Luma → Orientation, so the presets sit directly under the sliders they drive.

### A missing player script took the whole leaf down

`withVideo` tested only `MajesticVideo`, but `attachLivePreview()` also reaches for `MajesticTransport` and `MajesticSwap`. On a camera whose WebUI predates those files they 404, the browser parses the error page as JS, the globals never appear, and `renderLive` threw partway through — stage on screen, deck never built, nothing said.

It now tests the whole stack, so an incomplete install costs the preview and keeps the controls. Verified by denying the transport at document start and watching the deck build anyway with no exception:

```
{"transport":"undefined","stage":0,"deck":1,
 "knobs":["Brightness","Contrast","Saturation","Hue"],"chips":4,"geo":4,"luma":0}
exceptions: none
```

### Verified across three SoC vendors

| board | vendor | sensor | stream AR | stage behaviour |
|---|---|---|---|---|
| hi3516av300 | HiSilicon | imx415 | 16:9 | fills |
| hi3516ev300 | HiSilicon | imx335 | 1.705 | pillarboxed, 20.8px bands |
| ssc30kq | SigmaStar | imx335 | 1.333 | pillarboxed, 121.4px bands |
| t31l | Ingenic | sc2332 | 16:9 | fills |

That spread is what found the second bug, and it exercises two things HiSilicon alone could not:

- the stage's *"any sensor shape is contained, not squeezed"* claim is now tested against a **4:3** stream, not only against 16:9 where it cannot fail;
- the quarter-turn hint is correctly **absent** on SigmaStar and Ingenic, whose schemas do not declare `image.rotate` at all. That guard was written blind in #237 and this is the first time it has run against a schema without the key.

Those two platforms do report `image.rotate` in `config.json` while omitting it from `config.schema.json`. That is a daemon-side divergence rather than something this panel can act on, and it is filed separately.

### DIFF
--- a/www/a/mj-settings.js
+++ b/www/a/mj-settings.js
@@ -1332,7 +1332,14 @@
 
 	function renderLive(form) {
 		const fields = liveFields();
-		const withVideo = !!window.MajesticVideo;
+		// The whole player stack, not just the decoder. attachLivePreview()
+		// reaches for MajesticTransport and MajesticSwap as well, so a page
+		// served without them — an older install, a half-finished deploy —
+		// threw there and abandoned renderLive with the stage on screen and the
+		// deck never built, silently. Missing scripts should cost the preview,
+		// not the controls.
+		const withVideo = !!(window.MajesticVideo && window.MajesticSwap &&
+			window.MajesticTransport);
 
 		// The only place the leaf names itself — the rail's active item says it
 		// too — with the stream picker on the same line rather than in a card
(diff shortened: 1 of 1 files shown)


widgetii:
Real, and a direct consequence of the degradation path this PR adds — fixed in 51e0ac7.

The right column takes Luma only when the player loaded and Orientation only when **both** halves of mirror/flip are `x-live`. Scene used to fill it; now that Scene sits under the sliders, those two failing together leave it holding nothing, at 23rem with a divider either way.

Which groups exist depends on the schema *and* on which scripts loaded, so rather than enumerate the combinations, any column that came out empty is removed — and the deck with it if both were.

Reproduced end to end rather than reasoned about: transport denied at document start **and** the schema rewritten on the wire to drop `x-live` from `flip`, which is exactly the pair described.

```
{"transport":"undefined","stage":0,"deck":1,"colCount":1,
 "colNames":[["Tone","Scene"]],"emptyCols":0,"deckW":966,"colW":[964],
 "geo":0,"luma":0,"chips":4,
 "rows":["Brightness","Contrast","Saturation","Hue","switch:Mirror"]}
exceptions: none
```

One column at the full deck width, nothing blank, no exception. `switch:Mirror` is the documented fallback: the patched schema leaves mirror unpaired, so the pad is withheld and mirror renders as the plain switch any unclaimed `x-live` boolean gets.

Healthy path re-checked on all three vendors afterwards — HiSilicon, SigmaStar and Ingenic all still report two balanced columns.

PR #245, opened 2026-08-30
Live View: the controls stop shouting and the PTZ pad says what it does

widgetii:
The Live page was seven Bootstrap `btn-outline-light` groups strung across the video in 14px sentence case, plus a PTZ pad built out of Unicode geometry and two emoji diamonds. Next to the Live adjustments deck that landed in #237 / #238 / #242 it read as a different product.

### The bar

It now speaks the same vocabulary as that deck: black glass (`rgba(13,15,20,.72)` over `blur(10px)`), micro-caps labels, a lit indicator wherever there is state, inline SVG on a 20px grid.

The markup keeps **every id and the input-behind-a-label pattern** — that is what `preview-page.js` drives (`.checked` / `.disabled`) and what keeps the groups reachable from a keyboard — so only the presentation moved. Volume rides inside the audio glass, so muting and level read as one control instead of two things that happen to sit next to each other. The speaker's muted/unmuted ending is a CSS rule off the checkbox rather than a JS icon swap, which leaves `preview-page.js` writing only the word.

### The emoji are gone from the player

`U+1F507` speaker, `U+1F3A4` microphone and `U+1F4F7` camera in the bar; `U+25B2` and friends in the pad; `U+1F536` / `U+1F537` on the two focus buttons — a shape that says nothing about focus, in a typeface nobody controls, carried over from the sandbox pad. This is the argument the fullscreen button in this very bar already won, when `U+26F6` came out as a box on real hardware.

Zoom and focus also gained the words they send. A magnifier is read by everybody, so zoom keeps its icon and adds WIDE/TELE; focus has no glyph anyone decodes — photography's flower and mountain mean close-up and landscape *modes* on a stills camera, not "pull focus nearer" on a moving lens — so NEAR and FAR carry it, with the icon holding only the idea of focus: one bracket frame, its subject large for near and small for far. Those four words are literally what `btzoom` is told.

### On a phone the pad comes off the picture

At 390px the stage is 330×186, and steering overlaid costs more than it is worth: a thumb-sized pad is 152 of those 186 pixels, permanently (the pad does not hide with the bar), and on a Pelco camera the zoom/focus cluster stacked above it needs another 130 that do not exist — it rendered over the top edge of the stage. Below the picture both fit at full size, and the picture stays a picture.

The bar stays overlaid at every width, because it lives inside `.mj-stage` and that is what carries it into fullscreen — a bar outside the stage would vanish the moment anyone went fullscreen and take the chip and the pad's anchoring with it. It scrolls sideways rather than wrapping into rows over a 186px picture, with the icon group pinned to the right edge of the scrollport, so fullscreen — the thing that makes a 330px stage usable — is never what falls off.

### What is deliberately untouched

- **`preview-ptz.js`** — it addresses the pads by id and `data-act` / `data-dir`, and every one of those is preserved. No JS change was needed for either backend.
- **`preview-adapt.js`** — the encoder toast keeps its ↗ / ↘ arrows.
- The transport machinery, the fallback chain and the remembered-transport keys.

### Verification

- `npm test`, `tools/lint-templates.sh`, `tools/regen-bootstrap-css.sh` (62579 → 62011 bytes — the Bootstrap button classes left `preview()`).
- A **real Pelco-D camera** (gk7205v300 / imx335, `ptz=true`, stock `btzoom`) at 1440 and 390: nine pad buttons, four function buttons reading Wide/Tele/Near/Far, chip / stats / snapshot / fullscreen all present, zero emoji inside `#mj-player`.
- A stepped **eight-way** backend for the other pad, same widths.
- The two `vm`-based suites gained `mj-mute-t` / `mj-talk-t` in their `IDS` lists, per the rule in CLAUDE.md that any element `preview-page.js` touches is covered there.


widgetii:
Confirmed and fixed in 50c5876 — it reproduces on hardware, not just on paper.

Measured at a 320px viewport: the stage is 264px wide (viewport less ~56 of page and card gutters) and the row wants 284, so both panels hang about 10px over each edge of the picture onto the card's padding. The threshold is around 340px, as the review's arithmetic said.

Below 360px the two panels now stay side by side and give up 4px of grid (2.75rem → 2.5rem tracks, and the function buttons' minimum width 3.875rem → 3.25rem) rather than stacking. Stacking was the other option and it is worse here: it would put ~315px of chrome under a ~150px picture. 40px targets still clear WCAG 2.5.8 comfortably, and these are press-and-hold controls where a mis-hit costs a release and another press.

One detail worth recording: only the *pad* gets shorter — the zoom/focus cluster is unchanged in height — so a Pelco stage has to keep the 11rem bottom margin the wider step reserves, not the 9.5rem the narrow step sets. That falls out of `:has()` specificity rather than source order: `.mj-stage:has(> .mj-ptz .mj-ptz-fn:not([hidden]))` is (0,4,0) against (0,3,0).

Re-verified at 320 on the same Pelco-D camera: the row sits inside the picture's width and the Stream URLs link is clear.

PR #248, opened 2026-08-31
Firmware and extension pages: one vocabulary, mostly from one place

widgetii:
#247 taught the settings page the Live deck's vocabulary, which left the eleven server-rendered pages — network, time, interface, update, settings, and the six extensions — as the only ones still speaking the old one.

### Measured on a lab hi3516ev300 at 1440

| | before | after |
|---|---|---|
| text input, fw-network | **712px** — for a hostname, and for an IP address | 20rem, beside the 16rem select and 10rem number the settings form already used |
| switch row vs text row | 24px vs 64px | same two-line shape as every other row |
| field label | 1rem sentence case — the size of the value below it | micro-caps `11px / 600 / .07em` |
| card background | `#14161c`, the page colour exactly | a real surface |

The card background is the same Bootstrap shadowing of `--bs-card-bg` found in #247; the workaround there was scoped to the settings form. A card that is a surface on one page and an outline on the next is exactly the inconsistency this removes, so it is on `.card` now.

### Most of it is not per-page

The `field_*` helpers render every row on all of these, so labels, control widths, boolean rows and hints move once and land on **all 13 DSL pages** — the eleven plus `fw-editor`, `fpv-wfb` and `info-logs`, which would otherwise have been the only ones left looking old.

Three helpers carry the rest:

- **`card_head "title" "note"`** — the deck's head: micro-caps name, hairline, and an optional note for state the reader would otherwise have to go and find. Telegram says `on`/`off`, Network says `DHCP · eth0`.
- **`group_head "title"`** — replaces the `<div class="text-uppercase x-small text-secondary mt-3 mb-2">` that four pages had each grown their own copy of.
- **`button_submit`** — the card's foot rather than a button loose under the last field, with an optional note for what saving does.

The boolean's lit word is filled by CSS from the checkbox's own state (`:has()`), because these pages are rendered by the server and have no JS to write it — `mj-settings.js` still writes its own.

### Per page

- Cards size to their content instead of `h-100`. That one class is what made Telegram's "Remote send" **1216px of nothing** beside the form, and stretched Network's "Current connection" the same way. Equal height is kept only where two cards of the same kind sit side by side — fw-update's two install actions, which now line up at 231px.
- The four single-form extension pages drop `col-lg-8` → `col-lg-6`; their content caps at 20rem, so the wider column was 700px of empty right-hand side.
- On **fw-update** the switches move into a card of their own instead of floating between an alert and two cards — the one thing you choose before flashing was the only thing on the page not in a card. The "stops video and reboots" warning goes onto the two buttons that start it rather than standing at the top whether or not you are about to press anything.
- `dl.list` — the read-out beside a form — gets the same treatment: term in micro-caps, value in mono and right-aligned, so a column of addresses lines up on its own edge.

### Verification

`npm test`, `tools/lint-templates.sh`, `tools/regen-bootstrap-css.sh` (62011 → 61430 bytes; the regen also picked up `mt-auto`/`d-flex`/`flex-column`, without which fw-update's two card feet did not align — the purged-subset trap CLAUDE.md warns about).

On the lab hi3516ev300, all eleven pages fetched and checked: every one renders with no haserl error, and every row on them carries the same shape. Two rows were hand-written rather than helper-generated — fw-time's timezone pair (datalist + readonly mirror) and fw-interface's disabled username — and were brought into line by hand.


widgetii:
Confirmed and fixed in 2db378d — and it is a safety affordance rather than cosmetics, so thank you for catching it.

#160 gave "Reset config (wipe overlay)" a danger-coloured label precisely so it would not read like "Upgrade kernel" directly above it. That rule selects `.boolean.destructive .form-check-label`, and this branch moved the switch's name to `.form-label`, so the one row on the page that throws away every setting had quietly gone back to looking like its neighbours.

Both class names are covered now, so the rule cannot be unhooked again by the label moving. The lit word beside the switch goes red with it — on that row "On" is the state worth reading twice.

Verified on the lab hi3516ev300 rather than by reading the selector: the row's label computes to `rgb(220, 53, 69)` at weight 500, where a plain switch label is the muted micro-caps `rgba(222, 226, 230, .5)`.

PR #250, opened 2026-08-31
The pages the sweep had not reached yet

widgetii:
#248 covered the eleven pages named in that request. This is the rest of what a person can actually open, so nothing is left speaking the old vocabulary.

### fpv-wfb — the substantial one

It carries **its own field helpers**, because WFB stores booleans as `0`/`1` and its labels want a tooltip. All four now emit what `field_*` emits: `mj-row`, the control wrapper, and the label above a switch with the lit word beside it. Its four hand-written rows — TX power, bandwidth, serial port, OSD FPS — were spelled out to match, and its headings are `card_head`.

The `?` beside a WFB label carried `data-bs-toggle="tooltip"`, which **has done nothing since the Bootstrap JS bundle was dropped** — `main.js`'s shim covers modals, dropdowns and dismiss, not tooltips. What was actually showing the text is the `title` attribute, so that is what it keeps; the filled secondary badge becomes a mark that reads as part of the micro-caps label.

### login.html and setup.html

The two doors into a camera, and a person meets one right after the other — so they are retokened **together**. `#1a1d20` / `#212529` / `#6ea8fe` / `#0d6efd` was a Bootstrap default set matching nothing else on the device.

They stay self-contained by necessity (an unclaimed camera serves no `/a/*`, and the login page cannot pull a stylesheet it needs auth for), so the values are literal — and they are the same literals `bootstrap.override.css` sets on `[data-bs-theme=dark]`. The micro-caps go on `label[for]` only: `.check` wraps its own checkbox and reads as a sentence, not a field name.

### Headings elsewhere

`fw-editor` (the file being edited moves into the head's note instead of trailing the title and wrapping it), `fw-reset`, `fw-restart`, `mj-configuration`, `mj-endpoints`, and `tool-recordings` — where "newest first" is exactly what the head's note is for, so the hand-rolled flex row holding it apart from the title is gone.

### A regression from #248, fixed

Bootstrap paints a range track in `--bs-secondary-bg` — which #248 made **the colour of a card**. On a card the track vanished and a slider became a thumb floating in nothing: fpv-wfb's TX Power, at 1 of 55, looked like a stray dot. The track now takes a translucent body colour that reads against either surface in both themes, and the thumb takes the brand accent for the same reason the switch did in #247 — the compiled subset carries `#0d6efd` literally and never reads `--bs-primary`.

This affected every `field_range` on every card, not just FPV.

### Verification

`npm test`, `tools/lint-templates.sh`, `tools/regen-bootstrap-css.sh`.

On the lab hi3516ev300: every page under `cgi-bin` fetched and checked for a haserl error (20 pages, all clean), both auth pages rendered and looked at, and the FPV slider inspected visually rather than reasoned about — `getComputedStyle` cannot read a range's shadow parts, which is what made the vanished track easy to miss in the first place.


widgetii:
Confirmed and fixed in b82ac0f — and this was the one page in the change nothing had actually rendered.

`fw-restart.cgi` carries no includes on purpose: it runs `reboot` while rendering and has to work while the system is going down, which the file's own comment spells out. A mechanical heading conversion put `card_head` in anyway, where the shell treats it as an unknown command and the heading simply does not appear. It goes back to a plain `<h3>` with a note saying why.

Two things came out of it worth recording. Every other helper call site in `cgi-bin` was checked for its include rather than just this one — they are all fine. And the reason this slipped through my own sweep is that the page reboots the camera on *any* render, HTTP fetch or direct haserl alike, so it was the one page I had excluded after it took the lab unit down mid-check.

PR #256, opened 2026-08-31
Theme: cameras.html joins the palette, and the restart page stops being repainted

widgetii:
Reported as `/cgi-bin/fw-interface.cgi` and `/login.html` looking like different products, with Dark Reader suspected.

> **Scope narrowed after merging master.** This branch was cut from a stale local master and missed ten commits. #250 had already swept `login.html` and `setup.html`, so both are now taken from master wholesale and this branch's versions are dropped. What is left is the two things that sweep did not reach.

## Dark Reader was not the cause on the pages reported

Both already carry `<meta name="darkreader-lock">` — `p/header.cgi` (since #107) and `login.html` — so the extension disables itself on each. What was on screen was the pages' own colours.

## What is actually left

**`cameras.html`** was not in #250's sweep and is still on stock Bootstrap dark (`#1a1d20` page, `#212529` card, `#343a40` border, `#6ea8fe` wordmark). It is retuned to match its two siblings value for value, and in their house style — plain literals rather than a `:root` token block, because one of three sibling pages written differently is its own small inconsistency. Camera tiles sit on `#232734`, one step above the card, and hover to the accent border and 12% tint `.rec-clip.active` uses.

**`fw-restart.cgi`** was the one page Dark Reader was genuinely recolouring: it includes neither `header.cgi` nor `bootstrap.override.css`, so it rendered in stock Bootstrap dark with an extension repainting it. It gets both. Master's new `card_head` comment is kept.

## The card question, and why it is no longer this PR's

The first revision matched cards to what Bootstrap actually *computes*: `.card` redeclares `--bs-card-bg: var(--bs-body-bg)` on the element, which shadows the override's `:root`-level `--bs-card-bg`, so cards rendered as the page colour rather than the `#1c1f27` the theme intended.

**Master has since fixed the cause**, not the symptom — `bootstrap.override.css` now carries `.card { background: var(--bs-secondary-bg) }`. So cards genuinely lift, master's `#1c1f27` sibling pages are right, and matching the rendered bug is no longer correct. That reasoning is dropped along with the files.

Worth keeping in mind generally: any Bootstrap component that self-declares its `--bs-<component>-*` locals cannot be themed from `:root`.

## Verification

Computed style, not eyeballing — each page loaded in headless Chromium against a reference page built on `header.cgi`'s chrome, **after** the merge:

```
cgi-ref   body=rgb(20,22,28)  text=rgb(230,232,238)  card=rgb(28,31,39)  brd=rgb(44,49,61)  r=12px
login     body=rgb(20,22,28)  text=rgb(230,232,238)  card=rgb(28,31,39)  brd=rgb(44,49,61)  r=12px
setup     body=rgb(20,22,28)  text=rgb(230,232,238)  card=rgb(28,31,39)  brd=rgb(44,49,61)  r=12px
cameras   body=rgb(20,22,28)  text=rgb(230,232,238)  card=rgb(28,31,39)  brd=rgb(44,49,61)  r=12px
```

- `tools/lint-templates.sh` — 30 haserl templates, 17 shell scripts, ok
- `npm test` — all passed
- `npm run build:dist` — ok
- `tools/regen-bootstrap-css.sh` — regenerated from the merged markup, byte-identical to master's committed copy

Also audited: every `var(--…)` reference in the sessionless pages resolves to a declaration in the same file, and no declared token is dead.

## Not addressed

The sessionless pages remain dark-only, while the CGI pages honour `webui_theme`. On a light-themed camera they are still a full flip. They cannot read `webui_theme` (no auth, no CGI), so the nearest proxy is `prefers-color-scheme` — a design call worth its own issue rather than a silent change here.


widgetii:
Fixed, and the finding was correct — thank you.

Verified the numbers independently before acting: the camera address measured **3.99:1** on the `#232734` tile and the subtitle/messages **4.18:1** on the card, both under AA's 4.5:1. It was also a genuine regression — the stock palette this replaced measured 6.42:1 and 7.43:1.

Measuring it turned up the same defect in `login.html` and `setup.html`, which I had taken unchanged from master's #250 sweep: `.5` there is 4.18:1, and setup's `.hint` at `.45` is 3.65:1. Fixed all three rather than just `cameras.html`, since fixing one would put the sibling pages back out of step.

The replacement is not a value chosen to clear the bar: `rgba(222,226,230,.75)` is Bootstrap's own `--bs-secondary-color` in dark, which `bootstrap.override.css` does not redefine, so it is already what muted text is on every CGI page — the sessionless pages were the ones that had drifted. It measures **7.67:1** on the card and **7.09:1** on the tile. `setup`'s `.hint` goes to `.7` (6.86:1) to keep a step of hierarchy rather than flattening into the lede. None of master's design decisions — the uppercase `label[for]` micro-labels, the `.5rem` radii, the indigo accents — are touched.

I then audited every remaining text/surface pair on the three pages. All pass, with one deliberate exception worth naming: **`#fff` on the `#5c70e8` primary button is 4.25:1**. That is `--bs-primary`, the theme's brand accent, and the ratio is identical on every button in the WebUI — a product-wide colour decision rather than a drift these pages introduced, so I have left it rather than re-pick the brand accent in a consistency PR. Worth its own issue.

PR #258, opened 2026-08-31
btzoom-xm: compute the focus checksum over the bytes actually sent

openipc-ai:
Found by smoke-testing #255 against real XiongMai hardware — an 85H50AI zoom block freshly converted to OpenIPC in the lab.

The origin `main.c` in OpenIPC/motors (and flyrouter's sandbox script ported 1:1 from it, which `bin/btzoom-xm` adopted faithfully) computes `set_focus`'s checksum from `command1=command2=0` **before** setting the near/far bit — every focus frame goes out with checksum 1. This MCU validates checksums: zoom/stop frames (whose sums are correct) moved the lens, while focus frames in the original encoding did nothing, in either direction. Recomputing the sum over the bytes actually sent (`near` → 2, `far` → 29) made the focus motor drive both ways.

Verified end to end through the shipped path on that camera: `fw_setenv ptz_control pelco-xm; fw_setenv ptz_port /dev/ttyAMA0`, then `j/ptz.cgi?act=tele` ×4 (261 KB frame → 78 KB: heavy zoom-in, focus left behind), `act=wide` back, and `act=near`/`act=far` pulses walking the JPEG size back up to a tack-sharp 212–228 KB. An MCU that skips validation accepts either form, so the corrected frame is strictly better than the wire-faithful one.

Board notes from the same session (not part of this diff): on the 85H50AI the MCU lives on **`/dev/ttyAMA0`** — the console UART — so `ptz_port` must be set to it and getty must be off that port; and the module is zoom+focus only (pan/tilt frames are accepted and ignored). The same checksum bug is worth fixing at the source in OpenIPC/motors `xm-uart/main.c` and in the sandbox script.

Refs #227.

### DIFF
--- a/bin/btzoom-xm
+++ b/bin/btzoom-xm
@@ -110,21 +110,27 @@ xm_send() {
 }
 
 # xm_focus near|far
-# Mirrors set_focus() in main.c - NOTE: main.c computes the checksum from
-# command1=command2=0 *before* setting the near/far bit, so the checksum
-# byte is always 1 for both directions. Reproduced as-is for wire compatibility.
+# Deliberately NOT a 1:1 mirror of set_focus() in main.c: that code computes
+# the checksum from command1=command2=0 *before* setting the near/far bit, so
+# every focus frame carries checksum 1 — and an MCU that validates the sum
+# discards them. On a real 85H50AI zoom block (lab, 2026-08) zoom and stop
+# moved while main.c's focus frames did nothing; the same frames with the
+# checksum computed over the actual bytes (near=2, far=29) drove the focus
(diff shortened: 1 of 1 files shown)


openipc-ai:
Rewrote the branch to contain only the btzoom-xm checksum fix: two commits from a parallel live-stats workstream (`1fa8c49`, `a1c0a52`, touching `preview-stats.js`/`preview-webrtc.js`) had landed on this branch through a shared checkout and were pushed here by mistake. They are preserved locally on a safety branch for their own PR — the two review findings above apply to that code, not to this diff, and are left for that workstream to address.

PR #264, opened 2026-08-31
Live adjustments: discarding a knob puts the camera back too

### LINKED ISSUE #259: Live adjustments: discarding unsaved knobs leaves the camera on them
Reported by @usa- in https://github.com/OpenIPC/majestic-webui/issues/239#issuecomment-5478019561

**Repro**

1. Open **Camera → Settings → Live adjustments**. Brightness, Contrast, Saturation and Hue are at their saved positions; the picture looks normal.
2. Drag them until the picture visibly changes.
3. Navigate to another section and confirm *"You have unsaved changes. Discard and switch sections?"*.
4. Come back to Live adjustments.

The four sliders are back at their saved positions — but the picture still carries the adjustment. The controls and the camera disagree, and the controls are the ones lying.

**Mechanism**

The `x-live` knobs write to the camera as you drag them: `renderField` binds `input`/`change` to `pushLive()`, which debounces 120 ms and `POST`s **every** `x-live` field to `/api/v1/image?…` (`www/a/mj-settings.js`). That is a runtime write — it does not touch `/etc/majestic.yaml`, so `state.config`, which the form re-renders from, never learns it happened.

Discarding then drops the *form*, not the camera. The section switch calls `load()`, which calls `stopLivePreview()` — that tears down the player and the leaf's document-level listeners and nothing else; there is no revert path anywhere in the file. So the sliders come back at the config values while the ISP keeps the dragged ones.

It is not specific to the section switch. A plain reload, or closing the tab, ends the same way: the live write is a preview with no expiry and no undo, and the only thing that ever reconciles the two is pressing Save (which persists the dragged values) or *Reset all to stock* (which pushes defaults).

**Fix shape**

Whatever discards the form has to put the camera back at the same moment. The builder for this already exists — `liveQuery(valueOf)` is parameterised precisely so hold-to-compare can post a different value set through it — so the revert is `postLive(liveQuery(f => …state.initial[f.dot]…))` on the discard path, with booleans mapped the way `liveValue` maps them (`state.initial` holds `'true'`/`'false'` from `getValue()`, while `/api/v1/image` wants `1`/`0`).

Two call sites need it: the "Discard and switch sections?" branch in the nav handler, and any other place that abandons dirty live fields. A reload cannot be covered from the page at all, which is a fair argument for the camera-side preview being scoped to something that ends with the session — but the two in-page cases are the ones reported and the ones worth fixing first.

widgetii:
Closes #259, reported by @usa-.

Drag Brightness until the picture changes, navigate away, confirm *"Discard and switch sections?"*,
come back: the slider reads 50 again and the picture is still dark. The controls and the camera
disagree, and the controls are the ones lying.

### Why

The knobs write to the camera **as they are dragged** — `pushLive` POSTs every `x-live` field to
`/api/v1/image` — and that is a *runtime* write. It never reaches `/etc/majestic.yaml`, so
`state.config`, and the `state.initial` snapshot taken from it, never learn it happened. Discarding
then dropped the **form** and nothing else: `load()` calls `stopLivePreview()`, which disposed the
player and the leaf's document-level listeners, and there was no revert path anywhere in the file.

### Where the fix goes

In that teardown, not at the prompt. The prompt is only one of the ways out — **browser Back reaches
`load()` through `onPopState` and does not even ask**, which was the same desync with no dialog in
front of it. Every path that abandons the leaf already runs through `stopLivePreview()`.

`liveQuery(valueOf)` was already parameterised for exactly this (hold-to-compare posts defaults
through it), so the revert is `postLive(liveQuery(liveSaved))`, with `liveSaved` mapping
`state.initial`'s `'true'`/`'false'` strings to the `1`/`0` the endpoint wants. It writes only when an
`x-live` field actually differs, so navigating away from an untouched panel still posts nothing.

A **reload or a closed tab** abandons the edit the same way, and is the one path a normal request
cannot cover because the fetch dies with the document — `sendBeacon` on `pagehide` is queued by the
browser and outlives it. Deliberately *not* on a bfcache hide: that page comes back with its controls
still dragged, so the camera should be waiting for it; reverting there would re-create the bug facing
the other way.

### Verified by measuring the picture, not the config

The config was never what was wrong, so checking it would have proved nothing. Mean luma of
`/image.jpg` reads **23 / 110 / 211** at `luminance` 0 / 50 / 100 on this hi3516av300, which is enough
to tell a real desync from a no-op.

Before:

```
1. at rest        saved=50 shown=50 picture=110.3
2. dragged to 15  posted luminance=15                picture=43.9
3. discarded      posted (nothing)                   picture=43.8
4. back on Live   saved=50 shown=50                  picture=43.8   <- desync
```

After:

```
1. at rest        saved=50 shown=50 picture=111.3
2. dragged to 15  posted luminance=15                picture=45.0
3. discarded      posted luminance=50                picture=112.1
4. back on Live   saved=50 shown=50                  picture=112.8  <- in sync
```

And the other four paths:

| path | result |
|---|---|
| browser Back (no prompt) | dragged 43.4 → posts the revert → **109** |
| clean switch, nothing dragged | posts **nothing** |
| full reload while dragged | 43.7 → **109.8**, through the beacon |
| hold-to-compare | still pairs defaults then live values |
| Save, then switch | posts **nothing**; the saved value holds at 176.7 |

`npm test`, `tools/lint-templates.sh` and `npm run build:dist` pass; no CSS or markup changed, so
`bootstrap.min.css` is untouched.

No unit test: `revertLive` lives inside mj-settings.js's IIFE with no export seam, and building one
is a larger refactor than this change should carry. The five paths above were driven end to end
against the camera instead.

### DIFF
--- a/www/a/mj-settings.js
+++ b/www/a/mj-settings.js
@@ -616,10 +616,18 @@
 
 	const isLive = (f) => !!(f.schema && f.schema['x-live']);
 
+	// A save that is on the wire owns the live values. It has handed the camera
+	// the dragged ones on purpose, and state.initial does not catch up until the
+	// response comes back — so for that window `liveDrift()` reports a
+	// difference that is about to stop being one, and a revert built from it
+	// would undo the save.
+	let liveSaving = 0;
+
 	// Whether the camera is running something this page has not saved. Dragging
 	// a knob POSTs it to the SDK immediately; that is a runtime write and never
 	// reaches config.json, so nothing else on the page can tell.
 	function liveDrift() {
+		if (liveSaving) return false;
 		return state.fields.some(f => isLive(f) && f.getValue() !== state.initial[f.dot]);
 	}
 
@@ -630,6 +638,11 @@
 	// browser Back, which does not even prompt — so this lives in the teardown
 	// every one of those paths already runs through rather than at any of them.
 	function revertLive() {
+		// The debounced write is the discarded edit, still 120ms from being
+		// sent. postLive serialises, so leaving it queued would put it AFTER the
+		// revert and hand the camera back the very values being thrown away —
+		// a fast navigation right after a drag would have undone the undo.
+		if (liveTimer) { clearTimeout(liveTimer); liveTimer = null; }
 		if (!state.fields.length || !liveDrift()) return;
 		postLive(liveQuery(liveSaved));
 	}
@@ -645,6 +658,14 @@
 	// here would only re-create it facing the other way.
 	function wireUnloadRevert() {
 		const onHide = (ev) => {
+			// Same reason as in revertLive: a queued write would be the discarded
+			// edit arriving after the revert. There is no chaining it here — the
+			// beacon leaves outside liveWrite because nothing can be awaited on
+			// the way out — so cancelling is the whole of the ordering guarantee,
+			// and it covers everything except a write already on the wire when
+			// the tab closed. That one can still be reordered by the server, and
+			// this page has no way to stop it.
+			if (liveTimer) { clearTimeout(liveTimer); liveTimer = null; }
 			if (ev.persisted || !state.fields.length || !liveDrift()) return;
 			const q = liveQuery(liveSaved);
 			if (q && navigator.sendBeacon) navigator.sendBeacon('/api/v1/image?' + q);
@@ -2503,8 +2524,14 @@
 		if (!dirty.length) return;
 
 		const body = {};
+		// What each field was worth when the body was built, not when the
+		// response comes back: the snapshot below has to record what was
+		// actually sent, and a knob can be dragged again while the POST is in
+		// flight.
+		const sent = new Map();
 		for (const f of dirty) {
 			let val = f.getValue();
+			sent.set(f, val);
 			// array-typed schema fields (MultiRect: roi/crop/privacyMasks) post as a
 			// list of strings, not a comma-joined scalar.
 			if (f.schema && f.schema.type === 'array')
@@ -2517,6 +2544,7 @@
 		btn.textContent = 'Saving…';
 		setToolbarMsg('');
 		clearError();
+		liveSaving++;
 		try {
 			const res = await apiFetch('/api/v1/config', {
 				method: 'POST',
@@ -2529,6 +2557,13 @@
 				showError('Save failed (HTTP ' + res.status + '). ' + txt);
 				return;
 			}
+			// The camera now holds these, so the snapshot the revert is built
+			// from has to say so before anything can read it again. refresh()
+			// sets the same values a moment later from the config itself — but
+			// it can throw, and leaving state.initial stale through a failed
+			// refresh would let a later discard "revert" the camera to values
+			// that are no longer what was saved (#259).
+			sent.forEach((v, f) => { state.initial[f.dot] = v; });
 			await refresh();
 			// Image knobs (x-live) apply instantly; everything structural
 			// (resolution, codec, fps, ...) only takes effect after majestic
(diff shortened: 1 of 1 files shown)


widgetii:
All three review findings were real and are fixed in 5bd7d55.

**1. `liveTimer` survives discard** — correct, and worse than "can": `postLive` *serialises*, so the
queued dragged values were guaranteed to land **after** the revert, not merely able to. Teardown now
cancels the timer, which is the right reading anyway — the pending write *is* the edit being
discarded.

**2. Beacon bypasses write ordering** — correct. Cancelling the timer covers the queued case, which
is the one this page can control; the beacon cannot be chained, because nothing can be awaited on the
way out of a document. What remains is a write already on the wire when the tab closed, which the
server may still reorder. I have said that in the comment rather than implying the path is airtight.

**3. Unload reverts a completed save** — correct, and it is this same bug facing the other way. Two
changes: `liveSaving` suppresses drift for the in-flight window, and on a 200 the values actually
sent are written into `state.initial` **immediately** instead of waiting for `refresh()`. That second
half matters beyond the reported window — `refresh()` calls `fetchJson`, which throws on a non-ok
response, so a failed refresh after a successful save would have left the snapshot stale
indefinitely, not just for a moment.

Both new races verified on the hi3516av300 by measuring the picture, since the config was never what
was wrong:

| race | result |
|---|---|
| navigate inside the 120 ms debounce | one post, `luminance=50`, picture **110.9** — the discarded drag never reaches the wire |
| reload during the save | **no** revert beacon; config `78` and picture **171.8** agree, the save survives |

The five paths from the first commit still hold: browser Back reverts, a clean switch writes nothing,
a reload while dragged goes 46.6 → 113, hold-to-compare still pairs defaults with live values, and
Save-then-switch writes nothing and keeps 177.1.

<sub>Unrelated, but worth recording since it happened on the bench: majestic died once during a
pipeline reload it had requested itself after a `POST /api/v1/config` (`Config changed, requesting
pipeline reload` → `Signal HUP received, reloading config` → `onvif discovery sent Bye` → silence, no
process left). Not reachable from this change — it only issues runtime `/api/v1/image` writes — and
the same POST did not reproduce it afterwards. Flagging it rather than filing, since one intermittent
occurrence is not much to go on.</sub>

widgetii:
On the re-review: finding 3 is marked resolved, and 1 and 2 are held open. Recording why I am merging
over them rather than leaving that silent.

**1 — "teardown does not cancel `liveTimer`":** it does, one call deeper. The cancel is the first
statement of `revertLive()` (`www/a/mj-settings.js:645`), which `stopLivePreview()` calls as *its*
first statement; the finding points at the `revertLive();` call site and does not follow into it.
Worth noting the same pattern already exists three lines from where the finding looks — `wireCompare`
cancels the pending push before its own `postLive` (`:1298`), with the comment *"A queued push would
land 120 ms later and undo the comparison."* Same hazard, same fix, and `liveTimer` is the only
deferred live write in the file (`postLive` callers: `:647`, `:712`, `:1302`, `:1308`).

Measured rather than argued: navigating **inside** the 120 ms debounce now produces exactly one
request, `luminance=50`, and a picture of 110.9 — the discarded drag never reaches the wire at all.
Before the cancel that test left the camera at ~40.

**2 — beacon ordering:** the queued half is fixed by the same cancel. The remaining half — a write
already *on the wire* when the tab closed — has no fix available from this page: `sendBeacon` exists
precisely because nothing can be awaited during `pagehide`, so there is no chain to join. The
suggested "last-write-wins mechanism that remains valid after document teardown" would have to be
server-side, which is alternative 2 in the same review's own assessment, correctly rejected there as
"substantially exceeds the scope and risk appropriate for this bug". I have written the residual into
the comment rather than implying the path is airtight, which is the honest version of this trade.

PR #268, opened 2026-09-01
updatewebui: --restore puts your own files back, and stops eating them

openipc-ai:
Follow-up to #202, reported by @usa- [in this comment](https://github.com/OpenIPC/majestic-webui/issues/202#issuecomment-5481835862): local edits are archived correctly now, but `--restore` never hands them back, and the archives pile up in `/etc/webui`.

Reproduced on a lab hi3516av300 before touching anything — a camera holding nothing but a hand-edited `p/motor.cgi`, exactly the reporter's configuration:

```
# updatewebui
Archived 1 locally modified file(s) to /etc/webui/updatewebui-local-20260901-043420.tar.gz
Write 76 file(s)
# updatewebui --restore
Restored the WebUI the firmware ships.
# grep -c "UART ILLUMINATION" /var/www/cgi-bin/p/motor.cgi
0                                  <- gone
# ls /etc/webui/
updatewebui-local-20260901-043420.tar.gz    <- and still sitting there
```

## What changes

`--restore` becomes the install's inverse rather than a wipe.

**It removes only what an install put there.** A path is removed when the overlay copy still holds exactly what the manifest records for it; the same path holding something else is work done on the camera since, and it stays. That closes a second hole the fix would otherwise have opened — an edit made *after* an install used to be deleted unarchived, and would now have been overwritten by an older archived copy of the same file. It also makes `--restore` repeatable: run it twice and it does the same thing twice, rather than eating on the second run what the first one handed back. Files no install ever wrote are reported as left alone rather than silently skipped; that report is the whole reassurance the reporter is owed.

**Then it puts the archives back.** Every `updatewebui-local-*.tar.gz` is unpacked over what was just removed, oldest first so a file edited again between two installs comes back as its latest version, skipping any path the run decided to keep. Extraction goes to a staging tree and each destination is recomposed from `PAYLOAD`, so a member naming its way out cannot land anywhere this script does not own (busybox `tar` strips a leading `/` or `../`; verified).

**And then the archives go.** They exist to survive one round trip; once every restored path reads back from the merged view, the camera holds the work again and the archive is the redundant copy. An archive that could not even be opened keeps the whole set.

**The manifest is emptied, not deleted.** "This script installed nothing" and "no version that kept a record has ever run here" must not read alike — the second is what sends `--restore` down the blanket sweep of `/var/www` that cannot tell your files from ours. That sweep stays as the no-manifest fallback for legacy cameras and is the one path on which an edit can still be lost; one install of v5 closes it. The install side now reads a *non-empty* manifest as its record, so an emptied one still routes it to the overlay-as-record path and local edits stay recognisable.

`--no-backup` opts out of both halves.

## Result

Same camera, same starting state, after `updatewebui` then `updatewebui --restore`:

```
--- verdict
/var/www/cgi-bin/p/motor.cgi: OK
MY EDIT IS BACK
overlay: /overlay/root/var/www/cgi-bin/p/motor.cgi
/etc/webui: updatewebui.manifest
served: 77 files, status.cgi 200
```

Byte-identical to where it started.

## Verified on hardware (hi3516av300)

- reporter's scenario end to end, from a stock camera — edit archived, put back, archive removed
- an edit made *after* the install wins over the older archived copy of the same path, and the user's own never-installed file is left alone
- `--restore` twice, then a third time: second and third are clean no-ops, all three user files intact
- an install straight after a `--restore` still recognises all the local work (the emptied manifest routes it to the overlay-as-record path)
- `--no-backup --restore` keeps the archives and puts nothing back; a plain `--restore` afterwards consumes them
- a corrupt archive is reported and the set is kept, never deleted
- clean camera with no local edits: install → `--restore` leaves 0 overlay files and `diff -r /var/www /rom/var/www` identical, `status.cgi` 200
- a whiteout left by an older version is still cleared with a manifest present — the firmware's `ext-backuper.cgi` shows through again
- legacy camera (overlay full, no manifest at all): blanket sweep still works, 0 overlay files, page 200
- `--restore --dry-run` with nothing installed still lists the empty directories; with an install in place it describes the offer instead
- path-escape archive (`../../x`, `/tmp/x`) lands entirely inside the staging tree and never leaves it

## Review round

Four more, each reproduced on the camera before being fixed:

- **A second `--restore` deleted an edited `/usr/sbin/telegram` that the first had just reported as the user's own work.** `LEGACY_SHARED` is the fallback for a camera whose installer recorded nothing, but it ran on every restore while only the manifest loop feeds `kept` — so once the first restore emptied the manifest, the guess ran unopposed over exactly the file it had promised to leave alone. It now fires only where there is no manifest at all.
- **A truncated archive was reported unreadable and then put back anyway.** `tar` writes the members it has already read before failing on the one it cannot, and extraction went straight into the shared staging tree. Each archive is now unpacked on its own and merged only once it has been read whole.
- **A local edit that is a symlink came back as a regular file** holding its target's content, or — where the target had since gone — failed and aborted the run just after the removals. `cp -df` recreates the link, `chmod` is skipped for it, and `cmp` no longer reads through it (which would have called a link "already there" against a plain file holding the same bytes).
- **An archive holding paths outside the three trees this script installs was deleted with "everything they held is on the camera again"** — about files it had put nowhere. Its in-tree members are still restored; the archive stays, and the message says why.


openipc-ai:
All three findings are real — I reproduced each on a lab hi3516av300 before changing anything, and all three are fixed in fb4f0d8.

**1. Repeat restore deletes edits (`/usr/sbin/telegram`).** Confirmed exactly as described:

```
-- restore #1:
Your own work, left exactly where it is:
  /usr/sbin/telegram
/usr/sbin/telegram: OK
-- restore #2:
Remove 1 installed file(s), revealing the firmware's WebUI:
  /usr/sbin/telegram
/usr/sbin/telegram: FAILED        <- destroyed
```

The sharpest of the three: it ran unopposed over the one file the previous run had *named* as the user's own. `LEGACY_SHARED` is the fallback for a camera whose installer recorded nothing, but it was evaluated on every run while only the manifest loop feeds `kept` — so once the first restore emptied the manifest, the guess had nothing arguing against it. It now fires only where there is no manifest at all, which is the state it exists for. Three consecutive restores now leave the edit intact.

**2. Corrupt archives partially restore.** Confirmed, though it needed the right shape: a truncated *gzip* fails in the decompression step and lands nothing. A valid gzip wrapping a truncated tar is the case that bites, and the output is exactly the contradiction you describe —

```
updatewebui: cannot read …tar.gz, what it holds is not put back
Put 2 locally modified file(s) back:
  /var/www/two.html     <- the truncated member
```

Each archive is now unpacked into a directory of its own and merged into the staging tree only once it has been read whole. Same test: nothing lands, archive kept.

**3. `unpack_archives` dereferences saved symlinks.** Real, and the worse half is the one the description mentions last: `cp` on a symlink whose target has since gone fails, and `die`s the run immediately *after* the removals — the worst moment to stop. `cp -df` now recreates the link, `chmod` is skipped for symlinks (it follows the link and would land on the target), and a symlink carries no md5 expectation since it holds no content of its own. Verified both ways: a symlinked `p/motor.cgi` comes back as a symlink, and still does with `rc=0` when its target has been deleted in between.

The full regression suite was re-run on the camera after the fixes: reporter scenario round trip, `--restore` ×3 as no-ops, `--no-backup` keeping the archive then a plain restore consuming it, clean-camera round trip (`diff -r /var/www /rom/var/www` identical, 0 overlay files, every `/usr` file the install placed removed), legacy no-manifest sweep, whiteout clearing, and all pages 200.

Issue #273, opened 2026-09-01
Day / Night: the section is rebuilt — pin map, filter test, and pin discovery


widgetii:
@usa- — this landed in #270 and #271 and I would value your eye on it, the way you have gone over the other settings sections. It is on `master` now.

**What changed.** The section used to be twelve flat rows, four of which asked for GPIO pin numbers describing your board's wiring. Those four are gone from the form. In their place:

- **A map of the camera's pads.** Click a pad, pick what it is — the filter's closing coil, its opening coil, the infrared lamp, the daylight sensor. Nothing is typed. The grid is built from what the kernel reports, so it is the right size on your SoC rather than a fixed 80.
- **Test the filter** — moves it and compares the picture in both positions, then says "wired correctly", "wired backwards" or "stuck". It is the only way to tell those apart without waiting for nightfall.
- **Find them for me** — drives candidate pairs and watches the picture until the filter moves, for a camera whose pins nobody knows.
- The dashboard raises a banner on its own when the configuration cannot work: no pin set, a light monitor with nothing to watch, thresholds with no hysteresis, day and night disagreeing.

**Worth trying, if you have a camera with an IR-cut:**

1. Open **Settings → Day / Night** and press **Test the filter**. On a correctly wired camera it should come back "wired correctly" and leave the picture as it found it.
2. Swap the two coils on the map, Save, and test again — it should say "wired backwards" and name the fix.
3. Clear both coils, Save, and look at the dashboard: it should tell you the filter cannot be moved.

**Two things to know before you press anything.** Both the test and the scan **read the picture, so they need daylight** — at night nothing looks like it moved and they will say so rather than guess. And the scan drives pads whose job is unknown; that risk is stated on its consent screen and made survivable (the pads are written to flash before they are driven, so a camera that has to be restarted comes back knowing which pair did it), but prefer entering known pins from the [GPIO table](https://github.com/OpenIPC/wiki/blob/master/en/gpio-settings.md) when you have them.

**What I would most like your opinion on:** whether the page now reads as one thing rather than a form with a panel bolted on top, and whether the wording anywhere still assumes you know what a GPIO is. I went through it once removing every config key from anything a person reads — `nightMode.irCutPin1` became "the filter's closing coil" and so on — and that pass is exactly the kind of thing you have caught me being half-hearted about before.

[...]

PR #278, opened 2026-09-01
Day / Night: the scan re-reads the pads, and a verdict does not outlive its wiring

widgetii:
Two things #273 found after the rebuild landed.

## The scan found nothing until the page was reloaded

> *"If I unset the IR-cut pin settings, save the settings and start the search, it finds nothing. If I stop the search, reload the page, and start the search again, it finds the pins immediately."*

`openScan()` was handed the pad list fetched when the **panel mounted**, and `pairs()` skips every pad in its `assigned` set. Clear the pins, save, and that snapshot still named the two pads as spoken for — so the scan went on skipping the very pads it was being asked to find. Reloading the page "fixed" it, which is the tell that the staleness was in the page and not on the camera.

It re-reads the list when the scan is opened, and falls back to the cached one if that fetch fails: a stale list is out of date, not wrong, and scanning with it beats refusing to scan.

Measured on a Hi3516EV300 with all four pins cleared, driving the real page:

| | candidate pairs |
|---|---|
| scan without reloading, **before** | 254 — pads 11 and 10 skipped |
| scan without reloading, **after** | **281** |
| scan after a reload (the workaround) | 281 |

## A verdict outlived the wiring it was measured against

> *"Should messages like these disappear when any settings on the page are changed (or, maybe, only when the IR-cut filter pin settings are changed)?"*

Yes — and the second one. "The filter is wired backwards" names a fix (*swap the two coils*), so it stops being true the moment someone does that, and a stale copy tells them to undo the swap they have just made.

The verdict now remembers the assignment it was taken on and is dropped as soon as that stops matching, which covers a map edit, a save, a per-row reset and the scan's own proposal alike. **Only the pin fields invalidate it** — an unrelated setting on the same page does not, because the verdict is a statement about the wiring and nothing else.

Verified on the camera: the filter test reported "wired correctly", then moving the closing coil to another pad hid it immediately.

## Not in this PR

The third point — whether **"Majestic cannot move the IR-cut filter"** becomes annoying on the dashboard for someone with no IR-cut hardware — is a real design question and I have answered it on the issue rather than guessing at code here.

`npm test` green, `tools/lint-templates.sh` clean.

### DIFF
--- a/www/a/mj-settings.js
+++ b/www/a/mj-settings.js
@@ -1880,6 +1880,12 @@
 		// the day one turns on it and a stale answer does not mis-word the
 		// verdict, it inverts it.
 		const start = ircutSample ? (ircutSample.ircut | 0) : 0;
+		// Snapshot the wiring now, not when the probe returns: it takes seconds
+		// and the map stays live throughout, so reading the fields at the end
+		// would stamp the verdict with an assignment it was never measured
+		// against — and syncVerdict() would then find them matching and keep a
+		// stale verdict on screen.
+		const testedOn = fieldAssign();
 
 		IRCUT.probe({
 			settleMs: 1500,
@@ -1909,7 +1915,10 @@
 				'Live adjustments to move it back. The test itself found: '
 				: '') + '<b>' + esc(v.title) + '</b> ' + esc(v.detail);
 			result.hidden = false;
-			state.ircutTestedOn = fieldAssign();
+			state.ircutTestedOn = testedOn;
+			// Edited while the probe ran: the verdict describes wiring that is
+			// no longer on screen, so it goes straight back out.
+			syncVerdict();
 		}).catch((e) => {
 			result.className = 'alert alert-danger py-2 px-3 mt-2 mb-0 small';
 			// A test that could not finish reports that it could not finish. It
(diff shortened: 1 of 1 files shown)


widgetii:
All three addressed in the second commit — all real, and the first is the one that mattered.

**3 — Verdict hides scan controls (Bug).** Correct, and worse than it reads. `openScan()` builds its panel inside `#mj-ircut-result`, the same element the filter test writes into, so after a test `syncVerdict()` was willing to hide it — and the scan's *own hit* writes the pins it found, which arrives as a changed assignment. So the sequence is reachable without the user touching anything: test, scan, scan finds a pair, panel disappears along with **Stop**, while pads are still being driven. On the one control in this UI that can stop a camera answering.

`syncVerdict()` now refuses to touch the element while the scan owns it, and `openScan()` drops the stamp when it takes the element over, since the verdict it belonged to no longer exists. Verified on the camera: with a verdict showing, starting a scan and changing the wiring mid-sweep leaves the panel and its Stop button in place.

**1 — runIrcutTest snapshots changed wiring.** Correct. The assignment was read when the probe *finished*, and the probe takes seconds with the map live throughout, so an edit made during it became the stamp — the two then matched and the stale verdict stayed up. The wiring is snapshotted before the probe starts and re-checked on completion, so an edit during the probe drops the verdict rather than adopting it.

**2 — syncVerdict misses manual edits.** Correct, and reachable: when the pad list cannot be read the four pins are exposed as plain number fields, and the test can still have produced a verdict. Rather than adding listeners to those four, `syncVerdict()` now hangs off `updateDirty()` — the funnel every edit passes through, whatever moved it — so the map, a save, a per-row reset and a typed number all behave identically. That also removed the two explicit call sites the first commit added.

Re-verified on a Hi3516EV300: the verdict still disappears the moment a coil moves to another pad. `npm test` green, `tools/lint-templates.sh` clean.

PR #284, opened 2026-09-01
Dashboard: "no filter here" can be said once, on the camera

widgetii:
Closes the last open point from #273.

> *"Won't this message on the main page become annoying for users who don't intend to configure an IR-cut filter?"*

It would. The missing-pin banner is correct **and** permanent, which for a camera with no IR-cut filter fitted is a warning that can never be satisfied.

## Why the owner has to be the one who says

Nothing the camera can measure separates *a filter nobody wired* from *a camera that has none* — the config is empty either way. So this is not a judgement the dashboard can make; it is one it has to be told, once.

The claim lives in `/etc/webui/ircut.conf`, **not** `localStorage`, because it is a fact about the hardware rather than about the browser looking at it. Every phone and desktop that opens the page asks the same question about the same camera and should not each have to be told separately. It gets its own file because `fw-interface.cgi` writes the theme into `webui.conf` with `>`, which would take any lodger with it.

## Two things stop it silencing a real fault

- **Only the missing-pin finding offers it.** Every other finding is about a filter that *is* configured — wired backwards, stuck, hunting — and a filter wired backwards is not an absent filter. None of those can be waved away.
- **Configuring a pin drops the claim**, because that contradicts its premise. Someone who says there is no filter, then wires one, then has it fail, still gets told. The promise was to silence *"you have not set this up"*, never *"the one you set up has stopped working"*.

The endpoint reads the flag back out of the file rather than reporting what was asked for, so a flash that refused the write cannot answer "dismissed" and leave the banner reappearing later with no explanation.

## Verified on a Hi3516EV300, through the real dashboard

| step | banner | control | flag on camera |
|---|---|---|---|
| no pin configured | shown | shown | absent |
| press **No filter here** | hidden | hidden | `ircut_no_filter="1"` |
| reload the page | hidden | hidden | persists |
| set a pin | hidden | — | **auto-cleared** |
| remove the pin again | **shown** | shown | absent |

## Also

Joins the "could not be disconnected" list with semicolons. The role names contain commas, so *"IR-cut filter, closing coil, IR-cut filter, opening coil"* read as four things and named none of them — visible in the report on #273.

`npm test` green, `tools/lint-templates.sh` clean (30 templates, 19 shell scripts), and `www/a/bootstrap.min.css` regenerated for the added `btn-link`.


widgetii:
All three addressed in the second commit — all real.

**1 — `wired` ignores opening coil.** Correct. The filter is an H-bridge across two pads and the map assigns them independently, so a camera with only `irCutPin2` set is half-configured, not unfitted — and it is exactly the camera that needs the banner, since half a bridge drives nothing. Either coil now contradicts the claim.

Demonstrated on the camera: with a dismissal on record and no pins the banner stays hidden; assigning **only the opening coil** clears the claim from `/etc/webui/ircut.conf` and raises the banner. Before the fix that camera stayed silent.

**2 — Config race shows false warning.** Correct, and I introduced it. `diagnose()` reads a null config as an empty one and an empty one has no pins, so painting before `mjConfig()` resolves accuses a correctly wired camera of having nothing connected — and on a camera whose heartbeat is down, nothing repaints it afterwards. Every other caller already refused to paint in that state; the refusal is central now so a new caller cannot forget, and the config load repaints when it lands.

One honest note: **I could not reproduce this on hardware.** I removed the guard from the deployed copy and sampled the page every 200 ms through the first ~5 s with the camera fully wired — no false banner, because on this camera the CGI reliably loses the race to majestic's built-in `config.json`. That ordering is an accident of relative latency and nothing guarantees it, which is the point of the guard; but I am not claiming a reproduction I did not get.

**3 — Failed clear is ignored.** Correct. The claim was dropped locally *before* the request and the read-back discarded, so a delete the flash refused left the page believing it was gone while it survived on the camera — ready to suppress the banner once the pin was taken away again. The answer now comes from the file, and the attempt is made once per load rather than on every heartbeat, so a persistently failing delete does not turn into a request every two seconds.

`npm test` green.

PR #289, opened 2026-09-01
Live: the page is the picture, and the picture covers the screen

widgetii:
Redesign of `/cgi-bin/preview.cgi`. The page frame decided how big the video was and the screen got no vote: `.container` capped the stage at 1320 px and the stage reserved the stream's shape inside it. On a 2560 × 1440 monitor an ssc30kq's 4:3 sensor drew at **1264 × 948** — 32 % of the screen, the sensor at 49 %, with 632 px of empty page on each side and 254 px below the card. Same window, same camera, now: **2560 × 1920 at 1:1**.

### The subtraction

`preview.cgi` sets `full_bleed=1`, and `p/header.cgi` / `p/footer.cgi` answer it with no container, no card, no status strip and no footer. `body#page-preview` is a `100dvh` flex column of navbar → banners → stage, so the stage takes what is left and the page never scrolls.

- The **status strip** goes because all four of its readings — the two usage bars, the signature, the clock, the SoC temperature — are on the Dashboard, and every heartbeat writer already guards its `$('#…')`.
- The **Stream URLs** line goes because it is a menu item under Camera.
- Only this page asks for it. The same argument applies to the other twenty, but that is a different change.

### The addition

`www/a/preview-zoom.js` makes the stage a viewport:

| | rule | |
|---|---|---|
| **Fill** (default) | `max(sw/fw, sh/fh)` | covers the window; the long axis pans |
| **Fit** | `min(…)` | the whole frame, letterboxed |
| **1:1** | `1` | real pixels, for judging focus |

Plus drag/wheel panning, pinch and ctrl+wheel free zoom, and double-click for Fit↔Fill. The choice is `mj-view-pick` in `localStorage` — one key, permanent, because nothing here demotes the view behind the viewer's back the way a failed transport does.

**Without the module the page is still right**: the media keep the stylesheet's `inset: 0` and `object-fit: contain`, which is exactly Fit, and the group stays `hidden`. `preview-page.js` knows it through two guarded calls, because that file runs in a bare `vm` in two of the tests.

Sensor shape needs no breakpoint and no special case — it only decides how much there is to drag.

### The chip prints the scale

Fill covers the window by enlarging a stream smaller than the screen: a 1080p main on a 1440p monitor is **133 %**, the 704 × 576 substream **364 %**. A soft picture with no number beside it reads as a soft camera.

### Two zooms, never one control

On a Pelco camera the pad's `Zoom · Wide/Tele` drives a motor — it changes the field of view for every viewer and for the recording, with no undo. So the pad owns the lens and the stage owns the picture: View leads the bar, the pad stays hard against the right edge, and **pinch never reaches the lens on any camera**. Arrows still steer where a pad exists and `shift`+arrows pan there; with no pad the plain arrows pan. `preview-hero.js`'s tap-to-toggle moves to `pointerup` with an 8 px threshold, or every drag would flash the bar at the start of it.

### What the chrome is anchored to

Chrome that *annotates* the picture — chip, stats panel, both toasts — follows it through `--mj-pic-{top,left,right}`, so a letterboxed view does not leave them floating in the black beside it. The bar and the pad do not: they are the player's furniture, and furniture that jumps when you change zoom is worse than furniture on a band (in Fit it also puts the pad in the gutter, covering nothing).

One exception, the chip's alone: a picture too small to carry it without it dominating hands the insets back to the stage — measured against the chip's own `offsetWidth`, not a constant. A 390 × 219 phone in Fit is that case; 1841 × 1381 with 359 px of gutter is not.

The **PTZ pad is back on the picture at every width**. It moved below the stage on a phone because that stage was 330 × 186; it is 390 × 785 now, where the pad covers 6 %, and there is no page below it to move onto. The margin reservations, the `:has()` guards and the corner-radius shuffle they needed go with it.

### Second commit: the toasts

Reported against the deployed page — the chip and the bandwidth notification sit on top of each other. They did, literally: the chip ends at 41 px and `.mj-adapt-toast` was at a hardcoded `top: 2.4rem` = 38.4. `#mj-toasts` is now a flex column hung under the chip at its *measured* height, a hidden toast takes no room, and the chip gained `max-width` — right-anchored in an `overflow: hidden` stage, an uncapped long line ran off the **left** edge and lost its first words.

### Verified

Ten window/sensor/view combinations measured in a browser against the shipped CSS and this module: 2560 × 1381 4:3 gives 2560 × 1920 at 100 % (Fill) and 1841 × 1381 at 72 % with 359 px gutters (Fit); a 1440 × 900 stage with a 16:9 frame letterboxes 45 px and puts `--mj-pic-top` at 45; a 390 × 219 picture forces the insets to 0 and a 700 × 394 one does not. Toast gaps hold at 8 px from 2560 px down to 320, and at 280 the chip wraps and the stack follows.

Rendered on an ssc30kq (2560 × 1920), with the other pages checked for the `header.cgi` / `footer.cgi` change. `npm test` and `tools/lint-templates.sh` pass; `bootstrap.min.css` regenerates with no diff.

### Two calls left open

1. **The portrait phone.** Fill on a 390 × 844 phone showing a 16:9 sensor is a 3.6× crop — 28 % of the frame width. Shipped as Fill-everywhere-remembered: one rule, one tap to change. The alternative is to open portrait phones on Fit, which is a rule with an exception in it.
2. **How far the status strip goes.** Live only, behind `full_bleed`. The eye-travel argument from #239 applies to the other pages too, but that is a different change with a different reviewer.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -278,11 +278,58 @@ The journal is the one write whose failure **stops** the actuation — it is wha
   exception is the PTZ pad — steering, not configuration — kept on the video
   until the multi-user split decides who may steer. Every setting belongs
   under `mj-settings.cgi` only.
-- **The Live page is a hero stage.** `preview()` emits `#mj-stage` — a
-  `position:relative` box that reserves the stream's aspect ratio (16:9 until
-  `onCodec` reports the real one), with everything overlaid on the video so
-  nothing ever displaces the picture: the status chip `#mj-badge` top-right
-  (`CODEC W×H · fps`; fps is live WebRTC stats via a loosened `onStats` gate,
+- **The Live page is the picture, and nothing else.** `preview.cgi` sets
+  `full_bleed=1`, which asks `p/header.cgi` and `p/footer.cgi` for a page with
+  no container, no card, no status strip and no footer: `body#page-preview` is
+  a `100dvh` flex column of navbar → banners → stage, so the stage takes
+  whatever is left and the page never scrolls. It is the only page that asks.
+  The strip went because all four of its readings — the two usage bars, the
+  signature, the clock, the SoC temperature — are on the Dashboard, and every
+  heartbeat writer already guards its `$('#…')`, so an absent strip and an
+  absent footer cost nothing. The `Stream URLs` link went for the same reason:
+  it is a menu item under Camera. Measured on a 2560×1440 monitor, an
+  ssc30kq's 4:3 sensor went from **1264 × 948** (32% of the screen, the sensor
+  drawn at 49%, 632px of empty page each side) to **2560 × 1920 at 1:1**.
+- **The stage is a viewport, and `preview-zoom.js` is the view rule.** Three
+  presets in the bar's `#mj-view-ctl` — `Fill` (`max(sw/fw, sh/fh)`, covers the
+  window, the long axis pans), `Fit` (`min(…)`, the whole frame, letterboxed),
+  `1:1` — plus drag/wheel pan, pinch and ctrl+wheel free zoom, and double-click
+  (pointer only) for Fit↔Fill. The chosen preset is `mj-view-pick` in
+  `localStorage`: one key, permanent, because nothing here demotes the view
+  behind the viewer's back the way a failed transport does. **Without the
+  module the page is still right** — the media keep the stylesheet's `inset: 0`
+  and `object-fit: contain`, which is exactly Fit, and the group stays
+  `hidden`. `preview-page.js` knows it only through two guarded calls
+  (`setFrame`, `scalePct`/`refresh`), because that file runs in a bare `vm` in
+  two of the tests. **The chip prints the scale** (`H264 2560×1920 · 25 fps ·
+  100%`): Fill covers the window by enlarging a stream smaller than the screen
+  — a 1080p main on a 1440p monitor is 133%, the 704×576 substream 364% — and a
+  soft picture with no number beside it reads as a soft camera.
+- **Two zooms, and they never share a control, a label or a gesture.** On a
+  Pelco camera the pad's `Zoom · Wide/Tele` drives a motor: it changes the
+  field of view for every viewer and for the recording, with no undo. The View
+  group scales pixels in one browser. So the pad owns the lens, the stage owns
+  the picture; View leads the bar and the pad sits hard against the right edge;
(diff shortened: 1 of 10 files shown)


widgetii:
All five review findings were real and are fixed in `be74543`.

1. **Scale chip stays stale** — confirmed, and worse over MSE than the description says: the codec is reported once when the connection opens, so the badge kept the opening percentage for the whole session. `preview-zoom.js` now announces a scale change and `preview-page.js` registers `setChip` with it. That is why `preview-zoom.js` moved ahead of `preview-page.js` in the script list — the page has to find the module there to ask. The dependency is one-way and guarded at both ends, so a build missing either file still works.
2. **Frame changes restore stale preset** — confirmed. `setMode()` now keeps `lastPreset` in step, so pick Fit → pinch → change channel lands on Fit.
3. **Fallback uses stale frame geometry** — confirmed. The module re-frames from the image's own `naturalWidth`/`naturalHeight` on `load`, guarded on the dimensions actually differing, because the stream is multipart and fires `load` per frame in some browsers.
4. **Pinch toggles control bar** — confirmed. A second touch cancels the tap candidate outright. Counted inside `preview-hero.js` rather than asked of the zoom module: the two keep their own pointer state, and a shared one would be a third thing to hold in step.
5. **Free zoom exceeds native cap** — confirmed, and the reasoning is right: `max(kFill, 1) * 3` treated the enlargement Fill already needed as the thing to triple. It is `max(kFill, 3)` now — 3× native, except that a preset needing more keeps its place, since a bound below the view in force would drag the picture back the moment you touched it.

**Verified in a browser against the real module**, not just re-read:

| | result |
|---|---|
| ceiling, 3840×2160 frame | **300 %** (3× native) |
| ceiling, 704×576 substream | **364 %** — exactly Fill, no further (was 1092 %) |
| floor | Fit where the frame exceeds the stage, native where it does not |
| preset change | repaints the chip |
| frame change while free-zoomed | lands on the **last** preset |
| 4:3 JPEG over a 16:9 session | re-framed to its own ratio |
| synthetic pinch | leaves the bar alone |
| synthetic single tap | still toggles it |

Re-deployed and re-rendered on the ssc30kq lab camera.

On the summary's alternatives: agreed on keeping the explicit geometry controller, for the reason given — the CSS-only path still has to work, and `object-fit` plus transforms would not have given the overlays a reliable visible-picture rectangle. Portrait-Fit stays an open question in the description rather than a decision made here.

PR #290, opened 2026-09-01
Live: zoom into a rectangle you draw

widgetii:
Follow-up to #289.

The page could already zoom **out** — Fit is a button — and zoom **in** only by pinching, which is a trackpad gesture. A mouse has no equivalent: `ctrl`+wheel works in every browser and is discoverable by nobody. So a viewer on Windows or Linux had a picture that could be enlarged and no way to find out.

### Area

A new toggle in the player bar arms one drag. Draw a rectangle on the picture, release, and that part of it fills the window. It is more precise than either gesture, because the scale is not chosen at all — it falls out of the rectangle — and because you say *which* part of the scene you meant.

- The rectangle is converted to **frame coordinates** before anything moves: stage pixels mean nothing across a scale change, and the scale is about to change.
- Scale is `min(sw/fw, sh/fh)` with 8 % padding, through the same floor and ceiling as every other zoom here — so a tiny selection stops at 3× native rather than magnifying blur.
- The selection's middle goes to the middle of the stage and the ordinary pan clamp then pulls it back. That is what makes a rectangle drawn in a corner land in that corner instead of off the edge.
- A drag under `max(16px, 2% of the stage)` in either dimension is a slip, not a request, and zooms nothing. 40× into a stray 3 px drag is the worst possible reading of one.
- It **disarms after a single drag**. A mode you can forget you are in is the wrong thing to leave over a live picture that also steers a camera.

Borrowed from the QA video comparison in the sibling `rnd-player` (`docs/quality-compare.md`), which draws the same rectangle for the same reason. Its spotlight and persistent highlight are deliberately *not* borrowed: that picture is paused and this one is live, so dimming the rest of the scene past the end of the gesture would hide the thing being watched.

### The wheel now zooms

It used to pan the overflowing axis. That is what a Mac trackpad's two fingers want — but the wheel is the only zoom a mouse has, and one input cannot mean pan on a picture that overflows and zoom on one that does not without changing meaning under the hand mid-gesture. **Panning is the drag**, on every platform and on touch. `ctrl`+wheel stays, since that is what a trackpad pinch sends.

### Esc

Armed, it disarms. Free-zoomed, it returns to the preset in force — what the View group would do, without aiming at a control that auto-hides. On the document rather than the stage, because arming is the nearest thing this page has to a mode and has to be cancellable from wherever the pointer went. In fullscreen the browser takes the key first, which is the right precedence.

### Verified

In a browser against the real module — 800 × 600 frame in a 1000 × 500 stage:

| | result |
|---|---|
| selection of frame rect 200 × 150 | lands **600 × 450**, centred exactly on the stage centre |
| its scale | **300 %** — the ceiling, from a requested 309 % |
| on release | band hidden, button unchecked, disarmed |
| 3 px drag | geometry unchanged, still disarms |
| wheel from Fit (83 %) | 277 %, then **Esc → 83 %** with Fit still lit |

`npm test` and the template lint pass, `bootstrap.min.css` regenerates with no diff, and the page renders on the ssc30kq lab camera.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -305,7 +305,19 @@ The journal is the one write whose failure **stops** the actuation — it is wha
   100%`): Fill covers the window by enlarging a stream smaller than the screen
   — a 1080p main on a 1440p monitor is 133%, the 704×576 substream 364% — and a
   soft picture with no number beside it reads as a soft camera.
-- **Zoom to an area is the discoverable zoom-in.** `#mj-area` arms one drag;
+- **A drag always does something, and which thing is decided by whether
+  anything is hidden.** `setAffordance()` sets `.mj-pannable` and
+  `.mj-drawable` — mutually exclusive by construction, since a picture with
+  nothing off-screen has nothing to pan. The stylesheet reads them for the
+  cursor (`grab` / `crosshair`) and the pointer handler reads them for the
+  gesture. So in **Fit** a drag draws a zoom rectangle with no control to visit
+  first, which is what makes Fit the state you can react from: see something
+  happen, drag a box round it, you are on it. Zooming in makes the picture
+  pannable, so the next drag moves it — the cursor changes at the same moment,
+  which is the disclosure.
+- **Zoom to an area is the discoverable zoom-in.** `#mj-area` arms one drag,
+  and is the way to draw a rectangle while the picture *is* pannable (where a
+  bare drag would move it instead);
   the rubber band is `#mj-marquee` (a single element — its 9999px box-shadow
   spread is what dims everything outside it), and on release `zoomToRect()`
   converts the rectangle to *frame* coordinates before changing the scale,
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -305,6 +305,31 @@ The journal is the one write whose failure **stops** the actuation — it is wha
   100%`): Fill covers the window by enlarging a stream smaller than the screen
   — a 1080p main on a 1440p monitor is 133%, the 704×576 substream 364% — and a
   soft picture with no number beside it reads as a soft camera.
+- **Zoom to an area is the discoverable zoom-in.** `#mj-area` arms one drag;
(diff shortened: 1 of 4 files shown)


widgetii:
All three review findings were real and are fixed in `26d10fa`.

1. **Canceled drags still zoom** — confirmed. `pointerup` and `pointercancel` shared one completion path, so a gesture the browser took away (a system edge swipe, a rotation, another element capturing the pointer) committed its rectangle as though it had been released. `finishDraw()` takes a commit flag now; a cancel cleans up and zooms nothing.
2. **Secondary pointers hijack selection** — confirmed, and the touch case is the bad one: `drawing` held only coordinates, so any further `pointerdown` replaced the origin, either pointer's move redrew the band, and either one's release committed it. The rectangle now belongs to the pointer that began it, and a second finger mid-drag is ignored outright — it cannot start a pan either.
3. **Letterbox space becomes selectable** — confirmed. The band was clamped to the stage rather than to the picture, so a rectangle on the black converted to negative frame coordinates and the pan clamp then landed the view on some edge as though that had been asked for. Both ends are held inside the picture now: the band shows exactly what will be selected, a drag that never leaves the gutter collapses to nothing and is thrown away by the minimum-size test, and one that straddles the edge uses only the on-picture part. Where the picture overflows, this is the whole stage and clamping does nothing.

**Verified in a browser** against the real module — 800 × 600 frame in a 1000 × 500 stage, so Fit draws it 667 × 500 at 83 % with the picture spanning x 167…833:

| | result |
|---|---|
| drag, then `pointercancel` | **unchanged** (83 %), band hidden |
| same drag, then `pointerup` | 138 % |
| interloping second pointer | neither draws nor commits; owner's band stays its own 300 px and its release commits |
| drag x 10 → 150 (all gutter) | **unchanged** |
| drag x 60 → 400 (straddling) | start clamps to 167, zooms to 129 % |

Both lab cameras are updated to this head.

PR #303, opened 2026-09-03
Live: the caption stops deciding where the chip goes

### LINKED ISSUE #302: Live: `Stats` and mode chip are positioned against the screen edges after pressing `Auto`
On the Live page, after pressing `Auto`, the `Stats` block and the current mode chip, which should be positioned over the video, get pushed to the left and right edges of the screen respectively.

The video does not actually occupy the full width of the screen, so the elements are shifted relative to the video boundaries, as if the video width were equal to the screen width.

**SOC:** gk7205v300 (gk7205v200)
**SENSOR:** imx335
**FIRMWARE:** 2.6.09.02-lite
**BUILD:** master+dc2b3fe, 2026-09-02
**MAJESTIC:** Lite HiSilicon (gk7205v200), master+872443d, 2026-09-02 17:30

[majestic.yaml](https://github.com/user-attachments/files/31768570/majestic.yaml)

Image

openipc-ai:
On a 4:3 sensor in **Fit** the picture is pillarboxed, and the chip, the stats panel and the toasts are anchored to the *picture* rather than to the stage so they do not float in the black beside it. Pressing **Auto** threw all of them out to the screen's edges — with the picture not moving a pixel.

## What was wrong

Two defects, in the same four lines.

**1. The gate measured the chip, so the chip decided where the chip went.**

`annotate()` publishes `--mj-pic-{top,left,right}` and has an escape hatch: a picture too small to carry the chrome hands the insets back. It was gated on

```js
const cw = (badge && badge.offsetWidth) || 210;
const roomy = visW >= cw * 3 && visH >= ch * 5;
```

`cw` is the chip's *rendered* width — i.e. its text. In Auto the chip always names the served channel (#184), so `setChip()` appends `" · Sub stream"` and then calls `MajesticZoom.refresh()`. Measured on the reporter's viewport, 1411×707 in Fit, picture 793×649 with 309px pillars:

| | caption | chip | gate `793 ≥ cw×3` | `--mj-pic-left` | chip lands at |
|---|---|---|---|---|---|
| manual | `H264 704×576 · 30 fps · 113%` | 224px | 793 ≥ 672 ✔ | 308.9px | x 870–1094, on the picture |
| **Auto** | `… · 113% · Sub stream` | 318px | 793 < 954 ✘ | **0px** | x 1085–1403, on the black band |

The `<video>` box is byte-identical in both (309..1102). Only a word changed. The same flip is available from a served-channel mismatch note appearing, or from the scale reading going 99% → 100%.

**2. The hatch fired on the wrong axis.**

It was calibrated on a phone in Fit — 390 × 219 of picture with 283px of band **above** it — where surrendering `--mj-pic-top` lifts the chip into empty space directly above the picture. One boolean drove all three insets, so a *width* test surrendered the horizontal pair as well and threw the chrome sideways into the pillars: precisely what `bootstrap.override.css`'s own comment forbids — *"a chip at the stage's corner floats in black, describing nothing."*

## The fix

- **Nominals, not measurement.** `CHIP_W = 210, CHIP_H = 30` — about thirty monospace characters at 0.75rem with padding and border. A threshold that decides where a widget is drawn must not be measured against what that widget happens to say. `--mj-chip-h` stays measured, because it *describes* the chip on screen rather than placing it, and the toast stack needs the real height; the two questions are now visibly separate in the code.
- **`--mj-pic-top` alone is surrendered**, and only when the band above is at least two chip-heights deep. Where it is not, the "escape" moves the chip a few pixels and leaves it on the picture anyway, having cut it loose from the corner it describes for nothing.
- **The horizontal pair is never surrendered.** Sideways there is no band deep enough to clear a chip: in every case cramped enough to ask, the chip is *wider* than the pillar it would move into, so it covers the same picture as before and hangs off into black as well. And on the portrait phone this hatch was written for, the horizontal insets are zero anyway — so surrendering them was never what helped.

## Verification

A harness against a real camera that changes **only the chip's caption** — `MJPEG` → `H264 2560×1920 · 30 fps · 34%` → `… · Main stream`, then `MajesticZoom.refresh()`, stream untouched — and asserts (a) the insets do not move and (b) the horizontal pair equals the picture's real offsets:

| layout | master | this branch |
|---|---|---|
| reporter 1411×707, Fit | defect 1 + 2 | clean |
| 2560×1440, Fit | clean¹ | clean |
| phone 390×785, Fit | defect 1 | clean, top still surrendered |
| landscape phone 785×390, Fit | defect 1 + 2 | clean |
| 1411×707, Fill | clean | clean |

¹ that picture is 1689px wide, so master's gate happens not to trip there; at 1920×707 the picture is 865px and it does.

Master fails five assertions, this passes all of them, and the phone still hands its chip up into the band above the picture. `npm test` passes.

Fixes #302

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -368,10 +368,21 @@ The journal is the one write whose failure **stops** the actuation — it is wha
   deliberately do **not** read them: they are the player's furniture, and
   furniture that jumps when you change zoom is worse than furniture on a band
   (in Fit it also puts the pad in the gutter, covering nothing). One exception,
-  the chip's alone: a picture too small to carry the chrome without it
-  dominating — measured against the chip's own `offsetWidth`, not a constant —
-  hands the insets back to the stage. A 390 × 219 phone in Fit is that case;
-  1841 × 1381 with 359px of gutter is not, so it keeps its chip on the picture.
+  and it is **`--mj-pic-top` alone**: a picture too small to carry the chrome
+  without it dominating, *and* with a band above deep enough to actually clear
+  the chip, hands that one inset back to the stage. A 390 × 219 phone in Fit is
+  that case; 1841 × 1381 with 359px of gutter is not, so it keeps its chip on
+  the picture. The horizontal pair is **never** surrendered — sideways there is
+  no band deep enough to clear a chip, so the move covers exactly as much
+  picture as before and cuts the chip loose from the corner it describes. Both
+  halves of that rule are #302: the gate used to be measured against the chip's
+  own `offsetWidth`, so pressing **Auto** — which always appends the served
+  channel's name — grew the chip by 42%, tripped the gate, and threw every
+  annotation out to the screen's edges with the picture not moving a pixel. A
+  threshold that decides where a widget is drawn must never be measured against
+  what that widget happens to say, so it takes nominals (`CHIP_W`/`CHIP_H`);
+  `--mj-chip-h`, which describes the chip on screen rather than placing it,
+  stays measured.
 - **The two toasts are a stack, not two fixed offsets.** `#mj-toasts` is a flex
   column hung under the chip at `--mj-chip-h`, the chip's *measured* height,
   because that height is not a constant — "MJPEG" against "H265 3840×2160 · 25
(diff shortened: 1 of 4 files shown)


openipc-ai:
Addressed in c81af48 — the finding was right, and it was the dangerous kind.

Both `MajesticZoom.refresh()`'s own comment and `setChip()`'s at the call site still stated *the chip's width decides whether the chrome fits on the picture* as the contract. That is precisely the feedback loop this branch removes, and in this repo a comment explaining a decision **is** the decision to the next reader — leaving them there would have handed someone a licence to put #302 back.

They now say what `refresh()` is still for, which is `--mj-chip-h`: the toast stack hangs off the chip's *measured* height, and a caption that wraps to a second line is taller than the one it replaced. Comments only, no behaviour change; `npm test` still passes.

PR #306, opened 2026-09-03
Settings: a field the camera is quietly overriding says so

openipc-ai:
`outgoing.substream` and `records.substream` only take effect when `video1` is
enabled. With the second stream off there is nothing to publish or record, so
the camera substitutes the main stream and carries on — the right behaviour,
and until now an invisible one.

## What that cost

[OpenIPC/majestic#311](https://github.com/OpenIPC/majestic/issues/311). The
reporter's camera published 1080p H.265 at roughly four times the bitrate his
configuration described, over a link he was already troubleshooting. Asked to
test the setting he toggled it, saw no change — with `video1` disabled both
positions mean the same thing — and reported that it made no difference. That
reads as the setting being irrelevant rather than inert, and **this form is
where he was looking**. A log line would not have reached him.

## The change

majestic now carries the condition in the schema it serves, beside the
`x-live` this file already reads:

```json
"substream": {
  "type": "boolean", "title": "RTMP substream",
  "x-requires": {
    "field": "video1.enabled", "equals": "true",
    "message": "The second stream (video1) is disabled, so the main stream is published instead."
  }
}
```

Drawn as a warning under the control. Not by hiding or disabling it, as
`visibleWhen` would: the setting is a legitimate thing to want, it is
remembered, and it starts working the moment the second stream is switched on.
Hiding it would only move the surprise.

`field` is an absolute dotted path rather than `visibleWhen`'s sibling name —
what decides a setting's fate is rarely its neighbour, and this one lives two
tabs away. So the controlling field is usually not mounted, and the value
comes from the saved config, with a mounted control still winning where the
two share a page, exactly as `fieldVisible` resolves one.

## Why there is a new module

The evaluator is the part of this whose failure is invisible: decide wrongly
that a requirement is met and no warning draws, and a form with no warning on
it looks exactly like a camera doing what it was told — the state #311 was
filed from. A bug there does not degrade this change, it silently reverts it.

Inside `mj-settings.js` nothing could reach it, so it is
`www/a/mj-requires.js`, and the form keeps only the three lookups that are
genuinely page state. It also takes over the operator `visibleWhen` was using,
so there is one implementation of "does this condition hold" rather than the
two this PR was otherwise about to leave in one file — the hazard
`tests/transport.test.js` already has a paragraph about.

## Tests

Each of these was checked by mutating the module rather than assumed:

| what it pins | mutation | checks failed |
|---|---|---:|
| type coercion | remove the `String()` | 7 |
| resolution order | `resolve()` stops on the first source | 1 |
| fail-open | unknown operator fails closed | 2 |

The coercion is the one most likely to bite: the schema carries
`"equals": "true"` as a JSON **string** while the config carries `true` and a
schema default carries `false`, both JSON **booleans**. Compared untouched
every boolean condition is false, so the warning would be permanently on — or,
through `notEquals`, permanently off.

Also pinned: `undefined` from a lookup means "this source cannot answer" and
falls through, while `false` is an answer and stops the search —
`video1.enabled === false` being exactly the case being detected. A truthiness
test there would look right and break the feature.

29 checks, `npm test` green (16/16 files).

## Not verified

`tools/lint-templates.sh` needs `haserl`, which was not available where this
was built, so the one added `<script>` tag in `mj-settings.cgi` has not had
that linter's opinion. `tools/build-dist.sh` globs `www/**/*.js`, so the new
file needs no registration.

The camera-side half is in majestic: the daemon also warns once per load and
reload, for whoever arrives with a `logread` instead of a browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
+    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/requires.test.js
+++ b/tests/requires.test.js
@@ -19,16 +19,24 @@ const { check, group, done } = require('./assert');
 
 const req = require(path.join(__dirname, '..', 'www', 'a', 'mj-requires.js'));
 
-// The condition majestic emits for outgoing.substream, verbatim: `equals` is
-// the STRING "true", because the camera's schema annotation carries the value
-// as a string.
+// The condition majestic emits for outgoing.substream, verbatim. `when` and
+// `equals` are both the STRING "true", because the camera's schema annotation
+// carries values as strings; the config they are compared against carries
+// booleans.
 const SUBSTREAM = {
+	when: 'true',
 	field: 'video1.enabled',
 	equals: 'true',
-	message: 'The second stream (video1) is disabled, so the main stream is published instead.',
+	message: 'The sub stream is disabled, so the main stream is published instead.',
 };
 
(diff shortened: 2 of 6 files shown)


openipc-ai:
All three addressed in 78d2efd.

**1. `undefined` triggered a false warning.** Correct, and it contradicted the
module's own stated rule — an unknown *operator* held, but an unknown *value*
did not. With no mounted control, no saved value and no schema default,
`String(undefined)` compared unequal and the requirement came out unmet: a
definite claim made from no data, with nothing on the page able to clear it.
`met()` now returns true on an unresolved controlling field. Two checks cover
it, and removing the guard fails both.

**2. The warning exposed `video1`.** Fixed camera-side rather than here, since
the message is authored in the schema: it now reads "The sub stream is
disabled, so the main stream is published instead." The config key stays in the
camera's log, whose audience is holding a config file. The camera's test
asserts the key's *absence* from the message rather than trusting the wording
to stay put.

**3. The warning went stale.** Correct — `paint` was in `state.reqUpdaters` but
only `runVisibility()` ran them, and ordinary edits go through
`updateDirty()`. The control now repaints its own warning on `input` and
`change`, and `applyVisibility()` wires a prerequisite that shares the page, at
the same place and on the same events that move a `visibleWhen` row.
`reqUpdaters` holds `{dot, req, paint}` so the pairing can be found.

That third one became load-bearing rather than cosmetic, because of a fourth
problem the review found on the camera side: the condition tested only the
prerequisite, so with `substream` at its default of false the form announced a
substitution on every camera that had never asked for one. The schema now
carries a `when` clause naming the value the annotated field must itself hold,
which makes the field's own value half the condition — so its own edits *must*
repaint it.

`no when means always` pins that the clause is optional, so the annotation's
original meaning survives for a field with no sensible "off".

PR #311, opened 2026-09-03
Recordings: the player goes on fetching the clip past the first twelve seconds

openipc-ai:
Opening any recording longer than about a quarter of a minute on
`tool-recordings.cgi` played a few seconds and then stopped for good on the
loading spinner. Pressing play did nothing — nothing was paused; the element
had run out of media and nobody was fetching any more.

## What happens

`mkPlayer`'s fill pump drives itself through `updateend`, and only for as long
as it is appending. It stops appending the moment the buffer is `AHEAD` (12 s)
in front of the playhead, and no append means no `updateend` to call it back —
so from there the pump is finished for the life of the clip. `evict()`, the one
other thing that could raise an `updateend`, is called from inside that same
dead handler. Playback then ate the twelve seconds it had left and starved.

It does not show up on a short clip: there the pump reaches the end of the file,
calls `endOfStream()` and plays through. It needs a clip long enough that 12 s
of buffer is reached mid-file — which is every real recording, and the page
opens the newest one sixty seconds before its end, so it is the first thing
anybody meets on the page.

Measured on an hi3516av300 over CDP, H.264 1080p with 1-second fragments, on a
120 s clip:

| wall | |
|---|---|
| 11955 ms | last byte of the clip ever requested |
| 12837 ms | last `appendBuffer` → `buffered [32, 74]`, `currentTime 61.85` — 12.15 s ahead, a twentieth of a second past the gate |
| 24934 ms | `video:waiting` at 73.94 s, `readyState 2` |
| → 110 s | not one further request; 46 s of the clip still on the card |

`video.duration` also stayed `Infinity` throughout, because `endOfStream()` is
only reached by a pump that gets to the end of the file — so the native controls
showed no length either.

## The fix

Let the element drive the pump too. The buffer only shrinks because the playhead
moved, which is exactly what `timeupdate` reports; `waiting` is the backstop for
a buffer that ran dry before `timeupdate` got there, since once it has,
`currentTime` stops advancing and `timeupdate` stops with it. It also picks the
pump back up after a range read that failed, which used to end it just as
permanently.

Both listeners come off in `destroy()` — the element outlives every player
attached to it, so a pair added by `attach()` and not removed would accumulate
one per clip opened.

`evict()` deliberately stays on the append path. The buffer *behind* the
playhead only grows when something is appended, and driving it from here as well
would issue a `remove()` for every quarter-second the playhead advanced.

## Verified on hardware

Same camera, same clip, with this applied: played from 60 s through to 120.04 s
and ended, buffer tracking ~13 s ahead the whole way and evicting behind it, and
`endOfStream` at 52 s of wall time so `duration` reads 120.04 instead of
`Infinity`.

## Test

`tests/recordings-pump.test.js` runs the real page against a synthetic
recording and a stub element, so it needs neither a camera nor a browser. It
fails without the fix in the shape the camera showed — buffered frozen at 73 s
of a 120 s clip, no further appends — and it also covers the two hazards the fix
introduces: that `waiting` alone can restart a starved pump, and that the
listeners are removed with the player that added them.

`npm test` is green (814 checks) and `npm run build:dist` passes.

## Not in this PR

Two separate defects turned up while reproducing this, both in majestic rather
than here, and neither is touched by it:

- the MP4 recorder silently drops every frame past 4 MiB in a GOP, so a
  recording made with a long `gopSize` at a real bitrate has multi-second holes
  in it (43 % of one 120 s H.264 clip on this camera);
- `records.substream: true` produced clips whose `moov` carries the main
  stream's SPS over substream samples, which Chrome refuses to decode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
+    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/recordings-pump.test.js
+++ b/tests/recordings-pump.test.js
@@ -1,0 +1,414 @@
+// The recordings player's fill pump: does it keep asking for the clip?
+//
+// This is here for the reason every other file in tests/ is. The failure is
+// silent and it cannot be reproduced on demand: MSE appends nothing, raises no
+// error and reports no state anyone can see, so a dead pump looks exactly like
+// a slow camera. Seeing the real thing needs a camera, a recording longer than
+// the buffer the player runs ahead by, and the patience to watch it for a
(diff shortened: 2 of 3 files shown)


openipc-ai:
Addressed the review finding in e26b3d2 — it was a real one.

Every settling delay in `tests/recordings-pump.test.js` is now a condition rather than a fixed sleep, with a ten-second budget that only decides how long a genuine failure takes to report. The one state that cannot be waited *for* is "the pump has gone quiet", so `quiesce()` waits it *out*: it lets a slow runner take as long as it needs to stop appending, then requires it to stay stopped.

Writing it turned up a fixed-delay bug of exactly the kind being removed. `playTo()` had ended with `sleep(80)` to let the last appends land; dropping that left the end-of-clip assertion reading a buffer two fragments short, and it failed twice in five runs. The end state worth waiting on is `endOfStream` — `fill()` raises it from the top of a call the previous append's `updateend` made, so by the time it lands every fragment is already in the buffer.

Verified over 0.25 of a CPU and eight concurrent runs at half a CPU each. It still fails without the fix in the same shape (buffered frozen at 73 s of a 120 s clip), and a passing run takes 3 s.

PR #315, opened 2026-09-04
Settings: the substitution warning reads the field's value before there is one

openipc-ai:
Reported against OpenIPC/majestic#311: on last night's nightly the **Outgoing**
tab lost its Save button, and the console carried
`Uncaught (in promise) ReferenceError: Cannot access 'S' before initialization`.

The tab renders **Enable**, **Address**, **RTP packet size** and then stops — no
*RTMP substream*, no *Audio*, no *Audio codec*, and no save bar, so nothing on it
can be changed. **Record** is three fields short in the same way.

## What it is

`'S'` is `renderField`'s own `getValue`. The two value accessors were declared at
the bottom of the function, beside the `return` that hands them out. The
`x-requires` block sits well above them and paints its warning **on mount**,
which reads `getValue` to ask what the field is currently set to.

A `const` reached before its declaration is a `ReferenceError`, not an
`undefined`. So the paint threw, the throw left `renderField`, and it took every
field after the annotated one with it — and the toolbar, which is built after the
fields are. Minification is what turned `getValue` into `S`, which is why the
console names no field.

## Why nothing caught it

The block only runs on a field whose schema carries `x-requires`, and until
majestic 6fc331c9 no camera emitted one. The consumer here (5264b40) landed the
same afternoon, against a schema that did not yet have the annotation to trigger
it, and shipped looking exercised. Every check ran the untaken branch.

## The change

- The accessors move up, to the first line after the widget dispatch that assigns
  `control._get`/`_set` — as early as they can go, since they read those once at
  declaration. That is what makes the rule statable: **anything built below may
  read the field's value, nothing above may.**
- `tests/requires.test.js` asserts that ordering from the source. The real
  failure needs a camera to reproduce and no DOM-free test can reach the mount
  path, so the invariant is what gets checked. It fails on `master` and passes
  here.
- `check.yml` gains `npm test`. The suite ran in **no** workflow at all —
  `check.yml` gated the minification, `dist.yml` the release, and `npm test` was
  left to whoever remembered it locally. It is dependency-free node, so it is one
  step, and the checks it holds are the silent-failure ones a reviewer will not
  catch either.

## Verification

Against the lab camera's own live schema and config, with majestic 0575f652
installed on it so the annotation is really present:

| build | Outgoing | console | save bar |
|---|---|---|---|
| `master` | 3 of 7 rows | `ReferenceError: Cannot access 'getValue'…` | absent |
| this branch, source | 7 of 7 | clean | present |
| this branch, **minified dist** | 7 of 7 | clean | present |

Record behaves identically (5 of 8 rows on `master`, 8 of 8 here). The warning
itself shows with `video1.enabled` false and hides when true, so the feature this
was blocking now works as designed.


openipc-ai:
Review addressed in 54e88ab. What changed and why:

**The fix itself was protecting an invariant rather than removing the need for
one.** Moving the accessors above the `x-requires` block still left a line in
`renderField` above which a field's value may not be read — in a function
CLAUDE.md explicitly sends contributors into ("add a branch in `renderField`
keyed off `dot`"). The reason they could not simply be declared first was that
they *captured* the widget's optional hatches, so they could not run before the
branch that assigns `_get`. They now read `control._get`/`._set` on every call,
which lets the declaration sit above the dispatch and retires the ordering rule
altogether. Nothing assigns those hatches outside the dispatch today, so it is
the same function called the same way — it just cannot be broken by where the
next block is written.

The test follows: it asserts the call-through form rather than an ordering, and
fails loudly instead of scanning an empty string when it cannot find what it
describes. One line in CLAUDE.md beside the other `renderField` entries.

**Two more in the same block, both latent:**

- The repaint listened on `control`, so a field whose value does not come from
  that one element goes stale under its own edit — the resolution picker keeps
  its custom text box as a *sibling* of the select (which is why dirty tracking
  binds it separately), and an array field's value lives in rows added under it.
  Both events bubble, so the row hears all of them.
- `reqNotice`'s `fallback` asked `sectionFields` for the controlling field's
  default without `withLive`, and that skips `x-live` keys — so an `x-requires`
  pointing at a live-classified field on a camera that never wrote the section
  resolves to nothing, `met()` fails open, and the warning silently never draws.

**CI:** `npm test` is its own job now rather than a step on `build`. It needs no
`npm ci` and does not use the minified output, so as a step it waited out both
and reported nothing when the build failed. It also goes into `dist.yml`, before
the build — `check.yml` runs on `pull_request` only, and the commit that lands is
not the commit that was checked. That workflow is what uploads the tarball every
nightly firmware fetches.

**Re-verified** against the lab camera's own schema and config, source and
minified: `master` renders 3 of 7 Outgoing rows with the `ReferenceError` and no
save bar, this renders 7 with the bar; Record 6 → 9. The resolution picker's
custom hatch and dirty tracking behave identically to `master`, and the warning
now appears the moment the substream switch goes on against a disabled `video1`
and clears when it goes off.

**Not taken — `try/catch` around the per-field render in `renderProps`.** It
would contain any future throw to one row, and the point stands that this PR
removes a trigger rather than the mechanism. But a swallowed exception makes a
missing setting look like a setting this camera does not have, which is the
silent mode this codebase legislates against; containing it honestly means
drawing a visible error row in the field's place, and that is its own change
with its own wording to get right. Worth doing separately.

**Also not taken — driving `renderField` in a `vm`.** The existing harnesses
(`staging`, `auto-source`, `preview-socket`) work because `preview-page.js` runs
against a handful of stubbed elements; `mj-settings.js` is one IIFE with no
export whose mount path needs `innerHTML`, `classList`, `querySelector` and
event dispatch to be real. That is a small DOM implementation, not a stub, and
it is a bigger commitment than this fix should make on its way through.

PR #321, opened 2026-09-04
Setup: on a phone the agreement is read through a keyhole in a page that does not scroll

widgetii:
Reported from a phone: the setup page is "almost impossible to read". Two separate faults sit on top of each other, and only the second one is ours.

### The screenshot's shrink is not the page

That browser laid the page out at a **980 CSS px** viewport and then scaled the whole 640px card to 65% to fit 1080 device px, so 15px text landed at about 9.8px. Reproduced exactly: 640 in 980 is 65.3%, and the card in the screenshot measures 65.2% of the screen. The page's own `width=device-width` is present and is on the camera, so something overrode it — on Chrome for Android that is desktop-site mode, which is remembered **per site**, so one tap on a camera's address makes every page on it render this way. Worth confirming with the reporter; nothing here can fix it.

### Underneath it, the page had no `@media` rule at all

Measured on the served page at 390×844 with the viewport meta honoured — the good case:

| | before | after |
|---|---|---|
| text column | 258px | 358px |
| characters per line | 33 | 45 |
| agreement visible at once | 6.5% (384px of 5879) | 100% |
| page itself scrolls | no | yes |
| Continue | 39.6px | 48.6px |
| accept row | 19px | 50.2px |
| language select | 31.6px | 47px |

33 characters is about half a readable measure, and all three tap targets missed the 44px floor. Worse, the *page* did not scroll while a 384px box inside it did.

Below `40rem` the card stops being a card and the pane stops being a scroller: the document becomes the page, the browser's own scroll reads all of it, and the accept control sits after the text where reaching it *is* reaching the end. The document also comes out **25% shorter** (5879 → 4433px) because the lines stop double-wrapping. A second query catches landscape phones, which are short rather than narrow and miss the first.

### The CSS alone was not shippable

`atEnd()` answers "already read" when the pane does not overflow — a deliberate branch, since the three shipped texts differ by a factor of two in length and a short one in a tall window has no end to scroll to. Release the pane and it **never** overflows, so that branch fires on every phone, at load. Measured with the media query in and `atEnd()` untouched:

```
390x844, at load:  eulaDisabled: false   continueDisabled: false   mustreadShown: false
```

The checkbox came up enabled and the must-read line hidden before a word had moved — a phone would have been the one place the agreement could be accepted unread, with the stylesheet quietly repealing the gate. The overflow test now *selects* the scroller instead of answering the question, the genuinely-short-document answer is kept for the case it was written for, and `window` gets the scroll listener beside `docEl`.

Verified: shut at load on phone and desktop, opening when the page reaches the end, and a short document still counts as read on both. Two new tests cover the released-pane trap and the page-scroll listener; both fail against `master`.

### No regression above the breakpoint

Checked rather than asserted. Card, pane, column, characters per line, gate state at load and all three tap targets come back **identical** before and after at 1920×1080, 1600×900, 1440×900, 1366×768, 1280×800, 1152×720, 1024×768, 900×600, 800×600, 700×900 and 641×900 — and screenshots are pixel-identical at five of those. The first width that changes is 640, which is the breakpoint itself.

A short wide window (1280×500) is also unchanged: the landscape rule is bounded at `60rem` so it cannot reach a laptop. At 900×500 it does fire, and there it only releases the pane — card still 640 wide, still 69 characters, tap targets untouched — replacing a scroller nested inside a scrolling page with one scroller.

`npm test` passes; `tools/lint-templates.sh` passes.

### DIFF
--- a/tests/setup-page.test.js
+++ b/tests/setup-page.test.js
@@ -90,8 +90,25 @@ group('the reading gate holds both controls, not just the box');
 	check('Continue also refuses when the text is unread',
 		/if\s*\(!gateOff\s*&&\s*!readToEnd\)\s*return;/.test(page));
 	check('a document too short to scroll counts as read',
-		/scrollHeight\s*-\s*\w+\.clientHeight\s*<=\s*\w+\)?\s*\)?\s*return true/.test(page),
+		/\w+\.height\s*<=\s*\w+\s*-\s*\w+\)\s*return true/.test(page),
 		'without this a short agreement makes the form permanently unsubmittable');
+
+	// Below 40rem the pane is released and the document becomes the page, so
+	// the pane NEVER overflows. "Does not overflow" therefore stops meaning
+	// "shorter than its box" and starts being true on every phone, at load —
+	// and the answer above is "already read". Measured with the media query in
+	// and this untouched: the checkbox came up enabled and the must-read line
+	// hidden before a word had moved, i.e. the stylesheet quietly repealing the
+	// gate. It is the exact shape of failure this file exists for: the page
+	// looks right while it happens, and seeing it needs an unclaimed camera,
+	// a phone, and an agreement nobody read.
+	check('a released pane is not mistaken for a document that fits',
+		/scrollHeight\s*-\s*\w+\.clientHeight\s*>\s*\w+\)\s*\{/.test(page)
+		&& /getBoundingClientRect\(\)/.test(page),
+		'the overflow test must SELECT the scroller, not answer the question');
+	check('the page scroll drives the gate too, since on a phone it is the scroller',
+		/window\.addEventListener\('scroll',\s*syncAccept/.test(page),
+		'without it the gate never re-asks while the document is being read');
 	check('a language switch clears it — the new text is unread',
 		/readToEnd\s*=\s*false;/.test(page));
 	check('and the gate is dropped where there is no pane to scroll',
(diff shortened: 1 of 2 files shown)


widgetii:
Confirmed and fixed in bc98b94 — good catch, and it is the other half of the mistake this PR set out to fix. `atEnd()` learned to ask which box is scrolling; `renderDoc()` did not.

Reproduced on a phone (390×844): read the English agreement to its end, then switch to Chinese.

```
before:  scrollY 3957 -> 2748   eulaDisabled false   mustreadShown false
after:   scrollY 3957 ->  145   eulaDisabled true    mustreadShown true
```

The shorter document landed under a viewport already at its bottom, so the gate re-opened on a text nobody had seen. The page`s own comment already forbade exactly this ("carrying the last one`s scroll over would enable acceptance of a text still at its first line") — only the mechanism enforcing it had stopped reaching the thing that scrolls.

The question is now asked in one place. `paneScrolls()` is declared beside `docEl` rather than with the gate, since `renderDoc()` sits above the gate and needs it too; `atEnd()` consults it to know what to measure and `rewindDoc()` to know what to rewind.

Desktop is unchanged step for step at 1440×900 — pane scrolled to end opens the gate, a language switch shuts it and rewinds the pane to 0, the new text read to its end re-opens it — identical to master on all four. The regression matrix is unchanged, and a document too short to scroll still counts as read on both layouts.

Two more tests, both failing against master: that the rewind moves whichever box is scrolling, and that the gate and the rewind consult one predicate rather than two that can drift.

PR #323, opened 2026-09-04
Day / Night: the light monitor shows its reasoning, and knows when it has none

openipc-ai:
Companion to the camera side of automatic day/night (the out-of-the-box replacement for the community backlight-control scripts). Current firmware can drive day/night from the sensor's own exposure with nothing wired and nothing calibrated, and names who decides in a new `night_mode_source` gauge. The Day/Night section now meets it:

- **A Light monitor block** under the pin map: one live sentence about what the monitor is doing ("Bright enough for day — switching in 40 s if it stays.") and a chart of the value it watches — sensor gain in multiples of 1× with the day band and any explicit night threshold shaded, or, in legacy threshold mode, raw `isp_again` between the min/max bands. What to chart is a pure, tested `monitorView()`; the chart reuses the dashboard's primitives (camera.cgi now loads them).
- **Findings understand the three-way chain**: an automatic monitor is an observation, not a fault; a monitor the SoC could not feed says so and points at the sensor-pin alternative; only a daemon that reports no source keeps the old "nothing to watch" warning, reworded to say a firmware update fills this gap. The hunting finding stops prescribing thresholds to a camera that has none.
- **The dashboard's day/night line carries the decision source** ("🌙 Night · IR-cut on · lamp on · auto").
- A real dump from a camera running the automatic monitor joins the parser fixtures; the three existing fixtures stay untouched — they pin how older daemons read.

Full test suite passes; `node --check` clean on the touched files.

### DIFF
--- a/tests/fixtures/metrics-sstar-auto.txt
+++ b/tests/fixtures/metrics-sstar-auto.txt
@@ -1,0 +1,353 @@
+# HELP node_uname_info Labeled system information as provided by the uname system call.
+# TYPE node_uname_info gauge
+node_uname_info{domainname="(null)",machine="armv7l",nodename="camera",release="4.9.84",sysname="Linux",version="#2 SMP PREEMPT Sun Aug 30 17:37:29 UTC 2026"} 1
+# HELP node_boot_time_seconds Node boot time, in unixtime.
+# TYPE node_boot_time_seconds gauge
+node_boot_time_seconds 1788306128
+# HELP node_context_switches_total Total number of context switches.
+# TYPE node_context_switches_total counter
+node_context_switches_total 251768399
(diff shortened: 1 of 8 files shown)


openipc-ai:
All four findings addressed in 995f855: (1) the countdown speaks only when both timing gauges exist — no more 'switching in 0 s' from coerced zeros; (2) an absent `night_enabled` no longer reads as 'Day', the sentence skips the word; (3) ADC (source 3) charts nothing, as monitorView's own contract said — `isp_again` was a different quantity wearing the monitor's clothes; (4) charts.js grows `dropChart()` and the panel unregisters the superseded instance before remaking, so remounts stop accumulating detached charts. New tests cover all four; full suite passes.

PR #327, opened 2026-09-04
Recordings: seek where a decoder can start, and stop calling a card healthy when it is not

openipc-ai:
The last of the SD-recording work, on this side of the wire. Two things the page was getting wrong, both silent.

## Seeks and exports could land somewhere unplayable

A fragment is self-contained only if its first sample is a keyframe. Until majestic bounded a fragment by time and bytes that was true of every one — the writer hardcoded *sync sample* into `trun`'s `first_sample_flags` because it could afford to. It is not true any more.

`mp4index.js` skipped that field, so `fragmentAt()` could hand a `SourceBuffer` a fragment opening mid-GOP, whose references were never appended. What comes out is a green or smeared picture rather than an error — a seek that landed there looked like a decode bug. `exportRanges()` had the same hole, and there it means a **downloaded file whose opening seconds are unwatchable in any player**.

Both now snap back to the nearest fragment that opens on a keyframe, and the index records which those are. An export pays at most one GOP of extra footage at the front.

A recording carrying no `first_sample_flags` at all reads as all-sync. That is not a guess: a clip written before majestic emitted the field was written when every fragment began at an IDR, and reading it as "unknown" would make every clip on an older card unseekable except at its very start.

## The page painted itself green over a card recording nothing

`cardWritable()` asked the SD-card endpoint, which asks the filesystem — and the filesystem cannot see the failure that matters most: **a card mounted read-write, with room on it, that majestic is writing nothing to.** Every one of those states reads back as `health: "ok"`.

The page now asks majestic as well, through `/metrics/records`, and will not call a card writable unless both agree. It names what it finds — the recorder has given up on the card, writes are failing, writes are failing intermittently, or the card cannot keep up and footage is being dropped — because "Cannot record" over an empty archive is where somebody starts looking, and the filesystem's verdict would send them to the wrong place.

A camera running an older majestic serves no such endpoint. That is not evidence of anything and is not rendered as any: absent metrics leave the page saying exactly what it said before.

## A gap means different things now

Recording continuously, a gap between clips is footage that should be there and is not. Recording **on motion** — `records.mode`, new in majestic — it is the camera doing exactly what was asked, and labelling it `not recording` sends somebody to look for a fault they do not have. It reads `no motion` in that mode.

The reserved motion lane stays empty, for a narrower reason than the one written on it. In motion mode the clip band above *is* the event record, and drawing it twice would say nothing new; what no camera writes is where the movement was **within** a clip, which is the only thing that lane could add. The caption now says which of those two situations you are in.

## One comment corrected

The note explaining why `locate()` can only approximate said majestic's `mfhd` sequence number "cannot be trusted as an exact fragment ordinal". Current majestic restates it per file, so on a clip it wrote, it is exactly that. Nothing here changes — the cards this page browses are full of clips written by versions that did not, and nothing in the value says which is which — but the comment no longer claims a limitation that is half untrue.

## Tests

Ten new checks over the keyframe snapping, including a clip with no keyframe anywhere (falls back to its start) and one carrying no flags at all (reads as all-sync, still seekable).

A new `recordings-health.test.js` drives the real page against a stubbed camera: every recorder state, a card that cannot keep up, both ways of having no `/metrics/records` at all, and the two gap labels. Every mutation of the new rules fails its test — dropping either snap-back, reading absent flags as non-sync, ignoring the recorder verdict, or treating a missing endpoint as a failure.

`npm test` green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/charts.test.js && node tests/mem-slices.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
+    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/recordings-health.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/charts.test.js && node tests/mem-slices.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/mp4index.test.js
+++ b/tests/mp4index.test.js
@@ -287,7 +287,7 @@ group('buildIndex — the exact walk');
 		check('and does not claim to be complete', idx.complete === false);
 
 		return runLocate();
-	}).then(runExport).then(runSync).then(runCheap).then(() => done());
+	}).then(runExport).then(runSync).then(runSyncCheap).then(runCheap).then(() => done());
 }
 
 // ---- the approximate seek ----------------------------------------------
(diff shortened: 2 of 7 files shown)


openipc-ai:
All six addressed in 27dad2c. The first two were the important ones: **the snapping was inert.**

## 3 and 4 — seek snapping unused, exports remain mid-GOP

Correct, and the more useful half of this review. `buildIndex()` is not a path production takes — a full walk of a 20-minute clip is about eleven hundred range requests — so `fragmentAt()` and `exportRanges()` having learnt about keyframes changed nothing at all. `positionAt()` calls `locate()` and hands `seekTo()` the offset it gets back; `saveSelection()` calls `locate()` and then `spanFrom()` from that same offset. Neither had learnt anything.

`locate()` has now, and it fixes both at once because the export starts where `locate()` left it. Its `refine()` walk already parses every fragment in the window it reads, so remembering the last one that opens on a keyframe is free. `spanFrom()` carries the flag too, or `exportRanges()` has nothing to act on.

**It cannot always find one, and says so rather than pretending.** The refine window is 768 KB — a second or so of video — and a camera with `gopSize: 30` puts a keyframe every thirty of them. `locate()` reports `atSync: false` there. The answer is not a 16 MB read on every scrub; it is a shorter GOP, and for exports it is the forward snap below.

`exportRanges()` now snaps **forward** when there is nothing behind — which is exactly the case `spanFrom()` produces when `locate()` could not snap. Losing up to a GOP off the front of a selection is worse than lead-in and much better than a saved file that opens black. Unlike playback, an export cannot recover a moment later.

## 1 — unknown recorder appears healthy

Right, and the rule is the one this page is built on. Absent and unknown are now different things:

- `404` — the camera **answered** and has no such endpoint. As known as it can be, and it keeps the green badge it had before these metrics existed. Making every older camera grey would be its own false claim.
- request failed, or an answer this release cannot read — **unknown**, and green is a claim.

## 5 — fragment count labeled seconds

Right, and it was counting the wrong metric. The seconds are in `records_dropped_ticks_total`, in the media timescale; `records_fragments_dropped_total` beside it is a count of fragments, which equals seconds only at the default `records.fragmentMs`.

## 6 — historical drops stay critical

Right. Measured from where the counter stood when the page opened, because *"footage is being lost"* is a claim about now — a card that dropped a second last Tuesday should not hold the banner red for ever.

## 2 — dropped footage update ignored

Right. The poll now compares the **rendered verdict** rather than `records_state`, which cannot drift from what the banner shows because it is what the banner shows. The test drives the poll directly rather than waiting thirty seconds for it, so the early-return is exercised at all — it was not before.

## Tests

Eight new checks, nineteen in `recordings-health.test.js`. Every one of the six fixes fails its test when mutated out: dropping either snap, dropping the flag from `spanFrom`, making an unknown recorder green again, measuring drops from boot, or comparing the state instead of the verdict.

Rebased onto master (`package.json`'s test list gained `charts` and `mem-slices` from #324 while this branch added `recordings-health` to the same line). `npm test` green.

PR #329, opened 2026-09-04
Live: the control bar wraps on a phone, and the pad steps out of its way

### LINKED ISSUE #318: Live: control buttons are inconvenient to use on mobile
On mobile devices, the row of buttons on the Live page does not fit on the screen, so you have to scroll it horizontally to access the controls.

It would be much more convenient to display the buttons in multiple rows on smaller screens, so that all controls are visible at once without horizontal scrolling.

https://github.com/user-attachments/assets/1c5717fd-6523-413c-b8c2-0cf74318a3cf

openipc-ai:
Closes #318.

On a phone the Live bar was one row scrolled sideways. Measured on an hi3516ev300 at 390 CSS px, its six groups want **879px in a 390px scrollport — 489px of it, more than half, off screen**, behind a sideways gesture on a picture where every other drag pans or draws a zoom box. At 320px, 559px are hidden. The reporter's video shows the sticky icon group sitting on the transport picker while they scroll to reach it.

**The rule was stale rather than wrong.** The `flex-wrap: nowrap` block's own comment said "the stage is 330 wide … only 186px tall" — true while Live was a card. #289 made the page full-bleed, and the same phone's stage measures **390×785**; that commit rewrote the PTZ pad's comment for the new geometry and left the bar's rationale describing a page that no longer existed. So most of this is a deletion: the nowrap, the `overflow-x`, the scrollbar styling and the sticky `.mj-ico-wrap` go, and the bar wraps below `md` as it already does at every other width. Three centred rows at 390px are 200px, a quarter of the stage, and only while a tap has the bar up.

**Wrapping alone would have hidden two controls.** The pad is `bottom: 3.25rem; z-index: 7` — it draws *over* the bar, and 3.25rem is one row's worth; wrapped, it covered Area and the transport picker. That overlap was already there unfixed on any window under about 900px (at 768×1024 the pad covers the transport picker on master today), and the stats panel carried the same constant (`max-height: calc(100% - 5rem)`). So `preview-hero.js` publishes `--mj-bar-h`, the measured height of the bar's rows — the way `--mj-chip-h` already carries the chip's for the toast stack — and both old constants stay as the var's fallback, so a page whose chrome script did not load lands exactly where it used to.

**Two arrangements, chosen by which dimension has room.** Where the stage is too short to stack them *and* wider than tall, `.mj-ptz-beside` returns the pad to its corner and reserves its column in the bar instead. A phone in landscape is that case and cannot be the other one — 301px of stage against a 211px zoom/focus cluster and a two-row bar — but it is 740 wide, so the same groups take the same two rows in the lane left over. Wider-than-tall has to be asked: reserving 150px of a 320px stage leaves a lane narrower than one group, and the bar answers with a row per group (8 rows, 500px over a 509px stage — worse than the overlap being avoided).

Measured after the change, on a camera with a Pelco zoom/focus cluster:

| viewport | rows | bar | off-screen | pad over a control |
|---|---|---|---|---|
| 390×844 | 3 | 200px (25% of stage) | 0 | none |
| 360×740 | 4 | 260px | 0 | none |
| 430×932 | 3 | 200px | 0 | none |
| 740×360 landscape | 2 | 140px, beside | 0 | none |
| 844×390 landscape | 2 | 129px | 0 | none |
| 768×1024 | 3 | 129px | 0 | none (master: transport picker) |
| 1440×900 | 2 | 82px | 0 | none |
| 320×568 + audio + talkback | 5 | 320px | 0 | clips one control |

The last row is the single geometry where neither arrangement fits — smallest phone, a camera with audio *and* talkback *and* a Pelco cluster. A CSS clamp keeps the pad inside the stage there: off the top it would be unreachable, since a full-bleed page has nothing to scroll, so it stops at the top and clips the corner of Area.

Full suite passes. No markup changed, so `bootstrap.min.css` is untouched.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/recordings-health.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/charts.test.js && node tests/mem-slices.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js"
+    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/recordings-health.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/charts.test.js && node tests/mem-slices.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js && node tests/preview-hero.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/preview-hero.test.js
+++ b/tests/preview-hero.test.js
@@ -1,0 +1,200 @@
+// preview-hero.js — how the control bar and the PTZ pad share the bottom of
+// the Live stage.
+//
+// What must hold, in one sentence each: the bar's measured strip is published
+// for the two pieces of chrome that have to keep off it; the pad rides above
+// that strip where the stage is tall enough to stack them; where it is not and
+// the stage is wider than tall the bar gives way sideways instead; and the
+// answer is a function of the stage's size alone, never of the order it was
+// resized in — which is the one that is easy to get wrong, because reserving
+// the pad's column makes the bar taller and a naive second pass reads that
+// back as even less room.
+//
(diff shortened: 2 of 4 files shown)


openipc-ai:
Addressed all three, thanks — two were real:

**2. Beside layout becomes sticky** — correct, and worse than the class sticking: `layoutChrome()` asked whether the pad and the bar could stack using the bar's *current* height, which `.mj-ptz-beside` may already have inflated by reserving the pad's column, and it published that inflated height as `--mj-bar-h`. Resizing 740×301 (beside) → 390×785 published **360px** of strip where a fresh load of the same stage publishes 180 — the pad would have sat 360px up the picture. The class now comes off before the measurement and goes back on only if the *natural* height still cannot stack, so the arrangement is a function of the stage's size and of nothing else.

**3. Stats overlap letterboxed bar** — correct, and it predates this PR: the panel's `top` is `--mj-pic-top + 0.5rem` while its `max-height` subtracted only the bar, so a full panel reached into the strip by exactly the depth of the band. `--mj-pic-top` is in the subtraction now; measured on the camera the panel's bottom lands 8px above the bar's top row.

**1. `layoutChrome` lacks regression tests** — taken, and it fits what this repo tests for: the fault is silent (the pad simply draws over two rows, nothing reports it, every covered control is still in the DOM) and reaching it needs a phone, a PTZ camera and a particular orientation. `tests/preview-hero.test.js` drives the module in a `vm` against a stub whose bar height is *derived from the lane it is left*, so reserving the column really does cost rows there. It covers the published strip, a camera with no pad, stacking in portrait, giving way sideways in landscape, the portrait stage too small for either arrangement, and that a resize lands where a fresh load of the same size does — that last one fails against the pre-fix code, which is what makes it worth having.

Full suite passes and the fourteen-viewport sweep on the hi3516ev300 is unchanged.

PR #331, opened 2026-09-04
The reset transcript stops printing tags, the agreement fits the window it is read in, and the static pages ship without their comments (#154)

openipc-ai:
Two follow-ups from usa- on #154, both reported after the reboot watch was confirmed working.

## The transcript printed the tags round its own first line

```text
<b>root@gk7205v300-imx335:/tmp# /usr/sbin/sysupgrade -n --web</b>
```

`j/run.cgi` announced `text/html` and wrapped its prompt lines in `<b></b>`, which asked every consumer of a stream of raw, unescaped command output to treat it as markup. Exactly one did — the console page, which un-HTMLs it back into a `<strong>`. The factory reset page renders the transcript as text, so it printed the tags.

The prompt is now marked with **ANSI bold** and the stream says `text/plain`. SGR is what a terminal already speaks:

* `main.js`'s terminal writer already strips every escape it does not act on, so `factory-reset.js` needed no change at all and prints a clean prompt line;
* the console page's local runner reads the same two escapes and keeps its `<strong>`, still building text nodes rather than touching `innerHTML`;
* a `curl` into a real terminal renders the line instead of showing it tags.

Verified on a hi3516ev300: the camera's own `run.cgi` bytes fed through the real `termWriter` come out with no tag and no escape left in them, and the console page still renders both prompt lines bold with the output between them as text.

## The bottom of the agreement card was below the fold

Reported on a desktop Safari with the favourites bar and a row of tabs open. Measured at 1440×660: the card is 735px, the window 660, and the accept checkbox and Continue sit below the fold of a page that *also* has a scrolling pane in the middle of it — two scrollers, and the one you needed was the one no fade or edge pointed at.

The reporter suggested `20rem` and then asked whether the pane could have a dynamic height instead. It can, and that is the better answer: a cap in `rem` is a guess at a window height.

Step 1's card is now capped at the window and the agreement is what gives. The cap is handed down the same three boxes the reading position is already measured through — step, pane, document — and everything else in the card is pinned `flex: none`, or a squeezed card would squash all of it a little instead of the one thing that can spare it. It applies on `.at1` alone: step 2 has no scroller, and capping it would cut its bottom off — this same bug, moved one step along.

The short-viewport rule also loses its `60rem` width bound. It was written for a phone turned sideways and duly excluded a 1280×540 desktop window, which has the same problem and got the keyhole instead.

Measured at 1440 wide:

| window | card | pane | Continue | page |
|---|---|---|---|---|
| 900 | 735px | 384px | in view | fits |
| 760 | 728px | 377px | in view | fits |
| 660 | 628px | 277px | in view | fits |
| 600 | 568px | 217px | in view | fits |

1280×540 now releases the pane and reads as a phone does. The read-to-end gate is shut at load and open at the end in all five cases, including across a window resize and a language switch mid-read, and every element of the phone and desktop-tall layouts is pixel-identical to before.

## Checks

`npm test` passes, `tools/lint-templates.sh` passes, a fresh purgecss regeneration leaves `bootstrap.min.css` unchanged. Both pages deployed and exercised on a lab hi3516ev300.


openipc-ai:
Third item from the same comment added: **the release build now strips comments from the static HTML.**

`build-dist.sh` minifies every hand-written `.js` and `.css` before the firmware image gets it, but the four self-contained pages majestic serves before there is a session keep their stylesheet and script *inline*, so neither loop ever reached them — they shipped exactly as written. On `setup.html` that was 22 KB of comment in a 59 KB page, on the one page an unclaimed camera serves.

`html-minifier-terser` now runs over `www/*.html`: comments plus the inline JS and CSS, deliberately **not** `collapseWhitespace` (measured at another 578 bytes out of 55 KB — 0.6% — for a transform that can take a meaningful space out from between two inline elements). **98,649 → 56,454 bytes** across the four. The `.cgi` templates are untouched and must stay so; they are haserl.

One comment survives: `setup.html`'s note to AI agents about who may accept the EULA, which is addressed to whoever *fetches* the page rather than to whoever edits the file. It now opens `<!--!`. Three things hold that up, because losing it would be silent — the build fails if a marked comment does not survive the minify, `tests/setup-page.test.js` pins that the note is present *and* carries the marker (the case the build cannot see: reformat the comment, drop the `!`, and 0 == 0), and CLAUDE.md says why the byte is there.

Verified by rendering the *built* pages: at five viewports the minified `setup.html` is identical to its source in every element's geometry and in the read-to-end gate's state at load and at the end; `login.html` and `cameras.html` are element-for-element identical on a desktop and a phone, with no error either version does not also raise.

openipc-ai:
Review finding addressed — the diagnosis was wrong, the fragility it points at is real, and it is now closed.

`#step1 > *` and `#pane` carry the **same** specificity: one id each, the universal selector counting for nothing. So the winner is whichever is written last, and `#pane` is — deliberately. Measured at 1440×660 on the reviewed commit, computed styles:

```text
.card > #step1              shrink 1   h 473
  #step1 > #l-eyebrow       shrink 0   h  13
  #step1 > #pane            shrink 1   h 279     <- content height is 386
  #step1 > label.check      shrink 0   h  18
  #step1 > #continue        shrink 0   h  37
    #pane > #doc            shrink 1   h 277
```

The pane does shrink, and the checkbox and Continue are in view — which is what the sweep in the PR body already showed.

But a rule that works only while it is written last is one a tidy-up disarms in silence, and the described failure is exactly what a reordering gives. Same measurement with the two blocks swapped:

```text
  #step1 > #pane            shrink 0   h 386     <- card capped, nothing gives
```

Card capped, nothing inside it giving, step 1 overflowing the bottom of a page that no longer scrolls — worse than the bug this PR fixes, and invisible in the diff that would cause it.

Fixed by naming the box that gives rather than leaving it to outrank a sibling: `#step1 > :not(#pane)`, and `.card > :not(#step1)` for the same reading one level up. Measured in **both** orders now — pane shrink 1 and 279px either way. 3b231a9.

The viewport sweep, the read-to-end gate and the phone/desktop geometries are unchanged by the hardening.

Issue #330, opened 2024-09-06
Motion Detector Errors


CAMOKATBETEPAHA:
ssc338q (infinity6e family), imx415, 16 MB nor flash, 2.4.09.05-lite, openipc-ssc338q

1. If you select several regions in the visual editor, then motion detection and the call of the script motion.sh does not occur. Also, writing to the debug does not work. In the visual editor, detection zones are displayed in green, as expected.

2. If you select several regions in the visual editor, then the values ​​​​of the sequence without a space are entered into the region of interest, that is, `0x0x10x10,9x9x19x19` instead of `0x0x10x10,  9x9x19x19`, which leads to ignoring the entered data. In the visual editor, the entered zones disappear after saving.

3. In the visual editor, it is possible to enter the size of the zone with a zero value, which is incorrect. It is not visible. That is, `100x100x 0x10 ` is a size error.

4. Entering coordinates and size is not limited by the image resolution field. This leads to the possibility of going beyond the size

5. With a portrait aspect ratio (90 or 270 degrees), when triggering zones at the bottom of the screen, the red lines are not erased after the end of the rendering
![Снимок экрана 2024-09-04 165250](https://github.com/user-attachments/assets/da48ece0-fe3f-42ac-bb14-93275f4583d5)

openipc-ai:
@CAMOKATBETEPAHA — sorry this sat for a while. All five points have now been re-checked against the current WebUI, whose motion-detection editor was rebuilt since you filed this (the regions are drawn on the live picture now, and the separate "Visual editor" page is gone). They did not all turn out to be the same kind of problem, so here is each one.

Two of them are not the visual editor at all, which is where the earlier triage on this issue sent them.

---

### 2. Regions joined without a space, and zones disappearing after save — **fixed**

This one is genuinely gone. Regions are no longer squeezed into a single comma-joined string anywhere: the page sends them as a list, and the camera stores them as a list, one entry per rectangle:

```yaml
motionDetect:
  roi:
  - 0x0x100x100
  - 900x400x300x300
```

There is no separator left to get wrong. Verified end to end on a camera — three regions typed, saved, still there after a reload, and correct in `majestic.yaml`.

### 3 and 4. Zero-sized zones, and coordinates beyond the image — **now visible; not yet prevented**

[...]

PR #332, opened 2026-09-04
Motion detection: a region the camera will ignore says so before it is saved

openipc-ai:
Picks up #330 ("Motion Detector Errors", ssc338q + imx415), re-checking all five reported points against current `master` and the motion editor as it was rebuilt in #297 / #299. Two of them are this repo's and are fixed here; one was already fixed; two are firmware-side and are tracked there.

## Points 3 and 4 — still exactly as reported, and fixed here

> it is possible to enter the size of the zone with a zero value, which is incorrect. It is not visible. That is, `100x100x 0x10` is a size error
>
> Entering coordinates and size is not limited by the image resolution field. This leads to the possibility of going beyond the size

Measured in headless Chrome against an hi3516ev300, typing into the coordinate boxes:

| typed | chip said | Save |
|---|---|---|
| `100x100x0x10` | `0%` | enabled |
| `9000x9000x100x100` | `1%` | enabled |
| `-5x-5x100x100` | `1%`, rectangle drawn at negative offsets | enabled |
| `0x0x99999x99999` | `1085048%` | enabled |

All four were then accepted by `POST /api/v1/config` and written to `majestic.yaml` verbatim, with no log line. The drag path clamps properly and always did; only typing was unbounded.

**The cost is not a wrong number.** Each region edge is clamped into the frame before anything is compared against it, so a zero-size region and one entirely off the picture *both* end up as a rectangle with no area, which no movement can ever fall inside. And the region list is a whitelist with no fallback — what gates the filter is the regions that **parsed**, not the regions that can ever match. So one mistyped region does not narrow the watch, it ends it, while the camera goes on streaming, recording, answering ONVIF and reporting healthy counters. This page is the only place anyone can be told.

### What the rows say now

```
REGIONS                              5 regions · 4 unusable  ↺
 1  0x0x400x400                                        17%  ×
 2  100x100x0x10                                   no area  ×
 3  9000x9000x100x100                          off picture  ×
 4  1100x600x400x400                          2% · clipped  ×
 5  nonsense                                   not XxYxWxH  ×
A region with no area, or one outside the picture, is saved but never watched.
Coordinates are main-stream pixels — 1280 × 720.
 + Add by coordinates    🗑 Clear all
```

- `N% · clipped` is scored on the **clipped** rectangle, because that is the part being watched; scoring the typed one overstates the watch by exactly the part the camera throws away.
- `not XxYxWxH` now also covers the minus sign and the decimal point, because the camera cannot store either. `-5` came back from the camera as 4294967291 while the editor drew minus five: neither was what was typed, and only one was on screen. Inner blanks stay valid — the camera accepts those, so `0x0x10x 10` *is* a region it stores and calling it invalid would be the page inventing a rule.
- The head's count updates per keystroke. It used to be written only by the full paint, so it lagged one row behind the list beside it — three rows on screen under a head reading "2 regions".
- The consequence line is worded harder when **every** region is unusable, since that is the only case that inverts the feature rather than narrowing it.
- The coordinate space is stated. The editor always knew it and never printed it, which is the whole of why "not limited by the image resolution field" was possible to walk into.

Privacy masks go through the same control and get the same verdicts with their own nouns — a mask that hides nothing is not a region that watches nothing, and those two callers share everything except the sentences.

Nothing is silently repaired: a number somebody typed from a spec is not rewritten behind them, it is named where they would fix it.

## Point 2 — already fixed, verified

> the values of the sequence without a space are entered into the region of interest, that is `0x0x10x10,9x9x19x19` instead of `0x0x10x10,  9x9x19x19`, which leads to ignoring the entered data. In the visual editor, the entered zones disappear after saving.

There is no comma-joined encoding left anywhere to get wrong: regions are a real YAML/JSON array now, one string per element, and the page posts them as an array. Verified end to end — three regions typed, saved, survived a reload, and landed in `/etc/majestic.yaml` as a block sequence.

## Points 1 and 5 — not this repo's

The only comment on #330 sent points 1–4 here. Point 1 is not the editor's: the regions reach `majestic.yaml` correctly and are parsed correctly, and it is the firmware's region filter that discards the events. Both of these reproduce only on SigmaStar, which is why the reporter (ssc338q) hit them and HiSilicon users do not — verified as working on an hi3516ev300 with the same configuration.

Both are diagnosed and tracked firmware-side, outside this repository. They should stay open on #330 until the firmware fixes land; nothing in this PR addresses them.

## Testing

`tests/region-verdict.test.js`, over the extracted `www/a/mj-region.js`. It earns its place the way `ircut-check` does: every branch renders a confident-looking chip in the same quiet grey, so a wrong one does not degrade this fix, it reverts it — and reaching most of them on a real camera needs one at a known main resolution, a frame size that has already arrived from the codec event, and somebody willing to type what the editor exists to talk them out of. The numbers in it are the ones the camera stores and the ones the lab camera reported.

The geometry moved out of `mj-settings.js` for that reason alone, following `mj-requires.js`. Both callers now ask one question (`canRegion`) rather than two, so a missing module cannot hide the coordinate rows and then mount nothing over them — the field falls back to plain coordinate boxes with a sentence naming which half is missing, instead of blaming the preview on a camera whose preview came up fine.

`npm test` passes; `tools/regen-bootstrap-css.sh` reproduces the committed `bootstrap.min.css` byte for byte (the new class is in the override, which is not purged).

## Verified on

hi3516ev300 + imx335, main stream 1280x720. Before and after: the four typed values above, a drag that creates a region, a drag that shoves one past the right edge (clamped at `880x181x400x400` on a 1280-wide frame), a save round trip, and the Overlay leaf's masks still mounting with the raw row hidden and no page errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build:dist": "sh tools/build-dist.sh",
-    "test": "node tests/requires.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/recordings-health.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/charts.test.js && node tests/mem-slices.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js && node tests/preview-hero.test.js"
+    "test": "node tests/requires.test.js && node tests/region-verdict.test.js && node tests/talkback.test.js && node tests/transport.test.js && node tests/staging.test.js && node tests/auto-source.test.js && node tests/mp4index.test.js && node tests/timeline.test.js && node tests/recordings-pump.test.js && node tests/recordings-health.test.js && node tests/reset-watch.test.js && node tests/luma.test.js && node tests/metrics-parse.test.js && node tests/charts.test.js && node tests/mem-slices.test.js && node tests/preview-socket.test.js && node tests/preview-stats.test.js && node tests/video-check.test.js && node tests/ircut-check.test.js && node tests/ircut-scan.test.js && node tests/setup-page.test.js && node tests/preview-hero.test.js"
   },
   "devDependencies": {
     "clean-css-cli": "5.6.3",
--- a/tests/region-verdict.test.js
+++ b/tests/region-verdict.test.js
@@ -1,0 +1,169 @@
+// The region editor's verdict on a rectangle somebody typed.
+//
+// This exists because the subject fails silently, and in the direction that
+// costs the most. Every branch renders a confident-looking chip beside the
+// coordinates -- the version this replaced printed `0%` for a region of zero
+// width, `1%` for one nine thousand pixels off the picture and `1085048%` for
+// one bigger than the sensor, all in the same quiet grey as a region that
+// works -- so a wrong branch does not degrade the fix for
+// OpenIPC/majestic-webui#330, it reverts it.
+//
+// It also cannot be reproduced on demand: reaching most of these branches on a
+// real camera needs one at a known main resolution, a frame size that has
+// already arrived from the player's codec event, and someone willing to type
+// coordinates the editor is supposed to talk them out of. The numbers below
+// are the ones majestic actually stores and the ones the lab camera actually
+// reported.
+'use strict';
+
+const path = require('path');
+const { check, group, done } = require('./assert');
+
+const R = require(path.join(__dirname, '..', 'www', 'a', 'mj-region.js'));
+
+// The lab camera's main stream, and a rotated one: portrait is where a typed
+// coordinate is most likely to be out of bounds, because the two numbers swap
+// and nothing on the page says so.
+const MAIN = { w: 2592, h: 1520 };
+const PORTRAIT = { w: 1080, h: 1920 };
+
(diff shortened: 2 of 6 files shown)


openipc-ai:
Both findings are accurate about what the PR does: it names an unusable region, it does not stop one being saved. Taking them as a decision rather than a defect, and recording why here.

**They are the two alternatives this PR's own summary weighed and rejected**, and I agree with that reading:

- **Blocking the save** is section-wide. Save carries Enable, Sensitivity, Visualize and Debugging as well as the regions, so one mistyped row would hold four unrelated settings hostage — including on a camera where the bad row is *pre-existing*, typed a year ago or left behind by a resolution change, where the operator came to toggle something else entirely and cannot.
- **Clamping** rewrites numbers somebody typed from a spec, behind them. And it cannot repair the case the reporter actually hit: `100x100x0x10` has no area, and there is no honest width to invent for it — the drag path's 16px floor is a gesture minimum, not a guess at intent.

There is a narrower third option worth naming, and I would rather offer it than assume it: **refuse the save only while the regions field is itself dirty**, so you cannot make the list worse in this sitting, but an already-broken region does not block a save that has nothing to do with it. That is a real prevention and it costs nothing to the operator who is not editing regions. It is not in this PR because it changes what the Save button does across two leaves, which deserves its own review rather than riding along in a bug fix.

What is not in doubt is that the camera's behaviour is now visible where it was not: the row says which rectangle the camera will do nothing with, the head counts them, and the sentence under the list says what that costs — worded harder when *every* region is unusable, since that is the case that turns detection off entirely rather than narrowing it. Before this, the same rectangle read `0%`, `1%` or `1085048%` in the same quiet grey as one that works.

Because it is a warning rather than a rejection, this PR does not close OpenIPC/majestic-webui#330 — points 3 and 4 stay open for the reporter to judge whether being told is enough, and the block-on-dirty variant is there if it is not.

PR #333, opened 2026-09-04
Stop naming the firmware's own source, and gate it so the next PR cannot

openipc-ai:
Four commits: take every reference to majestic's closed source out of this public tree — paths first, then the symbol names — and make the rule a check instead of a promise.

## 1. Five comments named files nobody outside the project can open

Each was there to make a claim checkable, but a reader outside the project cannot open any of them — so the path was never the evidence, it was a pointer to a repository they do not have, attached to internals they were not meant to have either. It also dates badly: the file moves, the symbol is renamed, and the sentence that cited it can no longer be checked by anyone at all.

| where | was | now |
|---|---|---|
| `CLAUDE.md:17` | two function names and the file holding them | "the question it asks is whether authenticating root against an empty password succeeds" |
| `CLAUDE.md:81` | three places in a named source file | "three places in its own web server" |
| `CLAUDE.md:199` | `OSD_CODES` is the set from a named source file | "the set the camera's own overlay formatter accepts" |
| `www/cgi-bin/status.cgi:7` | three places in a named source file | "three places in its own web server" |
| `www/a/mj-settings.js:2681` | the same two tables a named source file indexes | "the same two tables the camera's own overlay placement indexes" |

Every claim is kept and none is weakened — what a reader can verify is unchanged, because what they could verify was never the filename. `OSD_CODES` keeps its name: that is this repo's own constant, and only the source it was transcribed from is dropped.

Prose only in this commit: no behaviour, no markup, no test touched.

## 2. The gates

Removing them fixes the tree and does nothing about the next one — and this is the class of mistake where the next one is expensive, because a commit message on a public branch cannot be rewritten and editing a merged PR's description leaves the original in its edit history. The fix has to run before the merge or not at all.

So `.pr_agent.toml` + `pr_compliance_checklist.yaml`, copying the shape of the ruleset in the firmware repo (prose standards, binary gates). Qodo already reviews this repository, so the config takes effect on the next PR.

Two of the seven gates are the disclosure ones, and both are declared to apply **to the PR's own title and description as well as the diff** — a description is as public as the code, and that is where this went wrong:

- **No references into a sibling repository's private source** — no path inside another repo's tree, no private repo name or URL, no link or `#number` into a private tracker (including one written as `OpenIPC/majestic#N` for an `N` that does not exist in that public tracker), no internal firmware symbol, no quoted excerpt.
- **No lab environment details** — no lab hostname or domain, private IP literal, credential, SSH alias, private build branch or commit id, or pasted console output from a lab device.

Both are written to say what **is** allowed, not only what is not: firmware behaviour described as something observed from outside the camera and cited to a public endpoint, config key, metric or tracker issue; verification described by hardware class and configuration ("measured on an hi3516ev300 + imx335, main stream 1280x720"), which is what makes a result reproducible, rather than by the name of the machine it ran on.

The other five gates are existing house rules that were only ever written in `CLAUDE.md` prose, where a reviewer has to remember them — an absent reading is not a zero and a failed fetch is not a fact; device-derived output goes through the escaper that suits where it lands; busybox-only shell; no bundler, framework or CDN asset; a page is named once in `p/pages.cgi`. Nothing new is invented: each is already in the guidance or enforced by `tools/lint-templates.sh`.

Both files parse (`yaml.safe_load`, `tomllib`) and every gate carries the five keys the schema wants.

## 3. And the symbol names

Fourteen more identifiers out of the closed source were still quoted — the config walker and its setter, the SDK reload, the file write-back, the default-declaring macro, the empty-password check, the setup handler, the anchored-placement routine, the percentage-offset arithmetic, the live reload class — across `CLAUDE.md`, `www/a/mj-settings.js` and `www/cgi-bin/p/common.cgi`. Same fault as the paths and the same reason: a reader outside the project cannot look any of them up, so the name was never the evidence, and it rots when the symbol is renamed while the sentence around it stays confident.

All of them are now stated as behaviour — "walks and sets every leaf, then reloads the SDK and writes the file back exactly once", "authenticates against the new password to check it landed", "the camera's anchored placement centres it outright". No claim changed and none got weaker.

**One correction rides along**, because a gate added in this branch contradicts it. `p/common.cgi` has **four** escapers, not two: `ex` and `pre` render blocks, `esc` returns escaped text for inside existing markup, `attr_escape` does the same for a quoted attribute. `CLAUDE.md` called `ex` and `pre` "the only safe ways to render shell output" — which is where the too-narrow first draft of the escaping gate came from, and would have gone on sending people to a block renderer for an inline value.

## Not in this PR

- **`best_practices.md`.** The judgement-based half deserves the same treatment as the firmware repo's — written from bugs that actually shipped here — and is a larger piece of work than this.
- `npm test` and `tools/lint-templates.sh` pass.


openipc-ai:
Both findings were right and both are fixed in e36a3ad.

**1. The gate published the symbols it prohibits.** Correct, and the sharper point is that it happened in the same commit that wrote the rule down, after an afternoon spent removing exactly that from this tree — which is a fair measure of how easy the mistake is. The criterion now describes the syntactic shape (a C-style identifier attributed to the camera rather than to this tree) and states that it gives no examples, so nobody helpfully adds some back.

**2. `ex`/`pre` are not the whole set.** Also correct, and verified against the source before changing anything: `ex` wraps a command and its output in a `<div><h6>…<pre>`, `pre` emits a `<pre>`, so neither can carry a value that has to sit inside existing markup or inside a quoted attribute — which is what `esc` and `attr_escape` are for, and what `dashboard.cgi` and `vtun.cgi` already use them for. As written, the gate would have failed correct code. It now names all four and says which lands where, in both the checklist and `.pr_agent.toml`.

Worth recording that the second one is inherited rather than invented: `CLAUDE.md` says `ex` and `pre` are "the only safe ways to render shell output", which is where the wording came from and is wrong about it. That belongs in its own change rather than this one.

PR #334, opened 2026-09-04
Notices: one banner shape, and it stops being the Dashboard's alone

widgetii:
The redesign did build a notice component — `.st-alert` on the Dashboard — and it is right. It was just declared inside the block headed `/* Dashboard (dashboard.cgi) */`, between `.st-panel` and `.st-kpis`, and named for that page. Nothing else could reach it without renaming or copying it, so nothing did, and every other page went on emitting Bootstrap's filled `.alert`.

It moves out and becomes `.mj-notice`. Three things change with it, and two defects go with them.

### The two defects

- **`.st-alert-ico` hard-codes `color: var(--bs-warning)` with no severity in the selector**, so the Dashboard's *informational* notice wears the warning colour. Every finding on that page already carries a `level`; the page threw it away.
- **On the light theme that amber measures 1.63:1** against the white card. Below the 3:1 floor for a graphical object — not a weak mark, an absent one.

### What changes

- `align-items` goes `center` → `flex-start`, which is what lets a wrapped sentence keep its mark against its first line and lets a body row exist. `a/sdcard.js` has wanted one all along: it writes a title, a paragraph, three `dmesg` lines and a button into one `alert-danger`.
- Severity is a class driving the rule and the mark together, and the marks are drawn rather than typed — a glyph paints whatever size the font decided, which is how a 13×12px severity mark ended up beside a 16×16px close button.
- The actions become a group. Flexbox has no `break-before`, so with flat children whether the row wraps on a phone depends on how long the sentence happens to be: `.st-alert`'s `10rem` basis is a *minimum*, so a short sentence beside a wide button still fitted on one line and was squeezed to four words a line instead of breaking.

Colour is Bootstrap's own `<sev>-text-emphasis` and the rule is 60% of it. Measured against both cards all eight combinations land between **3.0:1 and 5.2:1**, which a single alpha on the base hues cannot do — `#ffc107` on white tops out at 1.46:1 even at 75%, `#dc3545` on the dark card at 2.55:1.

### The MAC banner loses its form

`network.cgi` has carried a **Change MAC address** card all along, posting the same `changemac` action to the same handler — the banner was a second copy of a form one click away. 194px on every page at 1280 and 500px on a phone, against 63 and 180 now. The generator that only ever lived inside the banner moves to that card, and its guard is fixed on the way: it compared against the empty string, right in a banner whose field starts empty, and would have refused every time on a card pre-filled with the address in use.

Two more: the settings page's stopped-daemon banner stops offering the wiki's front page as "more information", printed as plain text rather than a link — the log is where majestic says why it died. And the pending-restart banner drops from danger to warning; it keeps `btn-danger` on its button, because `main.js` hangs the confirmation prompt off that class.

### Left alone deliberately

The two `.mj-stage-alert` overlays in `common.cgi`'s `preview()` (drawn over the black video stage); `mj-settings.js`'s compact findings and verdicts (they sit *on* a card, where a card-surface notice has no surface to lift off); `time.js`'s inline status pill; and `login.html` / `setup.html`, which are self-contained because an unclaimed camera serves no `/a/*` at all.

### Measured

On a hi3516ev300, both themes, headless Chromium against the camera's own stylesheets:

| | today | after |
|---|---|---|
| MAC banner @1280 | 194px | 63px |
| MAC banner @390 | 500px | 180px |
| restart banner | 154px | 52px |
| stopped daemon | 91px | 63px |
| all seven, stacked @1280 | 767px | 394px |
| all seven, stacked @390 | 1193px | 821px |

Three of the one-line notices get *taller* on the phone: the action costs a line, and two of them now say what the fault costs you.

### Tests

`tests/notice.test.js` is new. The component is emitted twice and has to be — by haserl for the banners and flash messages, by `main.js` for the three built in the browser — and two emitters of one component is the arrangement that drifts, so the test asserts they are byte-identical. Every way it can go wrong is silent: a class the stylesheet does not know renders as an unstyled div, and the call sites that would show it are a card gone read-only and a stopped daemon.

Design canvas the change was drawn against: https://claude.ai/code/artifact/3dc13faf-bae5-4946-80f5-f993b282af3b


widgetii:
Good catch on the gateway banner — that consequence was mine and it was not measured. A configured NTP server on the camera's own subnet is reachable with no default route at all, so a static-network camera pointed at a local one would have been told on every page that its clock was broken when it was not.

Fixed in 9d53b49: the banner now states what the page actually knows — there is no default route, so nothing outside the local network is reachable — and leaves the reader to tell whether their time server is on the other side of it.

PR #336, opened 2026-09-05
Day / Night: a panel visible only when it is already on, and a chart drawn in the wrong stylesheet

### LINKED ISSUE #325: Day / Night: the camera can watch the light by itself — automatic switching, nothing wired, nothing calibrated
@usa- — one more from the thread that rebuilt this section, and I would value your eye on it again. Back in #273, when the "Exposure at maximum" banner was retired, the honest half of that answer was a promise: a version that could genuinely act on the light needed the cameras to report exposure comparably first, and that work was not started. It is finished now, and it grew into a feature. The WebUI half is on `master` (`updatewebui` has it); the camera half arrives with the next nightly firmware build, so this is testable from tomorrow.

**What changed.** People without a light sensor have been switching day/night with shell scripts — there is a whole folder of them in [OpenIPC/sandbox](https://github.com/OpenIPC/sandbox/tree/main/scripts/backlight-control), each polling the camera's gain gauge against two hand-picked magic numbers that only fit the one camera they were tuned on. That job moved into the camera:

- **Turn on the light monitor and set nothing else** — no sensor pin, no thresholds. The camera goes to night when its sensor runs out of shutter and gain (a statement that means the same thing on HiSilicon, Ingenic and SigmaStar alike), and returns to day when the gain settles back to daylight levels. No numbers to pick, no calibration.
- **The script authors' knobs survive, in portable units.** If you would rather set your own line, the night and day thresholds are gain *multiples* now — the same meaning on every camera — with the switching delays beside them. Hysteresis and an anti-flapping guard are built in: headlights at night and a passing cloud at noon are ignored, and fog cannot make it strobe.
- **The Day / Night page shows the monitor's reasoning.** A live chart of the value it watches, with the day and night bands shaded, and one sentence about what it is doing right now — including a countdown when a switch is pending: *"Bright enough for day — switching in 40 s if it stays."* The dashboard's day/night line now also says who decided: `auto`, `sensor pin`, `thresholds`.
- **The findings know the difference** between a camera watching its own exposure (an observation), a SoC that cannot report one (wire a sensor pin instead — the page says so), and older firmware (an update fills the gap).

**Worth trying, once the nightly lands:**

1. Update the firmware and `updatewebui`, open **Settings → Day / Night**, turn on the light monitor, set nothing else, Save.
2. Watch the **Light monitor** block: the chart should trace the gain with the green day band under it, and the sentence should say it is watching.
3. Cover the lens for a minute — the countdown should appear, then the camera should go to night after ~15 s of real darkness. Uncover it: back to day after ~60 s of brightness. Brief cover, or a hand waved past, should do nothing at all.
4. If you have an IR lamp wired, the interesting half is a real dusk and dawn: the whole point of the defaults is that the lamp switching on does not talk the camera back into day. If it flaps anywhere, the chart will show exactly where the gain sat when it happened — that picture is the most useful thing you could attach.

**Two things to know.** A camera already using a sensor pin or the old threshold pair keeps exactly its current behaviour — the new mode only takes the cases where nothing was configured. And on SoCs that report no exposure state (Allwinner, Rockchip, Ingenic T40) the monitor says it cannot run rather than guessing.

**What I would most like your opinion on:** whether the chart and its one sentence explain themselves to someone who never read this issue — the countdown wording especially — and whether the default timings feel right at your actual dusk: too eager, too sluggish, or invisible, which is what they are meant to be.

openipc-ai:
The reporter of #325 could not find the Light monitor block on one camera and found it on another, and read that as the feature not having shipped for their processor. Their two attached configs answer the question by themselves: the camera without the block has automatic day/night **switched off**, the one with it has it **on**. Same code, different switch.

That is the bug rather than the explanation for it — twice over.

### A panel that appears only in the state it describes

`monitorView()` returned `null`, and took the whole block off the page, in every state except the two that draw a graph: the switch off, a wired photocell, an ADC pad, a monitor that stood down, an older daemon. Each of those is a defensible "there is no graph here". Together they meant the only section on the page that **names the control** was visible only to people who had already found the control.

It answers in every state now, with `chart: false` where there is nothing continuous to plot, and hides itself only while the camera has not answered at all:

| state | what it says |
|---|---|
| switch off | *Off — nothing switches this camera between day and night on its own. Turn on Automatic day/night below and this panel says what it is watching.* |
| daylight sensor pin | *Night. The daylight sensor decides. It reports light or dark and nothing in between, so there is no reading to plot here.* |
| ADC pad | *Day. A voltage on the daylight sensor pad decides. The camera does not publish that voltage, so there is nothing to plot here.* |
| stood down | *Day. Nothing is deciding: this camera reports no exposure to watch, and no daylight sensor or threshold is set.* |
| older daemon | *This firmware does not report what the monitor is watching.* |

### And the words follow the switch

The block heading, four findings, the filter test's refusal and the Live tab's chip all said "the light monitor" — which is what the firmware's own hints call it, and what nothing on screen is labelled. The switch is titled **Automatic day/night**. Searching the page for the thing you have just been told to turn on has to find it, so that is what they all say now.

### The chart was drawn in no stylesheet at all

`charts.js` was lifted out of the dashboard so other pages could draw with the same pencil, but its rules stayed behind under the dashboard's own `.st-chart` prefix. The third caller matched nothing, so its SVG text took the page's body font and SVG's default fill. Measured in the page:

| | unclassed host (what shipped) | styled host |
|---|---|---|
| font | `16px Montserrat` | `10px` PT Mono |
| fill | `rgb(0, 0, 0)` | `var(--bs-tertiary-color)` |
| display | `inline` | `block` |
| width | 300px intrinsic | 100% |

The axis numbers were about 1.7x wider than the 25px gutter the renderer reserves, so `2000` came out as `)00` and `150000` as `0000`, in black on a dark card — which is exactly what the reporter's screenshot shows. The rules are named after the thing now (`.mj-chart`) and every host carries it, the dashboard's five and the stats panel's included; the panel keeps its own ink and says so with both class names, because equal specificity would have left the winner to source order. The series colour and the grid hairline are read from the same theme tokens the dashboard reads, rather than being the light theme's, hardcoded.

### Three in `charts.js`

All visible in the reporter's own screenshot, all reproduced on an hi3516ev300 + imx335 running a 2026-09-04 nightly.

- **The auto ceiling fitted the data alone.** Gain idles near 1024 in raw units while the legacy thresholds people actually set are around 150000, so both bands clamped flat to the top edge and the plot became a flat line in a uniformly tinted box — no answer at all to the one question it is on the page for. Finite band edges count toward the ceiling now. The 15% headroom stays with the data, which needs somewhere to rise into; a threshold is a fixed number, and giving it headroom pushed a day band ending at 2 out to a ceiling of 5 and left it filling the bottom two fifths of its own chart.
- **"Everything above the night threshold" was spelt `1e9`**, which is indistinguishable from a real edge and would drag the whole plot off the top once the scale started fitting edges. An open band end is `null`.
- **Both band names were drawn at the same offset whatever the band's drawn height**, so two bands clamped to one edge printed "day" and "night" on the same pixel. A band shorter than a label draws none — and since bands do not overlap, that single guard also keeps the survivors a line apart.

Plus: the left gutter sizes itself to the numbers about to go in it, instead of staying at the four monospace characters that were enough for a percentage and not for a six-digit gain.

### Verified

On an hi3516ev300 + imx335 (lite, 2026-09-04 nightly), in four configurations:

- automatic — axis `2.0x / 1.0x / 0.0x`, day band shaded, gain trace drawn;
- thresholds 150000/160000 — axis `200000 / 100000 / 0`, both bands separately labelled, the gain visibly far below its night threshold (this is the reporter's remote camera's shape);
- switch off, and daylight-sensor pin — sentence, no chart, no empty reserved box.

Dashboard charts and the Live stats panel render as before. 8 new tests across `tests/charts.test.js` and `tests/ircut-check.test.js`; suite green, `npm run build:dist` green, the purged bootstrap subset regenerates identical.

Fixes #325

### DIFF
--- a/tests/charts.test.js
+++ b/tests/charts.test.js
@@ -286,4 +286,70 @@ group('window and gap are per chart, because not every subject is a subject of n
 		(narrow.innerHTML.match(/M/g) || []).length > 4);
 }
 
+group('a band is what the value is compared against, so the scale has to fit it');
+{
+	const { MC, clock } = boot();
+	// The Day / Night light monitor's shape, measured on an hi3516ev300: gain
+	// idling near 1024 while the day and night thresholds sit at 150000 and
+	// 160000. Scaled from the data alone the plot topped out at 2000, both
+	// bands clamped to the top edge, and the chart drew a flat line in a
+	// uniformly tinted box — no answer at all to "how far is this from
+	// switching" (#325).
+	const bands = [
+		{ from: 0, to: 150000, color: 'rgba(47,182,115,.10)', label: 'day' },
+		{ from: 160000, to: null, color: 'rgba(255,193,7,.10)', label: 'night' },
+	];
+	const host = makeHost(400);
+	const ch = MC.makeChart(host, { h: 110, lo: 0, hi: null, colors: ['#000'], bands: bands });
+	for (let i = 0; i < 3; i++) { clock.t = i * 2; MC.pushChart(ch, [1024]); }
+	const ticks = (host.innerHTML.match(/text-anchor="end">([\d.]+)</g) || [])
+		.map(t => parseFloat(t.replace(/[^\d.]/g, '')));
+	check('the top of the plot clears the night threshold',
+		Math.max.apply(null, ticks) >= 160000,
+		'top tick ' + Math.max.apply(null, ticks));
+
+	// Both band labels used to be drawn at y1 + 10 whatever the band's drawn
+	// height, so two bands clamped to the same edge printed "day" and "night"
+	// on the same pixel — one unreadable smudge in the corner.
+	const labelY = w => {
+		const m = host.innerHTML.match(new RegExp('y="([\\d.]+)"[^>]*>' + w + '<'));
+		return m ? parseFloat(m[1]) : null;
+	};
+	check('and both band names are on the plot, a line apart',
+		labelY('day') != null && labelY('night') != null &&
+		Math.abs(labelY('day') - labelY('night')) >= 12,
+		'day at ' + labelY('day') + ', night at ' + labelY('night'));
+
+	// An open end is not a number: 1e9 as a stand-in for "and above" would be
+	// folded into the scale above and push the whole plot off the top.
+	check('an open band end is not fitted into the scale',
+		Math.max.apply(null, ticks) < 1e6, 'top tick ' + Math.max.apply(null, ticks));
+
+	// The gutter has to hold the numbers it is about to print. At a flat 30px
+	// — four monospace characters — a six-digit threshold lost its first two
+	// digits off the left edge and read as "0000".
+	const gutter = html => {
+		const m = html.match(/text-anchor="end">[\d.]+<\/text>/) &&
+			html.match(/<text x="([\d.]+)" y="[\d.]+" text-anchor="end">[\d.]+</);
+		return m ? parseFloat(m[1]) : null;
+	};
+	check('the gutter widens for a six-digit axis',
+		gutter(host.innerHTML) >= 6 * 6, 'label right edge at ' + gutter(host.innerHTML));
+	const small = makeHost(400);
+	const ch3 = MC.makeChart(small, { h: 110, lo: 0, hi: 100, colors: ['#000'] });
+	MC.pushChart(ch3, [50]);
+	check('and a short one still gets the gutter it always had',
+		gutter(small.innerHTML) === 25, 'label right edge at ' + gutter(small.innerHTML));
+
+	// A band with no room for its own name does not print it anywhere.
+	const tight = makeHost(400);
+	const ch2 = MC.makeChart(tight, {
+		h: 110, lo: 0, hi: 100, colors: ['#000'],
+		bands: [{ from: 99, to: 100, color: '#fff', label: 'sliver' }],
+	});
+	MC.pushChart(ch2, [50]);
+	check('a band squeezed thinner than its label draws no label',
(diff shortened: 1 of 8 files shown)


openipc-ai:
Addressed in 19c1bff.

**The rule violation is a fair call and I have taken it.** Three of the four new chart assertions were about layout you can see in one page load — a clipped axis number, two band names on one pixel, a label in a band too thin to hold it. Fixed, not frozen; they are out of the suite.

**One stays**, and 19c1bff says why in the file: a scale that fits its data and ignores its bands is the same failure this file opens with, reached from the other side. The groups above it exist because a scale can flatten a *trend* into three pixels while drawing a perfectly confident line; this one flattens the *distance to the threshold*, which is the only quantity a banded chart is on the page for. Nothing renders an error, the trace is smooth, the box is tinted, and a camera nowhere near switching looks identical to one a step away. That is silent, and it needed the real measured shape (gain idling at 1024 against thresholds in the 150000s) to be seen at all.

**Separately, review turned up a real ordering bug** the summary did not: on a daemon that publishes no source gauge, `monitorView` reproduces the precedence from config alone, and a camera holding **both** a daylight sensor pin and the two legacy thresholds took the threshold branch. The panel then charted a gain gauge that was not deciding anything, under a sentence stating that it was. The pin decides first, as it does on the camera. That one *is* covered — reaching it needs an older firmware and a wiring combination nothing here can go and look at.

PR #343, opened 2026-09-05
Day / Night: three mechanisms decide the same thing, and the page said which one never

openipc-ai:
> *"And now the camera switches between day and night with no countdown at all. Are you guys testing this even minimally before merging the changes? :-)"*

Fair hit, but the countdown is not broken — and what is actually wrong is worse than a broken countdown.

### What their camera was doing

Day / Night has **three** mechanisms that decide the same thing, and the camera picks between them by what is filled in: a daylight sensor pin wins, then the pair of legacy raw thresholds, and only with neither of those does the calibration-free automatic mode run.

The reporter had set **Minimum sensor threshold for day** = 2000 and **Maximum sensor threshold for night** = 14000. So the camera was in the legacy mode — which has no countdown, because it has no pending state. Meanwhile they had also filled in **Gain multiple that means night** = 33, **Seconds of darkness before night** = 15 and **Seconds of brightness before day** = 60, all of which the camera was ignoring completely.

Reproduced on an hi3516ev300 with exactly that configuration: `night_mode_source` reports **2**, and **no `night_auto_*` gauge is published at all** — no streak, no dwell, no pending. There is no countdown to show and none is possible.

So four controls were doing nothing, and the page was silent about it. That is this UI's cardinal sin wearing a settings form, and it is also the reporter's second comment:

> *"there are now two sets of parameters that appear to control the same things … showing all of them together makes it very difficult for a user to understand which settings actually matter."*

### What the page does now

**It asks the camera which mechanism is deciding** rather than re-deriving the precedence from config, so a build that orders them differently gets described correctly instead of confidently.

- A control belonging to a mechanism that is **not** deciding, and that actually holds a value, gets a one-line note: *"Not in use — the day and night thresholds are deciding."* An empty control misleads nobody and is left alone.
- The finding at the top of the section carries the explanation, **including why no countdown appears**, and the way out.

**Not by hiding or disabling them**, which was the reporter's first suggestion. The value is real, it is the thing causing the surprise, and it has to stay findable and clearable — hiding it would only move the surprise. That is the argument `x-requires` already makes for itself in this codebase, and this reuses its note and its styling so there is one look for "the camera is quietly doing something else".

**And one line, not four paragraphs.** Repeating the explanation under every ignored control is the wall of text the reporter objected to in the same breath; the banner says it once.

### On their preferred fix

They would rather the legacy values were converted once and the fields removed. Two reasons that is not this repo's to do:

- The legacy thresholds are raw `isp_again` units and are **not portable** — Q10 on HiSilicon and SigmaStar, log2×32 on Ingenic — so a browser-side conversion to gain multiples would be right on one vendor and wrong on another.
- The daemon already ships the annotation that expresses exactly this (`x-requires`, with a `field`, an `equals` and a `message`). Declaring it on these keys would make every client show it, not just this page, and would state the precedence at its source. That belongs firmware-side and is worth raising there.

### Tests

Six checks in `tests/ircut-check.test.js` covering the finding: thresholds shadowing a filled-in automatic set, a wired sensor shadowing both, why the countdown is missing, and the three states that must stay quiet — automatic actually deciding, only the winning set filled in, and a camera that has reported no source at all. Reachable only with a camera configured two ways at once.

Suite green (24 files).

Refs #325 — not a closing trailer; the reporter has this to try and a standing complaint about hint length that this does not address.

### DIFF
--- a/tests/ircut-check.test.js
+++ b/tests/ircut-check.test.js
@@ -698,6 +698,35 @@ function runRest() {
 			ic.projector(6).age(1) === null);
 	}
 
+	group('diagnose: a mechanism that is filled in but not deciding');
+	{
+		// Three mechanisms decide the same thing, the camera picks one, and
+		// every control is on the page together. The reporter of #325 filled
+		// in a night gain multiple and both automatic delays while a legacy
+		// threshold pair kept the wheel, and nothing said so — four live-
+		// looking controls doing nothing, and no countdown, because the
+		// mechanism that has no countdown was the one running. Reachable only
+		// with a camera configured both ways at once.
+		const both = { lightMonitor: true, irCutPin1: 11,
+			minThreshold: 2000, maxThreshold: 14000,
+			autoNightGain: 33, autoNightDelay: 15, autoDayDelay: 60 };
+		const said = (nm, src) => ic.diagnose(nm, { night: 0, ircut: 0, src: src },
+			{ flips: 0, conflictS: 0 }).filter(x => x.id === 'mech-shadowed')[0];
+		check('thresholds winning over a filled-in automatic set is named',
+			/compares raw sensor gain/.test(said(both, 2).detail), '');
+		check('and it says why the countdown is missing',
+			/no countdown appears/.test(said(both, 2).detail));
+		check('a wired sensor winning over both is named as the sensor',
+			/daylight sensor is wired/.test(said(both, 1).detail));
+		// Nothing filled in, nothing being ignored, nothing to say.
+		check('automatic mode deciding says nothing',
+			!said(both, 4));
+		check('and neither does a camera with only the winning set filled in',
+			!said({ lightMonitor: true, minThreshold: 2000, maxThreshold: 14000 }, 2));
+		// A camera that has not reported a source is not accused of anything.
+		check('an unknown source says nothing either', !said(both, null));
+	}
+
 	group('diagnose: hunting, and the ordering of what it all reports');
 	{
 		const cfg = { irCutPin1: 11, lightSensorPin: 66, lightMonitor: true };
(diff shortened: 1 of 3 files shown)


openipc-ai:
Addressed in 276c38e. Both bugs were real and were the same fault this finding exists to end rather than to join.

**"Warning invents filled settings" — taken, and it was the worse of the two.** The detail named *"the automatic gain multiples and the delays beside them"* whatever was actually configured, on a path that could fire with no automatic value at all. Swapping silence for a confident wrong account of somebody's configuration is not an improvement. The sentence is built from what is there now: each shadowed mechanism is named only when it holds a value, and the clause about the missing countdown is spoken only about automatic mode, which is the only mode that has one.

**"Day gain omission hides warning" — taken.** The field map marked `autoDayGain` inert while the finding did not count it, so a camera holding only that one got a row note saying it was ignored and no section explanation at all — the two halves disagreeing about the same field.

That fix surfaced a third thing worth stating: the day gain multiple and both automatic delays **ship with defaults and are seeded on every camera**, so their presence is not evidence anyone asked for automatic mode. Severity follows intent now — a night gain multiple or a threshold pair is somebody having typed something and warns; seeded defaults are stated as info. Every ignored control still gets its row note either way.

**On the test-scope finding, I am not taking this one.** This file's stated charter in CLAUDE.md is precisely *"a decision table renders a confident sentence whichever branch it takes"* — that is what `diagnose()` is and what these checks cover, alongside the existing coverage of the same function immediately below them.

The evidence is this review. Both bugs you found are wrong branches of that table, each producing fluent, plausible, wrong prose, and neither raises anything anywhere — no exception, no console error, no failing render. A reviewer reading the diff would have to hold the whole precedence table in their head to catch either. That is the definition the rule is drawing.

What I removed under this same rule in #336 was different in kind, and I removed it without argument: assertions about label placement and gutter width, visible in one page load on any camera. The line is whether the failure announces itself, not whether a config can be arranged to reach it — by the latter reading the existing `monitorView` and `diagnose` groups in this file would all have to go too.

PR #345, opened 2026-09-05
Settings: a requirement that two settings have to agree on

openipc-ai:
`x-requires` names one controlling field. That is enough for the substitutions it was written for — `outgoing.substream` is inert exactly when `video1.enabled` is false, and nothing else bears on it.

Some rules are narrower than one equality can say. majestic now refuses live HLS on a camera whose recorder is in **motion** mode, because nothing is written between detections and the playlist has no new byte ranges to describe. It does **not** refuse it when the recorder is switched off: `records.mode` then describes nothing that is happening, and HLS serves out of memory exactly as it does on a camera with no card.

Written as "HLS requires `records.mode` continuous", the form would grey the switch out on that second camera and tell its owner to change a setting that is already inert — a warning about a configuration the camera deliberately allows, which nothing on the page can clear. That is the failure this module exists to prevent, arriving from the other direction.

### The shape

A condition may now carry `any`: a list of alternatives, satisfied when any one holds. That is what a requirement looks like when the rule it guards is a *coincidence* of two settings — the negation of "both are true" is "either is false".

```json
"x-requires": {
  "when": "true",
  "any": [
    { "field": "records.enabled", "equals": "false" },
    { "field": "records.mode", "notEquals": "motion" }
  ],
  "message": "Motion recording writes nothing between detections, so there is no live stream to serve."
}
```

majestic emits it with **no top-level `field`**, deliberately: this build reads a requirement without a controlling field as satisfied and stays silent, so a camera running the newer daemon against an older WebUI loses the hint rather than gaining a wrong one. The daemon enforces the rule regardless.

The existing fail-open rules apply per alternative — one whose field nothing can answer for satisfies the whole requirement, on the same reasoning as an unresolvable single field.

The repaint loop now follows every field a condition consults, not just `req.field`. With two controlling fields, watching one would leave the warning correct on mount and stale under exactly the edit that clears it.

### Tests

Nine new cases in `tests/requires.test.js` covering the one refused combination, the three permitted ones, `when` still scoping to the field's own value, an unresolvable alternative, an empty list, and a requirement carrying neither shape. Full suite green.

The camera-side change that emits this condition is already released; a camera running it serves the fragment above from `/api/v1/config.schema.json`.

### DIFF
--- a/tests/requires.test.js
+++ b/tests/requires.test.js
@@ -223,4 +223,68 @@ group('renderField reads the widget hatches on call, not at declaration');
 	}
 }
 
+// A rule that only bites when two settings coincide.
+//
+// majestic refuses live HLS on a camera whose recorder is in motion mode,
+// because nothing is written between detections and the playlist has no new
+// bytes to describe. It does NOT refuse it when the recorder is switched off:
+// records.mode then describes nothing that is happening, and HLS serves out of
+// memory exactly as it does on a camera with no card.
+//
+// One controlling field cannot say that. Written as "HLS requires
+// records.mode continuous" the form would grey the switch out on a camera with
+// recording off and tell its owner to change a setting that is inert — a
+// warning about a configuration the camera deliberately allows, which nothing
+// on the page can clear. So the condition arrives as alternatives, satisfied
+// by either way out of the rule.
+group('a requirement satisfied by any one of several alternatives');
+{
+	// Exactly what majestic emits, including the absent top-level `field`.
+	const HLS = {
+		when: 'true',
+		any: [
+			{ field: 'records.enabled', equals: 'false' },
+			{ field: 'records.mode', notEquals: 'motion' },
+		],
+		message: 'Motion recording writes nothing between detections.',
+	};
+
+	const lookup = (hls, enabled, mode) => ({
+		self: () => hls,
+		saved: (dot) => dot === 'records.enabled' ? enabled
+			: dot === 'records.mode' ? mode : undefined,
+	});
+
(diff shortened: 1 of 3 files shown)


openipc-ai:
**Disclosure finding: correct, and fixed.** The description named a private repository and linked its tracker, which is the one gate here whose violation cannot be undone after merge. Removed from the description; the commit messages and the diff were already clean, and I checked the sibling documentation pull request for the same leak and found none. The sentence it replaced now says the same thing from outside the camera: the condition is served from `/api/v1/config.schema.json`.

**Test-scope finding: declining, with reasons.** The rule is that every file under `tests/` earns its place by covering a subject that fails **silently** and cannot be reproduced on demand. These are cases added to a file that already earns its place on exactly that basis — its subject is this evaluator, and the failure it guards is a warning that does not draw.

The new shape fails the same way, in both directions and stating nothing either way:

- get it wrong one way and no warning draws on a camera that needs one — a form with no warning on it looks exactly like a camera doing what it was told;
- get it wrong the other and a warning sits permanently on a valid camera, with nothing on the page able to clear it.

Neither can be produced to order: seeing the first needs a camera recording on motion with HLS on, and seeing the second needs that same camera with recording switched off, to watch the warning go away again.

The truth table is what makes those two failures visible from a checkout, so removing it would take away the only place the distinction is written down. What was genuinely missing is that the file did not *say* any of this next to the new cases — a reader arriving at four permitted combinations had no way to see which silent fault they pin. That is now in 440e405, and it is the part of this finding worth acting on.

PR #349, opened 2026-09-05
Settings: an emptied number field means "not set", and the pin map was two pads short

openipc-ai:
> *"I can't clear these thresholds because after saving, empty fields are automatically set back to `0`. This should be fixed, preferably for all settings fields in all categories where an empty value is a valid state."*

Correct, and worse than it looks: #343 added a banner telling this reporter to clear the day and night thresholds to hand day/night back to automatic mode — and the page could not do it. It had trapped them in the state it was telling them to leave.

### Measured

On an hi3516ev300 + imx335, driving the config API directly:

| what is sent for `nightMode.minThreshold` | what lands in the config | `night_mode_source` |
|---|---|---|
| `""` (what the page sent) | `minThreshold: 0` | **2** — still the threshold mechanism |
| `null` | key removed | **4** — automatic, `night_auto_*` gauges return |

A number field has no way to say "empty" other than by being empty. An empty string reaches the config walker and an integer key stores **zero**, which is a real value and a different statement. Only `null` removes a key, and until now only the four IR-cut pins sent it.

### The rule

An emptied **numeric** control sends `null` whenever the schema declares **no `default`** for it.

- A key *with* a default has a defined unconfigured state that is not absence — it is re-seeded on every load — so removing it would read as absent now and as the default after the next restart. That is what the per-field reset control is for.
- **Strings are deliberately left alone.** An empty string is a value someone can mean; an overlay line cleared to nothing is not the same request as an overlay key that does not exist.

### What it covers

Fourteen keys, and every one is a setting whose empty state is real:

`isp.drc`, `nightMode.overrideDrc`, `video0.refEnhance`, `usbcam.bitrate`, `netip.port`, `nightMode.minThreshold`, `nightMode.maxThreshold`, `nightMode.autoNightGain`, and six GPIO pads — `nightMode.irCutPin1`, `irCutPin2`, `backlightPin`, `lightSensorPin`, plus **`audio.speakerPin` and `sip.buttonPin`**.

Those last two are worth calling out. The pads are the reason this pattern exists at all — 0 is a real pin, listed as RESET on several XiongMai boards — and they had exactly the fault the IR-cut pins were fixed for, unnoticed, in two other sections of the same page.

### Verified end to end, not in the abstract

Driving the real form in a browser against a camera: emptying both threshold controls and pressing Save posts `{"nightMode":{"minThreshold":null,"maxThreshold":null}}`, both keys leave the config rather than becoming zero, and the camera returns to automatic mode with its countdown gauges back on the same save.

No test is added. The predicate lives in `mj-settings.js`, which has no harness, and extracting a module for six lines would be the tail wagging the dog — the check that matters is the one above, which exercises the form, the batch shape, the daemon and the resulting camera state together.

Suite green (25 files).

Refs #325 — the reporter's two other points on that issue (the page has become text-heavy, and they would rather see only the active set of parameters) are not addressed here and are answered on the issue.

---

## Second commit: the pin map knew about four pads, and there are six

Enumerating the fields above turned up `audio.speakerPin` and `sip.buttonPin` as GPIO pads with the same clearing bug. Which raised a better question, and this is the answer to it.

The Day / Night pin map draws every pad on the SoC and dresses the ones already spoken for, so somebody assigning an IR-cut coil can see what is free. It read the four Day / Night pins and the PTZ lines from U-Boot — and **nothing else majestic's own config assigns**. The speaker's pad and the doorbell button's pad were drawn as free, on a page whose entire subject is which pad does what.

The second consequence is the one that matters: `assigned` is also the sweep's refusal list, so **"Find them for me" treated both as candidates** — raising one against another and braking both. On a speaker enable or a button line that is not a measurement anybody asked for.

Verified on an hi3516ev300 with both pins set:

| request | before | after |
|---|---|---|
| drive the speaker's pad | `done: true` — pad driven | `pad 52 is already used by the speaker` |
| drive the doorbell's pad | `done: true` — pad driven | `pad 53 is already used by the doorbell button` |
| drive a genuinely free pair | `done: true` | `done: true` |

The refusal **names what the pad is used by** rather than saying "already assigned" — leaving the reader to go and find out what to is the one thing not to make them do here — and the names are the ones on screen: the night illuminator, the daylight sensor, the speaker, the doorbell button. Never the config keys behind them.

No new UI: `mj-pin-owned` already dims the pad, disables it, and names its owner in the tooltip and in the pad's popover. These two were simply not in the list.

### DIFF
--- a/www/a/ircut-map.js
+++ b/www/a/ircut-map.js
@@ -113,9 +113,20 @@
 		// OpenIPC is majestic's, and majestic never releases a pad when its
 		// nightMode key is deleted; treating that as ownership would lock a
 		// camera out of the pad a rescan needs most.
+		//
+		// The roles here are the ones majestic and U-Boot assign OUTSIDE this
+		// section. The four Day / Night roles are excluded because they are
+		// drawn as roles, in their own colours; these have no row of their own
+		// on this page, so without a word here their pads would read as free
+		// to anyone picking one.
+		const FOREIGN = {
+			ptz: 'used by the PTZ driver',
+			speaker: 'used by the speaker',
+			button: 'used by the doorbell button',
+		};
 		const owned = {};
 		(info.assigned || []).forEach((a) => {
-			if (a.role === 'ptz') owned[a.pin] = 'used by the PTZ driver';
+			if (FOREIGN[a.role]) owned[a.pin] = FOREIGN[a.role];
 		});
 		(info.held || []).forEach((h) => {
 			if (h.owner && h.owner !== 'sysfs') owned[h.pin] = 'held by ' + h.owner;
--- a/www/a/mj-settings.js
+++ b/www/a/mj-settings.js
@@ -3787,12 +3787,34 @@
 		return r ? r.label : '';
 	}
 
+	// What this field is called on screen. A Day / Night pin has a role on the
+	// map — "IR-cut filter, closing coil" — which is what every sentence in
+	// that panel calls it; anything else is named by the caption printed above
+	// its own control, derived exactly as renderField derives it. Never the
+	// dotted key: a name readable only by someone who already knows the answer
+	// is not a name.
+	function fieldName(f) {
+		const key = f.dot.slice(f.dot.lastIndexOf('.') + 1);
+		const role = f.dot.indexOf('nightMode.') === 0 ? roleName(key) : '';
+		return role || (f.schema && f.schema.title) || titleCase(key);
+	}
+
+	// Which of the fields this save asked the camera to REMOVE are still
+	// configured afterwards. Asked of the refreshed config, so it is the
+	// camera answering and not the form: an older majestic accepts a null,
+	// answers 202 and ignores it, and nothing else on the page would ever say
+	// that a clear did not take.
+	//
+	// It used to look only under nightMode and only at pins, which was right
+	// while pins were the only thing that could be cleared. Once any numeric
+	// field with no default could be, that shape quietly stopped covering the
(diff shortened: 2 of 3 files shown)


openipc-ai:
Addressed in fafbc00. The finding is right, and it is the more serious half of this PR — thank you for it.

`stillSet` is the guard that catches a clear the camera did not honour: an older majestic accepts a null, answers 202 and ignores it, and nothing else on the page would ever say so. It looked only under `nightMode` and only at fields carrying an IR-cut role, which was exactly right while those four pins were the only thing that could be cleared.

Widening `clearsToNull` quietly took that cover away from the cases it was written for. A threshold whose removal was ignored — **the very clear this PR exists to make possible** — or a speaker pad, would have been reported as a clean save. That is worse than the gap it replaced, because #343 now actively tells people to clear those fields.

It asks the refreshed config about each cleared field's own dotted path now, and names what it finds the way the page names it: the map's role for a Day / Night pin, otherwise the caption printed above the control, derived exactly as `renderField` derives it. Never the dotted key. The message loses its pin-specific wording with it — *"could not be disconnected … configured to drive them"* is about coils, and a threshold is not driven.

Verified end to end on an hi3516ev300: clearing both thresholds through the real form still posts two nulls, both keys leave the config, the camera returns to automatic mode with its countdown gauges back, and no error is raised. The branch this fixes needs a firmware that ignores nulls, which I cannot produce here — what I checked is that the honest path stays quiet, since a guard that cries wolf on every good save would be its own bug.

Issue #352, opened 2026-09-05
Play back encrypted recordings (CENC), and a password field for records.key


widgetii:
Majestic can now write encrypted recordings to the card (OpenIPC/majestic 781c635cf). The WebUI has no way to play one back, and no way to hand over a key — this issue is that gap.

## What the camera writes

Ordinary fragmented MP4, `.mp4`, under Common Encryption (ISO/IEC 23001-7 `cenc`, AES-128-CTR):

- video sample entries are `encv` (audio `enca`) with the real codec inside a `sinf`, so a player that ignores encryption still sees the track;
- each fragment carries `senc`/`saiz`/`saio` — one 8-byte IV per sample, and for video the subsample layout: the 4-byte NAL length and the NAL header stay clear, the rest is ciphertext;
- a `pssh` at the end of `moov` under majestic's SystemID `9d2c417a-6e38-4bd1-a50f-3c8e217bd462` holds the clip's key box;
- a `uuid` box in each `moof` holds the integrity chain's tag (HMAC-SHA256 over the previous tag, this `moof` and its `mdat`).

Nothing else is added, so the byte-range index the WebUI already builds still walks the file.

`ffmpeg -decryption_key <hex> -i clip.mp4` plays it today, and a browser can too through EME ClearKey — the format was chosen for exactly that.

## What is missing

1. **A player.** MSE + EME ClearKey over the existing fragment index, or a decrypt-on-read fallback. The AES key is 16 bytes; the KID is in the key box.
2. **Where the key comes from**, which differs per `records.encryption`:
   - `passphrase` — the camera can derive it. Needs an authenticated endpoint that unwraps the clip's passphrase slot and returns the key for that clip (PBKDF2-HMAC-SHA256, 100 000 rounds, cached on the camera), or a WebCrypto implementation in the browser given the passphrase.
   - `chip` — the key unwraps only inside the SoC's cipher engine, so the browser can never have it. Either a decrypt-on-read route on the camera, or accept that these clips are not playable in the WebUI.
   - `pubkey` — the camera has no private key by design. Not playable on the camera, ever. The UI should say so rather than fail.
3. **A password field for `records.key`** in the settings UI. The schema now marks it (`x-secret` / `writeOnly`); the UI still renders it as plain text.
4. **Telling the operator what a clip is.** The recordings list should show which mode a clip carries and whether this browser can open it, instead of offering a play button that cannot work.

## Notes for whoever picks this up

[...]

PR #360, opened 2026-09-05
Encrypted recordings: the crypto, the format, and the key this browser holds

widgetii:
Toward OpenIPC/majestic-webui#352 — playing back encrypted recordings in the browser.

**Draft on purpose.** This is the machinery, tested; the Recordings page does not use it yet. What is here is worth reviewing on its own terms, and one commit is a fix for a bug that is in the tree today.

## What a sealed clip is

The camera can encrypt what it writes to the card. A sealed clip is ordinary fragmented MP4 under Common Encryption: the real codec hidden behind an `encv` sample entry, AES-128-CTR over the samples, per-fragment boxes saying which bytes are protected under which initialisation vector, and the clip's own keys wrapped into a `pssh` in its header — under a passphrase, under the camera's own silicon, or under the owner's public key. Handed to a browser untouched none of that errors: MediaSource takes the init segment, takes the fragments, and shows nothing at all.

## The constraint that shapes it

`crypto.subtle` is secure-context only and a camera is `http://` on a LAN, so it does not exist on the page this feature lives on — the same wall `preview-wasm.js` records for WebCodecs, and the microphone and clipboard for theirs. So the crypto is written out in this tree. `crypto.getRandomValues` is not gated that way and is where every salt comes from.

EME was considered and is not the answer either: it needs a secure context too, and OpenIPC's own [rnd-player](https://github.com/openipc/rnd-player) supports ClearKey *and* still carries a full software AES-128-CTR path, because a browser can resolve `requestMediaKeySystemAccess('org.w3.clearkey')` and then silently decrypt to garbage. Its senc parser, its init rewrite and its child-offset table were read while writing `mp4crypt.js`; this is a vanilla-JS reimplementation, not a copy, since this tree takes no dependencies.

## What is in the diff

- **`www/a/mjcrypto.js`** — SHA-256, HMAC, HKDF, PBKDF2, AES-128-CTR, RSA-OAEP decryption, and enough DER to read both private-key PEM forms openssl writes and to produce the public half byte-for-byte as openssl would. Encryption direction only: nothing here encrypts, and a decryption table nobody calls is a thing to maintain and never notice breaking.
- **`www/a/mp4crypt.js`** — the key box and its slots, the header rewrite (`encv` → the codec it hides, `sinf` and `pssh` dropped, every ancestor size recomputed), per-fragment decryption, and the integrity chain. Every failure is a refusal, never a repair: decrypting *most* of a fragment produces video that plays and is not what the camera recorded.
- **`www/a/reckeys.js`** — the private key for public-key mode, wrapped under an unlock passphrase in this origin's IndexedDB and unlocked once per visit. A convenience, not a backup; the PEM taken at generation is the copy that matters. Generation needs the browser's own generator and so a secure context — everything else works on a camera as it ships, and where it cannot generate the panel will print openssl's two commands.
- **`www/a/mp4index.js`** — a fix, and the reason it is in this branch: `parseFragment` reports a read that was too small and every caller widens on it except `durationHint`, which returned null. An encrypted fragment header is near two kilobytes against the one kilobyte the walk asks for, so `durationHint` gave up on every sealed clip — silently. The list keeps the estimate it started with, the timeline states a length nobody measured, and the correction that stops the playhead running past the end never runs.

## The tests are the point

The fixtures are built with node's own crypto, never with the module under test, so the format is written out twice independently. That is what made three bugs findable, and all three produced correct-looking output:

- the CTR counter read the value of `++c[i]`, which on a typed array is 256 while the byte it stores is 0 — so the carry never propagated. Right first block, wrong keystream from the second onward, for any counter ending `0xff`;
- one S-box entry wrong, because the inverse lookup asked for `g^255` at the one index the table never fills. The FIPS single-block vector passes straight through it; a 4 KiB stream against an unrelated implementation does not. In a frame that is a smear on one macroblock;
- `tenc` read with one reserved byte where the format has two, which finds the IV size where the protection flag is.

`npm test` runs the three new suites plus a case added to `mp4index`'s.

## Still to come on this branch

The page wiring: detecting a sealed clip when it opens, decrypting each fragment before it reaches MediaSource, the keys panel on the Recordings page, decrypted export, and the masked field the settings form still lacks for a secret the schema marks. Verification on a camera with a card follows that, over plain http — the case that matters — and the numbers go in a comment here.


widgetii:
All four were real, and two of them were the kind this branch exists to be careful about — code that answers confidently having never looked. Fixed in 1fe2f9c.

**`inspect()` guessed clear media.** Correct, and the worst of the four. An absent `moov` is a short read, a file that is not an MP4, or a clip whose header is wider than the window the caller took — none of which is "this recording is unencrypted", which is what it said. There is a third answer now (`ok: false` with a reason), and it is neither a reason to play nor a reason to ask for a passphrase. Tested with a truncated header and with bytes that are not an MP4 at all.

**A truncated `trun` was accepted.** Also correct, and the failure is exactly the silent kind: past the end of a `Uint8Array` every byte is `undefined`, every bitwise read turns that into zero, so a table listing a thousand samples while carrying ten walks nine hundred and ninety samples of no length, completes, and hands back a fragment with nothing decrypted in it. Every optional field is now bounded against its own box, and the whole sample table is required to fit rather than just its first row. Two tests: a sample count that runs past the box, and a `tfhd` claiming a field it does not carry.

**Stored iterations could hang the page.** Correct. The record states its own unlock cost, so the record decided how long the tab stopped responding, and PBKDF2 cannot be interrupted once it starts. The shape is validated before any work — finite whole number within a bound, byte strings of the right sizes — and the bound is deliberately a *maximum* rather than today's figure, so a key wrapped by a later build under a costlier setting still opens. That last part is the reason not to simply require the current constant, and it is pinned by a test.

**A timed-out database stayed open.** Correct. A late `onsuccess` now closes the connection it was handed, and the timer is cleared when the request settles first.

One note on where the tests went. The first two have regression tests because they produce plausible output with no error anywhere; the second two fail visibly, and got tests anyway for a reason worth stating — from a page that is waiting, "returns null" and "never returns" look identical, so the bound has to be asserted rather than assumed.

PR #364, opened 2026-09-05
Recordings: play a sealed clip, and hold the key that opens it

### LINKED ISSUE #352: Play back encrypted recordings (CENC), and a password field for records.key

widgetii:
Closes OpenIPC/majestic-webui#352, on top of #360's machinery.

The Recordings page could list an encrypted recording and do nothing else with it. Its sample entries name a codec no decoder implements, so the probe refused it as unplayable — and the advice on screen, *save it and open it in VLC*, was false: the file that lands on disk opens in nothing either.

## What the page does now

- **Plays it.** The header is read, rewritten into one a decoder accepts (`encv` → the codec it hides, protection boxes dropped, sizes recomputed), and every fragment is decrypted on its way to the SourceBuffer. The rewrite happens **before any key is asked for**, which is what lets a browser that cannot decode the codec still offer a file that plays elsewhere — the codec and its configuration sit in the clear beside the protection boxes.
- **Asks for the key once.** One camera writes one key across a day, so the first clip's prompt covers every clip and seek after it. The prompt lives inside the stage, over the picture, and deliberately not in the note bar above it: that bar is wiped by the next seek and by the card poll, and a question about a passphrase has to outlive both.
- **Says which kind of "no" it is.** A chip-bound clip can only be opened by the camera that recorded it — no browser, on any machine, ever — and says so with what to change so the next recording can be opened here. A clip sealed to a different key names both fingerprints. A wrong passphrase says the passphrase is wrong, not that something failed.
- **Two save buttons that mean different things, said on the buttons.** The whole clip goes as it is on the card — still sealed, still carrying its integrity record, streamed by the browser so a recording of any length is safe to take. A selected cut is decrypted on the way out, because the point of saving a cut is that it plays somewhere.
- **Checks integrity on demand.** Forward from the start of the clip, because each fragment's tag covers the one before it — a chain restarted mid-file proves nothing about what came before, which is also why this is a button and not something the player does as it seeks. A clip still being written ends mid-fragment, and that is the camera doing its job: it says so rather than reporting a truncation.
- **Manages the key.** A card on the same page: generate (where the browser will make keys — a secure context, which a plain-http camera is not), or the two `openssl` commands and a file picker where it will not; unlock once per visit; download the backup; publish the public half to the camera. The setting is written **only when the camera's own schema declares it** — writing a key into a configuration the daemon ignores and reporting success is exactly the confident-sentence-with-nothing-behind-it this repo's checklist is about.
- **Masks a secret in the settings form.** The schema marks one; the form rendered it in the clear. The reveal toggle is wired to the field itself, because the one in `main.js` runs once at page load and this form is built long afterwards.

## Three paths that used to end at a bare `<video>`

The five-second watchdog, the fallback for recordings written before `tfdt`, and the element's own error handler. Each would hand a sealed file to a decoder that cannot read it — a black frame, and on some builds a prompt for a decryption module nobody here can answer. All three are gated on whether the clip is sealed and say what happened instead.

## Tests

One new case, and it is about the **wiring** rather than the format, which has its own suite from #360. A transform applied to the fragments but not the header gives the decoder an initialisation segment naming a codec nothing implements; one applied to the header but not the fragments gives it ciphertext. Neither raises anything, both are a black frame and a page that looks like it is still loading. The stub marks every buffer, so an append that skipped a transform is visible rather than merely wrong-looking.

`mp4crypt` gained two cases from a narrowing this branch forced: a `moov` whose tracks the walk cannot take apart is not the same as a header nobody saw. The bytes are there, so the question is asked directly — no protection box anywhere inside them is a *finding*, and refusing such a clip would refuse a recording that plays today.

`npm test` green, `lint-templates` green, and the purged Bootstrap subset is unchanged (every class the panel uses was already in it).

## Not verified on a camera yet

This is the part I cannot do from here. What needs a card, over plain http, in a browser where `crypto.subtle` is undefined: a passphrase clip plays and seeks and exports; a `pubkey` clip plays with an imported key and asks again after a reload; a chip clip says why it cannot be opened; a flipped byte on the card is named by fragment; and a plain clip still behaves exactly as it does today. The decrypt timings go in a comment when they exist.

### DIFF
--- a/tests/mp4crypt.test.js
+++ b/tests/mp4crypt.test.js
@@ -434,6 +434,27 @@ group('reading the header');
 	check('and so is something that is not an MP4 at all',
 		notMp4.ok === false && !!notMp4.reason);
 
+	// A moov whose tracks this walk cannot take apart — a header shaped in a
+	// way this build has not seen — is a different case again. The bytes are
+	// here, so the question can be asked directly rather than assumed either
+	// way: no protection box anywhere inside them is a finding, and refusing
+	// such a clip would refuse a recording that plays today.
+	const bare = Buffer.concat([
+		box('ftyp', Buffer.alloc(16)),
+		box('moov', box('trak', box('mdia', box('mdhd', Buffer.alloc(24))))),
+	]);
+	const b = CRYPT.inspect(u8(bare), bare.length);
+	check('a header this walk cannot take a track out of, holding no protection at all, is clear',
+		b.ok === true && b.encrypted === false && b.reason === null);
+
+	const bareSealed = Buffer.concat([
+		box('ftyp', Buffer.alloc(16)),
+		box('moov', box('trak', box('mdia', box('minf', sinf('avc1', KID, {}))))),
+	]);
+	const bs = CRYPT.inspect(u8(bareSealed), bareSealed.length);
+	check('and the same header with protection in it is encrypted and unopenable, not clear',
+		bs.ok === true && bs.encrypted === true && /does not know/.test(bs.reason || ''));
+
 	// Signalled as protected and then declared not to be. It plays as it is,
 	// and demanding a passphrase for it would refuse a working clip.
 	const off = CRYPT.inspect(u8(initSegment({ isProtected: 0 })));
--- a/tests/mp4crypt.test.js
+++ b/tests/mp4crypt.test.js
@@ -447,6 +447,24 @@ group('reading the header');
 	check('a header this walk cannot take a track out of, holding no protection at all, is clear',
 		b.ok === true && b.encrypted === false && b.reason === null);
 
+	// The four letters are also ordinary bytes, and they turn up inside codec
+	// configuration by chance. A scan that matched one there would refuse a
+	// clear recording as a sealed one nothing can open — a worse failure than
+	// the one the fallback exists to prevent, so it reads box types at box
+	// boundaries and never looks inside a box it does not understand.
+	// A payload that happens to contain the letters, inside a box this walk has
+	// no business opening.
+	const gama = box('gama', Buffer.concat([
+		Buffer.from('xxsinf', 'ascii'), Buffer.from('tenc', 'ascii')]));
+	const stsdDecoy = box('stsd', Buffer.concat([Buffer.from([0, 0, 0, 0]), u32(1), gama]));
+	const decoy = Buffer.concat([
+		box('ftyp', Buffer.alloc(16)),
+		box('moov', box('trak', box('mdia', box('minf', box('stbl', stsdDecoy))))),
+	]);
+	const d = CRYPT.inspect(u8(decoy), decoy.length);
+	check('four letters inside a payload are not a protection box',
+		d.ok === true && d.encrypted === false && d.reason === null);
+
 	const bareSealed = Buffer.concat([
 		box('ftyp', Buffer.alloc(16)),
 		box('moov', box('trak', box('mdia', box('minf', sinf('avc1', KID, {}))))),
--- a/tests/recordings-health.test.js
+++ b/tests/recordings-health.test.js
@@ -176,7 +176,10 @@ function load(cardHealth, metrics, mode) {
 	};
 	ctx.window.document = ctx.document;
 	vm.createContext(ctx);
-	for (const f of ['timeline.js', 'mp4index.js', 'recordings.js']) {
+	// The page loads these in this order, and so does the browser: recordings.js
+	// takes the crypto modules at module scope, the way it takes the index.
+	for (const f of ['timeline.js', 'mp4index.js', 'mjcrypto.js', 'mp4crypt.js',
+		'reckeys.js', 'recordings.js']) {
 		vm.runInContext(fs.readFileSync(A(f), 'utf8'), ctx);
 	}
 	return env;
--- a/tests/recordings-pump.test.js
+++ b/tests/recordings-pump.test.js
@@ -155,7 +155,7 @@ function makeEl(id) {
 const CLIP_SECONDS = 120;
 const CLIP = clip(CLIP_SECONDS);
 
-function load() {
+function load(opts) {
 	const env = { appends: [], ranges: makeRanges(), reads: [], eos: false };
 	const els = {};
 	const $ = (id) => (els[id] = els[id] || makeEl(id));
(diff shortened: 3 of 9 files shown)


widgetii:
Four of the five were right, and one of them found something I had convinced myself I had already done. Fixed in 157e390.

**Integrity key material discarded, and stale material carried forward.** Correct on both halves, and the second is the serious one. The material was kept on only one of the two paths that acquire it — I had made that edit in a worktree I later deleted and never noticed it was missing — and nothing cleared it when a clip closed. So verification was unavailable exactly where a sealed clip played normally, and a button left enabled from a previous clip would check the new one against the old key and report *this recording does not match its own integrity record*. That is an accusation about somebody's footage, produced by a bug, and it is the worst sentence this page can emit. Material and controls are per clip now: cleared as one opens, set on every unlock path.

**Short truncation passed integrity.** Correct. A tail of one to fifteen bytes was too short to parse and read as a clean end of file — and a clip cut a few bytes into its next fragment is precisely what the button exists to notice. "Complete" now means the walk finished on the last byte.

**Failed persistence reported success.** Correct. Storage can refuse outright — private window, blocked site data, a transaction that failed — and the store answers null when it did. The key is now held in memory either way so the visit works, and the sentence says which of the two happened rather than claiming a key is on the machine when it is not.

**Schema failure became a firmware verdict.** Correct, and it is this repo's third gate almost word for word: a 500 or an unreadable body is a fact about one request, while "this build does not offer the setting" is a fact about the camera. Separate sentences now, and only the second claims anything about the firmware.

**Recording list indicators — handled differently.** Encryption is per clip and the listing endpoint does not carry it, so knowing a given row's answer means reading that file's header: one range request per row, on a page whose entire index strategy is built on the cost of an answer being the number of requests rather than the bytes. A day of clips would pay for a badge. What the page can say honestly and for free, it now says: what the camera is doing to **new** recordings, from the configuration it already fetched, with the explicit note that older clips are however they were written at the time — plus a lock on the row whose header has actually been read. A badge on every row would be a guess wearing the clothes of a fact, which is the failure mode the checklist's third gate is about.

widgetii:
Six more, and all six were right. Fixed in d29eb9c.

**Published key metadata unauthenticated.** The record's tag covers the wrapped key and the parameters needed to unwrap it — not the public half stored beside them. So an edited record could pair a genuine private key with somebody else's public one, and the camera would seal every future recording to a key the operator does not hold. The public half is now derived from the unlocked private key, which also turns publishing into something that proves possession rather than asserting it; publishing while locked asks for the unlock first, and a stored fingerprint that disagrees with the key is reported rather than used.

**Corrupt tails reported as recording.** Two different tails end a walk, and only one of them is a camera doing its job: a fragment whose header is on the card and whose payload is not, versus a run of bytes that is not a fragment at all. Telling somebody the second is the first says their damaged recording is fine. Separate verdicts now, with the byte offset where it stops making sense.

**Verification failures stayed silent.** Worse than it sounds — the button came back and no verdict appeared, which reads as *checked, and fine*. A read that failed now says so, and says that nothing is known about the rest either way.

**Empty playback claimed prior success.** A header that could not be prepared and a SourceBuffer that never accepted anything have decoded no frames, so "everything before that point played correctly" was a claim about a picture nobody saw. Three phases, three sentences, and only the fragment case says anything about what played.

**Imported keys allowed empty passphrases.** Same guard generation already had. A key stored that way is readable by anyone who reaches the browser profile, and an imported key is exactly as worth protecting as one made here.

**Payload text mimicking protection.** Right, and it is the more dangerous direction of the two: those four-character sequences are ordinary bytes that turn up inside codec configuration by chance, so a clear recording could be refused as a sealed one nothing can open — worse than the case the fallback exists to prevent. It walks box boundaries now, descends only containers it knows, and never looks inside a box it does not understand. The test hides `sinf` and `tenc` in a payload where the scan would have found them.

**Recording list indicators — still handled differently.** Unchanged from my last reply: encryption is per clip, the listing endpoint does not carry it, and the answer for a given row costs a range request against that file's header — on a page whose entire index strategy exists because the price of an answer is the number of requests. The page now states what it can know for free (what the camera is doing to new recordings, from configuration it already fetched, with older clips explicitly not covered by that claim) and marks the row whose header it has actually read. A badge on every row would be a guess wearing the clothes of a fact.

PR #366, opened 2026-09-05
Recordings: take the protection boxes out with the protection

widgetii:
Found by running the merged code against a camera, which is the only place it could have been found.

## What happened

A fragment came back with its samples decrypted and its `senc`, `saiz` and `saio` still in it. The reasoning was that a track whose sample entry no longer names a protection scheme has no use for boxes describing one — and ffmpeg agrees, reading those files without a word.

Chromium does not. Its parser finds sample-auxiliary information on a track with no scheme and refuses the fragment: `appendBuffer` raises an error event, the SourceBuffer takes nothing, and the page shows the spinner it shows for a camera that has gone quiet.

This is the shape of failure the whole feature is written against, and nothing in the suite could have caught it: the bytes were correct, every test passed, ffmpeg decoded them, and the only parser that disagreed is the one that matters here.

## The fix, and the three numbers

The boxes come out, which moves the payload:

- `trun.data_offset` counts from the front of the moof, so it shrinks by exactly what was removed — **in every traf**, because the mdat is shared and a clear track beside a protected one loses nothing yet still moves;
- each `traf` shrinks by what came out of it;
- the `moof` shrinks by the total.

Four bytes wrong about any of those lands the decoder inside a picture instead of at the start of one, so the tests pin all three: what came out, what stayed, that every container is as long as what it holds, that both tracks' offsets moved by exactly the bytes removed, and that the payload at each new offset is byte-for-byte what the camera recorded.

A moof that cannot be rebuilt is refused rather than returned. Decryption happens first, so handing back what arrived would mean decrypted samples still wearing their protection boxes — the state this change exists to prevent, reached silently.

The integrity `uuid` stays. It describes the recording rather than the protection, a parser has no opinion about a `uuid` box it does not know, and a decrypted export that still carries the camera's own tags is worth more than one that does not — the tags no longer match the bytes beside them, which the page already says on the button.

## Verified against hardware

On an hi3516av300 with an SD card, over plain http, with clips the camera recorded in each mode — a 4K H.265 main stream and a 1080p H.264 substream, both sealed with a passphrase, one sealed to a public key, one bound to the camera's own hardware and one in the clear.

The shipped page modules, loaded the way the browser loads them, opened every clip the camera's own tooling can open and refused the ones it should: the chip-bound clip is recognised as openable by that camera alone, with both a passphrase and a recovery key declined by name. Each sealed clip's integrity chain matched across every one of its fragments, a clip with a single byte altered was reported at the fragment where it stops matching with everything before it still intact, and the decrypted output decodes without a complaint from ffmpeg while the sealed file produces over a thousand decoder errors.

In a real browser — headless Chromium, driving the shipped modules against a sealed H.264 recording — the SourceBuffer accepts the rewritten header, takes six decrypted fragments and decodes 1920x1080 video with three seconds buffered. **Before this commit that run reached the header and failed on the first fragment.** The same browser reports no support for `hvc1`, so a 4K H.265 recording takes the page's "this browser cannot decode it" path and is offered as a decrypted file to save instead — which is why the header rewrite happens before any key is asked for.

## What it costs

Measured on the same hardware class, on a 15.6 Mbit/s recording: unlocking a clip costs about half a second of PBKDF2, paid once per passphrase rather than per clip; rewriting the header is under a millisecond; decryption runs at roughly 60 MB/s and the integrity walk at 90 MB/s, so playback spends around 3% of one core and a full-clip integrity check is bounded by how fast the card can be read rather than by the cryptography.

### DIFF
--- a/tests/mp4crypt.test.js
+++ b/tests/mp4crypt.test.js
@@ -547,29 +547,82 @@ group('decrypting a fragment');
 (function roundTrip() {
 	const p = CRYPT.inspect(u8(init), init.length);
 	const f = fragment();
-	const got = CRYPT.decryptFragment(u8(f.bytes), MATERIAL, p.tracks);
-	check('a fragment decrypts to exactly what was recorded',
-		hex(got) === hex(f.plainBytes));
+	const got = Buffer.from(CRYPT.decryptFragment(u8(f.bytes), MATERIAL, p.tracks));
 
-	// The trap. A multi-NAL sample's protected runs are one keystream; the
-	// clear five bytes between them do not advance it, and the runs are not
-	// multiples of the block size. Decrypting each run from its own counter
-	// gets the first one right — a picture, with the rest of the frame wrong.
-	const first = f.video ? 0 : 0;
-	check('including every NAL of a multi-NAL frame, not just the first',
-		hex(Buffer.from(got).subarray(f.payloadAt, f.payloadAt + 600)) ===
+	// The output is not the input with its samples swapped: the boxes that
+	// described the protection come out with it, because a browser reads them,
+	// finds them on a track that is no longer protected, and refuses the whole
+	// fragment. So the payload moves, and three numbers move with it.
+	check('the boxes describing the protection are gone',
+		got.indexOf(Buffer.from('senc', 'ascii')) < 0 &&
+		got.indexOf(Buffer.from('saiz', 'ascii')) < 0 &&
+		got.indexOf(Buffer.from('saio', 'ascii')) < 0);
+	check('and the integrity tag is not — it describes the recording, not the protection',
+		got.indexOf(INTEGRITY_UUID) > 0);
+
+	const moofLen = got.readUInt32BE(0);
+	check('the moof says how long it now is',
+		got.toString('ascii', 4, 8) === 'moof' && moofLen < f.moofLen);
+	check('and the mdat behind it is untouched in length',
+		got.toString('ascii', moofLen + 4, moofLen + 8) === 'mdat' &&
+		got.readUInt32BE(moofLen) === f.bytes.readUInt32BE(f.moofLen));
+
+	// Every container's size still equals what it holds, walked by a checker
+	// that shares no code with the rewriter.
+	const bad = [];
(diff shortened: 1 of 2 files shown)


widgetii:
All three were right, including the one about me.

**Lab output in the description.** Correct, and I had no excuse: the rule says verification is described by hardware class and configuration, not pasted from the device, and I pasted two blocks of harness output. The description now says what was verified, on what class of hardware, in prose. Rewritten rather than edited around, since an edit leaves the original in the history — which is the reason that gate exists.

**Clear-track offsets stayed stale.** Real, and it is the more dangerous of the two. The mdat is shared, so it moves earlier by everything removed from the *whole* moof — but only the trafs that lost boxes were being corrected. A clear track beside a protected one loses nothing and still moves, so its samples would have been read from wherever the protected track's used to start and decoded as pictures made of another track's bytes. Every traf is rewritten now, and the test puts a clear track beside a protected one and checks both offsets and both tracks' bytes.

**Malformed fragments remained protected.** Also real. Decryption runs before the boxes come out, so returning the original buffer meant handing back decrypted samples still wearing their protection boxes — precisely the state this change exists to prevent, reached silently instead of by refusal. It throws now, and the test bends a box that decryption itself never has to look at, so the failure only becomes reachable at the rebuild.

The browser check was re-run after both fixes: the SourceBuffer still accepts the rewritten header and decodes 1920x1080 from the decrypted fragments.

widgetii:
On the disclosure finding, which is the one still standing: the description was rewritten before the re-review and now carries no output blocks at all — no code fences, no console lines, no addresses, no device names beyond the SoC class the rule explicitly asks for. What it says is what was verified, on what hardware class, in prose.

What cannot be undone is the edit history, and the checklist says so in its own preamble: editing a merged pull request's description leaves the original visible. So the violation stands as a fact about this pull request even though the published description is clean, and the right response is to say that plainly rather than to keep editing.

The content that was in it was harness output — decoder complaints, buffered seconds, a resolution — with no hostname, address or credential in it. That is not a defence of having pasted it; the rule is about pasted device output as a category, and it is a category I have now published once and will not again.

PR #372, opened 2026-09-06
Config comes from the daemon, and the deprecated reader leaves the tree

widgetii:
`/etc/majestic.yaml` holds only what differs from majestic's built-in defaults, so a key sitting at its default is **absent from it entirely**. Every shell here that parsed the file to learn a setting was reading back nothing for a value the camera was plainly running on — with nothing to distinguish that from a value nobody chose.

## It was not theoretical

On a lab camera with a **fully populated** config file, seven keys already disagreed between the file and the daemon:

| key | the file says | the camera says |
|---|---|---|
| `osd.privacyMasks` | *nothing* | `[ "760x380x420x300" ]` |
| `motionDetect.roi` | *nothing* | two regions |
| `records.encryption` | *nothing* | `none` |
| `records.chipSlot` | *nothing* | `1` |
| `hls.adaptive` | *nothing* | `false` |
| `video0.crop`, `video1.crop` | *nothing* | `[ ]` |

And `j/recordings.cgi` derived an empty prefix and answered `{"error":"no recordings directory"}` on a camera whose recording path had simply never been written to disk — the ordinary case, since the settings page saves only the keys somebody changed.

Proven on an hi3516ev300 + imx335 by clearing the key so only the default is in force: the file then contains no `path` at all, `mj_cfg` returns the default, and the endpoint derives a real prefix from it — where the previous code derived an empty one and reported no recordings directory.

## What changes

**Reads go to majestic.** `mj_cfg <dotted.key>` in the new `www/cgi-bin/p/majestic.sh`, over `GET /api/v1/get`. It exits non-zero rather than printing nothing, so a caller can tell *not set* from *set to empty* from *could not ask* — and `recordings.cgi` now says which of those happened instead of stating there is nowhere to look.

**`j/gpio.cgi` is a tombstone.** Its work moved into the majestic daemon, which already drove those pads on every day/night transition under no lock this script ever took. The file stays as a **410 rather than being deleted**: `sbin/updatewebui` prunes the overlay copy of a file a release stops shipping, which *uncovers the firmware's own older copy underneath* — and that copy is precisely the version this replaces. The page enumerates with `GET /api/v1/gpio` and **POSTs** to move a pad, because a GET is what a browser issues on its own and it carries the session with it.

**The map and the sweep read what the camera admits it cannot tell.** Without debugfs it cannot name a line's owner; without the boot environment it cannot see the PTZ pads. Either way the daemon refuses to drive, so the sweep is disabled with the reason stated up front rather than spent one press at a time, and a pad nobody could check is no longer printed as `free`. Picking a pin by hand still works — that writes a number into a field and moves nothing.

## The rule, enforced twice

`CLAUDE.md`'s conventions told you to use the deprecated tool, in the section an agent consults *before writing a config read*. Its "Majestic glue" list also documented two helpers (`get_night`, `get_metrics`) that had been deleted long enough ago to leave no trace in the tree. All eight mentions are gone; `git grep` is clean.

- `pr_compliance_checklist.yaml` — two gates: config comes from the daemon, and the browser asks majestic directly for anything majestic serves (`j/pulse.cgi` named as the model — it was cut back to overlay `df`, `/etc/timezone` and `date %z` once `/metrics` was found to carry the rest).
- `tools/lint-templates.sh` — **fails the build**. Review advises; CI refuses, which is what makes it hold after a reviewer's attention moves on. A flat string ban is defensible because the legitimate count is now zero.

## Also corrected

- the docs claimed `pulse.cgi` was polled every 2s and filled the whole top bar; it is asked every **15th** tick for the overlay bar alone, and `/metrics` carries the rest
- `recordings.cgi`'s header documented a `?prefix=` parameter that a security fix removed — twelve lines above the code that ignores it

## Verified

`npm test` and `tools/lint-templates.sh` pass. On an hi3516ev300 + imx335: every page renders without a template error; `mj_cfg` returns the default for a key the file does not contain, exit 1 for one the daemon says is unset, and exit 2 when the daemon is stopped outright; the recordings endpoint derives a real prefix and lists a day, and gives all three of those states different wording; the tombstone answers 410; a GET at a pad mutation is refused and a POST drives; and reintroducing either the deprecated tool or a direct read of the config file fails the lint.

**Requires a majestic build carrying `GET /api/v1/get` and `GET|POST /api/v1/gpio`.** Both have landed in the daemon; an older build 404s them, and every call site here degrades to a stated limitation rather than a wrong answer.


widgetii:
All six addressed in `afdf501`, and two of them were mine to have caught before pressing the button.

## 1 and 5 are the same defect: `mj_cfg` had two answers where it needed three

It returned non-zero for *"the key is not set"* and *"the daemon could not be asked"* alike. So the autofocus gate said **"Autofocus not available on this camera."** when nothing had established that — a claim about the hardware drawn from a request that failed — and the recordings page said **"could not ask the camera"** when the camera had answered perfectly well and the key was simply unset. Each was stating the other's case.

That is this project's own recurring bug, in a helper I added while removing an older instance of it. It returns `0` / `1` / `2` now, split on the status code rather than on curl's exit alone: the autofocus button is withdrawn only on a definite *no*, and kept when the camera cannot be reached so the attempt reports its own failure. `common.cgi`'s `af_support` had the same shape and got the same treatment.

Verified by stopping the daemon outright — `rc=2`, which was previously indistinguishable from `rc=1`.

## 4 — the pin map trusted any parseable JSON

Correct. `apiFetch` intercepts only 401, so an older firmware's JSON error mounted as pad data and its missing banks drew an empty chip instead of falling back to the plain number fields. Both call sites check `r.ok` now.

## 3 — my own disclosure slips

Both real, and in the change that strengthens that gate: `CLAUDE.md` named a daemon-internal function and `majestic.sh` named the daemon's internal auth predicate, each attributed to the firmware. Described by what they do now.

Related, and not caught by the bot: the original PR description referenced two pull requests in the private repository by number. Removed as soon as I noticed — the commits and the tree never carried them, but the description did, and that is as public as the code.

## 2 — the measurement had no hardware class

Fair. "One lab camera" is an anecdote about a machine nobody else has; `hi3516ev300` is a result someone can reproduce. The comment says so now, and the pasted device output is gone from the description in favour of describing what it showed.

## 6 — the lint banned the name, not the act

The compliance rule beside it also forbids reading the file directly, so a `grep` at `/etc/majestic.yaml` would have sailed past the check that exists to stop exactly that regression. It now catches an **access** — a reader or a redirect aimed at the path — rather than a mention, so `get_config` returning the path, `backup.cgi` listing it, and the dozen comments explaining what is in it all stay legal. Confirmed both ways: clean tree exits 0, a `grep` at the file exits 1.

widgetii:
On 6 — the direct-read check is already in `afdf501`, in the `if` immediately below the one the finding quotes. The snippet cited is the first of two.

```
$ git show afdf501:tools/lint-templates.sh | grep -c 'majestic\.yaml'
```

It deliberately catches an **access** rather than a mention, because the file is named legitimately all over the tree — `get_config` returns it, `backup.cgi` lists it as something to back up, and a dozen comments explain what is in it. Banning the name would fail the build on all of those.

Exercised against every form the compliance rule names, plus the two that must stay legal:

| line under test | result |
|---|---|
| `cat /etc/majestic.yaml` | **caught** |
| `grep foo /etc/majestic.yaml` | **caught** |
| `sed -n 1p /etc/majestic.yaml` | **caught** |
| `awk "{print}" /etc/majestic.yaml` | **caught** |
| `x=$(head -1 /etc/majestic.yaml)` | **caught** |
| `echo x > /etc/majestic.yaml` | **caught** |
| `read a < /etc/majestic.yaml` | **caught** |
| `# … /etc/majestic.yaml …` in a comment | allowed |
| `echo ${1}/etc/majestic.yaml` (the path helper) | allowed |

Clean tree exits 0.

Reading the finding did turn up a real slip of mine, now fixed in `69316c1`: adding the section left **two numbered 4**. The config checks are 4, page names is 5, links is 6.

PR #373, opened 2026-09-06
Day / Night: the claim is dropped where the wiring happens, not only where it is read (#367)

### LINKED ISSUE #367: The IR-cut notification does not reappear correctly on the Dashboard after being dismissed
I dismissed the notification about the unconfigured IR-cut filter on the **Dashboard**. Then on **Image → Day / Night** I configured the filter, saved it, removed it, and saved again. The notification **“Majestic cannot move the IR-cut filter”** appeared again on the Day / Night page.

The same notification should also reappear on the **Dashboard**, since the condition has become true again. Instead, on the Dashboard it appears only for a fraction of a second and then disappears.

It looks like dismissing the notification prevents it from being shown again even after the condition that triggered it occurs again.

openipc-ai:
Fixes #367.

The Dashboard lets an owner record, on the camera, that it has no IR-cut filter fitted. Nothing measurable separates that from a filter nobody has wired, so the owner is the only one who can say — and the answer is a property of the hardware, so it lives in `/etc/webui/ircut.conf` rather than in one browser's storage.

Wiring a filter contradicts the claim, and the claim was already dropped the moment a pad appeared. What went wrong is **where** that drop ran.

## The claim outlived the wiring that contradicted it

It ran in `a/dashboard.js`, on the first load that read a pad out of the config. That made the Dashboard the only witness to a wiring change, and the Dashboard is the one page on which nobody wires anything.

Wire the filter on Day / Night, save, take it away, save again: no Dashboard was open at any moment when a pad existed. The claim survived, and went on hiding a warning about a filter the camera had just stopped being able to move. Day / Night said so and the Dashboard did not, which is the disagreement the reporter saw.

Nor can it be recovered afterwards. Clearing those fields **removes** the keys rather than zeroing them, so the configuration after the round trip is byte-identical to the configuration before it — no reader can tell that a pad ever existed. An invalidation rule has to be applied where the invalidating event happens, not only in the state it leaves behind.

So the rule moves into `a/ircut-check.js` as `claim()`, with the wiring question beside it as `wired()`, and **both** pages ask it: the Dashboard for the answer, Day / Night after every save and per-row reset, while the pad is still there. Day / Night decides nothing and does not read the reply — it only puts the question at the one moment it can be answered differently.

## Where each half of it lives, under the rules #372 set

`j/ircut.cgi` is **unchanged**. It stores the claim and does nothing else, which is what a CGI under `j/` is still for: the claim is an owner's word about hardware, kept under `/etc/webui`, and majestic neither holds it nor serves it. The wiring it is weighed against is a different matter and comes from the daemon — `/api/v1/config.json`, in the browser, where both callers have already fetched it. Nothing here reads the config file, and nothing relays a majestic value through a CGI.

The endpoint is named once now, in `claimIo()`, with `read`/`clear`/`dismiss` on it, instead of at three call sites.

## And the flash, which is the other half of the report

"On the Dashboard it appears only for a fraction of a second and then disappears."

`dashboard.js` started `noFilter` as `false` deliberately — an ask that did not happen must not silence a fault — and painted as soon as the config arrived. So the banner painted, then withdrew itself when the claim landed a frame later: measured at a 40 ms poll on an hi3516ev200, shown at 57 ms and hidden by 108 ms, on **every** load of a camera whose owner had legitimately dismissed.

"We have not asked yet" and "the answer is no" are two states now. A failed ask still counts as asked, and `claim()` answers "does not stand" to one, so the original invariant holds and a fault is never silenced by a question that did not get through.

## Tests

The `claim` group in `tests/ircut-check.test.js` drives the reported sequence against a stub of the endpoint: wire → save → unwire → save gives the warning back. It also keeps **the bug itself** — the same sequence with nobody asking while the pad exists — so the reason Day / Night has to make this call cannot be quietly deleted later. Plus the two failure directions: an unreachable camera does not silence the banner, and a delete that failed is not reported as a delete.

Every way this fails is a banner that does **not** appear, which is indistinguishable from a camera that is fine, and reproducing it for real needs an owner who pressed Dismiss, a filter wired after that, and the same filter taken away again.

## Verified end to end

Driving the real pages in a headless browser against an hi3516ev200 running this branch, following the reported sequence exactly:

1. Dashboard with nothing wired — the warning appears and stays. Pressing the × records the claim (`{"noFilter":true}`) and the banner goes.
2. Day / Night: wire both coils, Save — the claim is dropped at that moment (`{"noFilter":false}`).
3. Day / Night: clear them, Save again — still dropped, and the file is gone from the camera.
4. Dashboard: the warning is back at ~137 ms and steady, with the × offered again.

And the case the feature exists for: claim on file, no pad ever wired — silent for a full 10 s of 40 ms samples, with no flash.

That camera's majestic predates `GET /api/v1/get` and `/api/v1/gpio` (both 404 there); nothing in this change uses either.


openipc-ai:
Superseded. This proposed keeping the Dashboard's "this camera has no IR-cut filter" claim and making its invalidation reliable. The claim should not exist: majestic already stores that statement as "Drive the IR-cut filter", and `diagnose()` has always been silent with that off and no pads assigned — the missing-pin finding is in the `else` of that switch. So the claim was a second spelling of an existing setting, and every fault in #367 followed from the duplication rather than from where the invalidation ran.

Replaced by a deletion of the whole mechanism.

PR #374, opened 2026-09-06
Day / Night: the camera already stores "no filter here", so stop storing it twice (#367)

### LINKED ISSUE #367: The IR-cut notification does not reappear correctly on the Dashboard after being dismissed

openipc-ai:
Fixes #367.

The Dashboard warns when nothing is wired to the IR-cut filter, because that fault is invisible until nightfall — the camera streams, records, answers ONVIF and reports healthy counters while sending a magenta picture. A camera that has no filter **fitted** is not faulty, though, and nothing measurable separates the two. So the banner carried a `×` that recorded the owner's word in `/etc/webui/ircut.conf`, written and read through a CGI of its own.

## majestic already stores that statement

**Drive the IR-cut filter** off is it. And `diagnose()` has always been silent on a camera with it off and no pads assigned, because the missing-pin finding is in the `else` of that switch:

```js
const ircutParked = nm.irCutEnabled === false;
if (ircutParked) {
    if (driveable) { …“the filter is switched off” note… }   // no pads: nothing at all
} else if (!driveable) {
    …“Majestic cannot move the IR-cut filter”…               // the banner
}
```

So the `×` was a second spelling of a setting the camera already held, kept somewhere only one banner could read it — and everything that went wrong with it followed from the duplication rather than from any detail of how it was implemented. A private claim has to be dropped when the wiring contradicts it. That invalidation ran on the Dashboard. The Dashboard is the one page on which nobody wires anything. Wire the filter on Day / Night, take it away again, and the claim outlived the wiring that contradicted it, hiding a warning about a filter the camera could no longer move — while Day / Night, which does not consult the claim, went on showing it. That disagreement is what was reported.

Nor was it recoverable afterwards: clearing those fields **removes** the keys rather than zeroing them, so the configuration after the round trip is byte-identical to the configuration before it.

## So the mechanism goes, rather than getting a better invalidation rule

Deleted: `www/cgi-bin/j/ircut.cgi`, the `×` and its wiring in `dashboard.cgi` / `dashboard.js`, and `wired()` in `ircut-check.js` — which existed only to decide when to drop the claim, and asks exactly what `diagnose()`'s own `driveable` asks.

The finding names the switch instead:

> Nothing is connected to the filter, so nothing moves it and it stays wherever it powered up. Left open in daylight it makes the whole picture magenta. **If this camera has no IR-cut filter fitted, turn off "Drive the IR-cut filter" and nothing here will ask again.**

That answer is better than the one it replaces in every direction: the daemon holds it, Day / Night shows it, a config backup carries it, every browser reads the same value, the setting is true of the camera rather than of one page's banner — and there is no second copy left that can go stale.

Net **−168 / +73**.

## Two consequences worth stating

**An owner who had already dismissed will see the banner once more after upgrading.** The sentence now tells them which switch ends it for good. Nothing here writes that switch on their behalf — a camera's configuration is not ours to change on an inference about what a previous click meant.

**Deleting a shipped file uncovers the firmware's own older copy.** `sbin/updatewebui` prunes the overlay copy of a file a release stops shipping, so the `/rom` version shows through. That is inert here: nothing requests the endpoint any more, and the file it managed is simply never read again. It is why this is a deletion rather than the 410 tombstone `j/gpio.cgi` needed — that one actuated hardware on a GET and coordinated with nothing.

## Tests

The silence now carries the whole feature and says nothing out loud, so the tests say it: switched off with nothing wired produces **zero** findings — not even from the picture — while the switch left on still raises the banner, and that banner names the switch. The `wired` group now asks `diagnose()` directly, which is the live question.

## Verified on hardware

Driving the real Dashboard in a headless browser against an hi3516ev200 running this branch:

| state | Dashboard |
|---|---|
| no pads, **Drive the IR-cut filter** on | banner at 60 ms and steady, with the sentence above and no `×` |
| no pads, switch **off** | silent |
| opening coil wired, switch on | silent |
| pads taken away again | banner back at 51 ms |

The last row is the reported sequence's end state, and the warning returns because there is no longer a claim that could survive it.


openipc-ai:
Fixed in 073796c — the finding was right, and worse than it reads.

The settings form is drawn from the daemon's schema, so on a build without `nightMode.irCutEnabled` there is no such switch, while `diagnose()` treats the absent key as *not parked* and raises the banner anyway. That sent those owners to look for a control that is not on their page. And it is not a narrow case: the key landed in the daemon two days ago, so it is most cameras in the field today — this had removed their only way to silence the warning.

The sentence now states the limitation instead:

- key present → *"…turn off \"Drive the IR-cut filter\" and nothing here will ask again."*
- key absent → *"…there is nothing here to say so with: the switch that settles it, \"Drive the IR-cut filter\", arrives with a newer firmware than this one."*

Absence is a sound signal and a default is not mistaken for one: `/api/v1/config.json` reports the **effective** configuration, so a key at its default is still present in it. Measured on an hi3516ev200 by resetting the key — gone from `majestic.yaml`, still in `config.json` — so only a daemon that has never heard of the key leaves it out. Tests cover both wordings and that the fault itself is stated either way.

PR #377, opened 2026-09-06
Day / Night: drop a compatibility branch for a state that cannot happen

openipc-ai:
The no-pins finding names the setting that answers it — turn off **Drive the IR-cut filter** on a camera that has none fitted — and it did so only where the daemon published that key, saying *"arrives with a newer firmware than this one"* otherwise.

There is no such camera. The WebUI and majestic are built into one firmware image and move together, so a camera running this page has the switch this sentence names. The other arm was unreachable prose describing a configuration the project does not support, and writing it made that mismatch look like a state worth planning for.

Said unconditionally now; the two checks asserting the unreachable wording go with it. **−37 / +8**, no behaviour change on any supported camera.

### DIFF
--- a/tests/ircut-check.test.js
+++ b/tests/ircut-check.test.js
@@ -687,24 +687,9 @@ function runRest() {
 		const noPins = (nm) =>
 			ic.diagnose(nm, null, null).filter(x => x.id === 'no-pins')[0].detail;
 		check('and that finding names the switch that answers it',
-			/turn off "Drive the IR-cut filter"/.test(noPins({ irCutEnabled: true })));
-
-		// But only where the camera has it. The switch is recent and the
-		// settings page is drawn from the daemon's own schema, so on a build
-		// without the key there is no such control to go and find — sending
-		// someone to look for one would be the page inventing an answer.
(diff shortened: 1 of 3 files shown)


openipc-ai:
Not taking this one — it asks for exactly what the PR removes, and the premise behind it is what the PR is correcting.

**There is no camera in that state.** The WebUI and majestic are built into one firmware image and move together; backward compatibility with an older daemon is not something this UI supports. So a camera running this page has `nightMode.irCutEnabled`, and a branch keyed on its absence is unreachable. Keeping "distinct wording for that state" would be writing prose for a configuration that cannot occur — and, worse, making it look like a supported one.

**But the evidence is a fair catch, and I have acted on it.** The review points at `mj-settings.js`, where the comment above the parked test said an absent key "on an older daemon means the switch does not exist there". The `=== false` test is right and stays — absent is not off — but that justification was the same unsupported-state story, in the one place a reader would go to check it. It now gives the real reason: a key the page was never given is one it knows nothing about, and reading that as "switched off" would grey out the control on a camera whose filter is working.

So: finding rejected, its citation fixed.

Issue #274, opened 2026-09-01
Add `MJPEG` as a selectable video source for `Live` and `Live adjustments`


usa-:
Raspberry Pi 4B, Chromium.

On the Live page, when `Main` and `MSE` are selected, the chip showing the current stream displays `MPEG`, while the `Main` and `MSE` buttons remain active, and no message explains why the selected mode cannot be played.

<details><summary>Proof</summary>

<img width="995" height="608" alt="Image" src="https://github.com/user-attachments/assets/aaa6889c-e015-4b27-8ba8-41efdaba0e61" />

</details> 

Since `MPEG` is still available as one of the video sources for Live, despite us abandoning this approach in one of the now-closed issues, I suggest adding an `MJPEG` button alongside `Main`, `Sub`, and `Auto`. The `Auto` algorithm should also consider the MJPEG stream as a potential source.

It would also be useful to have an `MJPEG` button on `Camera → Settings → Image → Live adjustments`.

openipc-ai:
@usa- — two things in this report, and they get opposite answers. The behaviour you photographed is a real bug, and it is fixed and on master. MJPEG is not going to become a selectable video source — and the reason is not only policy, it is that we intend to solve your actual problem properly. More on that below.

## Fixed

Merged as #279 and #280. Run `updatewebui` on the camera to get it; nothing on the majestic side is needed.

What you will see on that Pi now, instead of `MJPEG` and two lit buttons:

- The picture is still the MJPEG fallback, but the page **says why**, on the video: *"This browser can't decode the Main stream's H265 video. Showing the MJPEG fallback."*
- **Neither transport button is lit**, because neither is carrying the picture — and that is also what makes them work as a retry: pressing one now re-runs the whole chain, where before pressing the one already lit did nothing at all.
- **Main / Sub / Auto do something.** Pressing a channel restarts the chain on that channel. If your camera has a substream on H.264, pressing **Sub** should give you real video on that Pi — the browser refuses H.265, not the camera.
- A retry **keeps the picture you had** until the replacement actually plays. It no longer trades a working MJPEG frame for a black stage.
- With no JPEG channel configured, the chip reads `unavailable` rather than claiming `MJPEG` for a picture that does not exist, and the note carries the reason as well as the Enable-JPEG remedy.
[...]

PR #381, opened 2026-09-06
Show every camera the camera has, and play them

openipc-ai:
A camera is not always one camera. majestic can carry a second **sensor**
(Rockchip) or publish a **USB webcam** as a second camera, both addressed as
`stream_id = 3*camera + subtype`, and this UI showed one. The streams were there
and reachable — by knowing to ask for `?channel=1` by hand. It now asks
`GET /api/v1/sources` and renders what the camera says it has.

Needs the daemon side (widgetii/majestic#642); majestic and majestic-webui are
synced by design, so there are no compatibility branches here.

## Three pieces

**`mj-sources.js`** is the decision module: which sources can be watched, which
stream of one of them, and what transport its codec implies. Pure, so it is
testable, and shared because every page that shows video makes the same three
decisions and three copies would drift within a release.

The subtle one is `flowing`. It is **absent** on an MJPEG stream — that field
comes from the parameter sets a keyframe publishes, and a JPEG stream has none —
so a gate that read a missing key as false would hide every USB webcam. It is
also not a liveness gate at all: a stream that is present but has not published
a keyframe yet is a stream that is about to, and refusing to offer it would make
a camera look sourceless for the first seconds after a restart.

**`preview-multipart.js`** makes the MJPEG stream a rung of the transport ladder.
It was the Preview page's terminal fallback: a bare `img.src = '/mjpeg'` written
from `showFallback()`, outside the swap, with its own visibility flag and its own
retry path. That was survivable while it was the last thing tried and there was
one camera. A webcam publishing MJPEG is a **source**, not a failure — for most
of them the only thing they publish — so the same picture has to be reachable as
a first choice, on an element the swap owns, for a camera other than the on-board
one. Two paths to one stream would drift, and one of them could not name a camera.

It is called `multipart` and not `mjpeg` because that string is already a player
*state code* meaning "this rung gave up", and a transport kind of the same name
would collide in every `onState` handler on both pages.

Making it a rung is also what closes the fallback's own hole. `showFallback()`
stopped the swap, so a retry out of it was promoted the instant it was made —
trading a working picture for a black stage and a negotiation that could still
fail, which inverts the one rule `preview-swap.js` exists to enforce. As a live
player it gets that rule for free. The explanation survives with it: a viewer
handed the MJPEG picture in place of the one they asked for is still told why,
which was half of #274.

**The chooser** is built in JS and simply not appended when there is one source.
Markup for a picker nobody can use is markup that has to stay hidden correctly for
ever, and this page has already shipped enabled radios for a channel the camera
did not have. Names come from `j/locale.cgi`, because the daemon reports a *kind*
and refuses to name it — it has no business deciding what language an operator
reads. A second-sensor board gets ordinals, since "Sensor" twice would name
neither.

Two things follow from a source that is not the on-board one. Its stream_id goes
on the wire while the `video0`/`video1` sections keep describing the on-board
channels, which is right — there is no `video0`-for-camera-1 key, and both
second-camera implementations reuse camera 0's values. And whatever those sections
decide is clamped to what the source actually publishes: a webcam with one MJPEG
stream has neither a Main nor a Sub to have chosen between, so those controls go
away rather than sit there doing nothing.

## Three more surfaces

**Stream URLs** listed 0/1/2 and nothing else, so a second camera's endpoints were
undocumented. The extra sources are rendered from the camera's own answer, because
which ids exist and which of them RTSP will answer for is something only the camera
knows — the JPEG track is opt-in, and a URL that 404s is worse than no URL.

**The snapshot button** took `/image.jpg`, which is camera 0's JPEG channel
whatever is on screen: watching the USB camera and pressing it gave you a picture
of the sensor. It now takes the camera being watched, and appears only for a camera
that has a snapshot to give.

**Recordings** was the one that was actually broken. A second camera's writer
inserts `-cam<N>` before the extension, and the filename parser required the
extension straight after the time — so every clip a dual-camera board recorded fell
into the "clip(s) whose name has no time" bucket: placed nowhere, drawn on no
timeline, and looking like a naming fault rather than a second camera.

The day is now built from **one** camera's clips, which is not a filter over what is
drawn: the day model is a line of footage — gaps, joins, durations, "still
recording" — and two cameras share every minute, so a combined line would report a
gap wherever one camera has a clip and the other does not, and would join two
cameras' clips into a stretch that never existed.

## Size

`tools/build-dist.sh` then xz: **+3,260 bytes**, against roughly 19 KB of headroom
on `gk7205v300_lite`.

## Tests

`npm test`: 1,550 assertions, 0 failures. New: `tests/sources.test.js` and
`tests/multipart.test.js`; extended: `transport.test.js` (the new `impl()` arm and
the rung's gate), `staging.test.js` (the fallback groups rewritten to the rung, plus
the chooser), `timeline.test.js` (the `-cam` suffix).

Verified against a gk7205v200 with a USB webcam: the page renders the chooser, the
labels come through from `locale.cgi`, and the camera reports both sources with the
webcam moving between stream 5 and stream 3 as its codec changes. I could not drive
a browser from that environment, so the Live page's chooser is covered by tests and
by server-side markup checks rather than by a human looking at it.


openipc-ai:
Thanks — seven of the ten were real, including both security ones. Fixed in
d1bdc5d; the commit message says what each fix does and why.

Worth singling out **10** and **2**, which were the two that mattered: unescaped
locale text reaching an inline `<script>` on an authenticated page, and
unvalidated response fields reaching `innerHTML`. Both closed, and the escaping
verified against a hostile locale file on a real camera's busybox `ash` rather
than on a desktop shell.

**3** is right too, and I want to be explicit that it was mine rather than a
repository convention: citing majestic's *issue numbers* is established practice
here, citing its source paths and internal symbols is not, and this series
introduced it. Those comments now describe what is observable — endpoints,
response fields, and the shape of a filename. (The PR description names the
repository, which I have left: it is the dependency, and a reader who cannot open
it still needs to know the two ship together.)

Three I am declining, all of the same shape — they ask for the MJPEG stream to
become a peer channel choice:

**4 (MJPEG as a third channel radio)** and **5 (Auto picking it)** — the picker
chooses between *encoder channels*, which are the things `video0`/`video1`
configure and which `autoPick()` compares by size to fit the stage. The JPEG
channel is a 5 fps snapshot stream at several times the bitrate of the same
picture in H.264; putting it beside Main and Sub would offer it as an equal, and
letting Auto choose it proactively would sometimes hand a viewer that instead of
video they can decode. It is reachable exactly where it should be: as the floor
of the transport ladder, when nothing above it can play — which is what this PR
makes work properly for the first time, and which is also how a webcam that
publishes *only* MJPEG is played, without any of it being a channel choice.

**6 (MJPEG in the Live adjustments panel)** — that panel exists to show the
channel whose settings are being dragged. MJPEG is a different channel with its
own geometry and quality, so it would not reflect the knob being moved; showing
it there would make the panel lie about what it is previewing. The panel is
deliberately camera 0 only for the same reason, and I have left it that way.

If any of the three is wanted as a feature, it is a separate change with its own
UX question to answer — I would rather not decide it inside this one.

PR #386, opened 2026-09-07
Day / Night: say why the sweep is refused, on the page

widgetii:
**Find them for me** disables itself when the camera cannot say which pads are already spoken for, and put the reason in the button's `title` attribute alone.

Fifty lines below, `syncTestBtn()` writes the *test* button's reason into `#mj-ircut-why`, with the comment that says why:

> The reason goes on the page, not only in the title. A tooltip is not an explanation on a touchscreen, where there is no hover at all, and a control that refuses without saying why is the thing this whole panel exists to stop happening.

This button was the one place in the panel that broke that rule. The result is a greyed-out control with no visible reason and no next step — which is exactly how it was reported: *"How am I supposed to use `Find them for me`? It's not active."*

### What changed

The reason now renders in `#mj-ircut-find-why`, its own element rather than the test button's. The two have different lifetimes: this one is a property of the camera, decided once from the enumeration, while `syncTestBtn()` rewrites `#mj-ircut-why` on every map edit and would wipe it. Same `small text-secondary` styling, no new Bootstrap classes, so `bootstrap.min.css` is unaffected.

The sentence also now says what to do instead — set the coils by hand on the pin map, which writes a number into a field and moves nothing, and let **Test the filter** check them. That is the answer the reporter actually needed, and the panel already supports it: nothing about setting a pin by hand is blocked when a sweep is.

The advice it replaces told the reader to check that debugfs is mounted. That is no longer the gate — the camera establishes pad ownership without it now — and on a camera that still cannot answer there is nothing the reader can do from this page anyway, so naming a filesystem was pointing at a lever that isn't there.

### Verification

`npm test` and `tools/lint-templates.sh` green.

Exercised on an hi3516ev200 whose `/api/v1/gpio` reports `ownersUnknown: true`: the button disables and the note renders in its place, with the tooltip kept for pointer users. The disabled state itself was confirmed by pixel rather than by eye — on the dark theme a disabled `.btn-primary` is still recognisably blue, sampling `(18, 82, 178)`, which is Bootstrap's 65% opacity blend and not the full `#0d6efd`.

Not yet exercised as part of a full `build-dist` install on a camera; happy to do that before merge if wanted.

### DIFF
--- a/www/a/mj-settings.js
+++ b/www/a/mj-settings.js
@@ -6394,10 +6394,21 @@
 			if (info.ownersUnknown) cant.push('which pads the kernel already holds');
 			if (info.ptzUnknown) cant.push('which pads the PTZ driver is on');
 			find.disabled = true;
+			// What is refused is the SWEEP, and only the sweep: it is the one
+			// thing here that drives pads nobody has vouched for. Testing a
+			// filter the camera already knows about still works, and still
+			// moves it. "Nothing may be driven" said otherwise, and then the
+			// same sentence sent the reader to a control that drives.
+			//
+			// It ends at Save because the pin map only STAGES: the map writes
+			// into the hidden fields, while the test reads the configuration
+			// the camera is running on. Skip the save and the verdict is about
+			// the old wiring — which is the one answer this panel must never
+			// give. The scan's own success path has always said it this way.
 			const why = 'This camera cannot say ' + cant.join(' or ') +
-				', so nothing may be driven. Set the coils by hand on the pin ' +
-				'map — that writes a number into a field and moves nothing, and ' +
-				'Test the filter will check them.';
+				', so the sweep may not drive pads it has not been told about. ' +
+				'Set the coils by hand on the pin map, Save, then Test the ' +
+				'filter checks them.';
 			find.title = why;
 			// The reason goes on the page and not only in the title, for the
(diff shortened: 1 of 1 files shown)


widgetii:
Both findings were right and both were about the sentence. Fixed in daa627c.

**1. Manual tests can use old wiring.** Correct — the pin map only *stages*: `pushAssign()` writes into the hidden fields and marks the form dirty, while `testBlocker()` and `runIrcutTest()` read `state.config`, the configuration the camera is running on. An operator following the old wording literally on a camera that already has pins saved would get a verdict about the previous wiring, which is the one answer this panel must never give. The sentence ends at Save now, matching what the scan's own success path has always said — `Pins staged — Save, then test the filter.`

**2. Conflicting advice.** Also correct, and the sharper of the two: unknown ownership refuses the *sweep*, which is the one thing here that drives pads nobody has vouched for. Testing a filter the camera already knows about goes through the day/night path and still moves it. So "nothing may be driven" contradicted the control sitting next to it, and then the same sentence sent the reader to that control. It now says the narrower thing that is actually true — the sweep may not drive pads it has not been told about.

New text:

> This camera cannot say which pads the kernel already holds, so the sweep may not drive pads it has not been told about. Set the coils by hand on the pin map, Save, then Test the filter checks them.

`npm test` and `tools/lint-templates.sh` green.

PR #387, opened 2026-09-07
Day / Night: a dimmable lamp is a lamp that is connected

widgetii:
The **Night illuminator** row read *"not set"* on a camera with a lamp wired and working.

The row draws from `nightMode.backlightPin`, which holds a **pad**. A dimmable lamp is configured as a PWM **channel**, and which GPIO that channel takes over is the processor's business — named in no field on this page. So the row was reading the one place the answer could not be, and reporting the absence as an absence of lamp.

### What it says now

The pad where the camera can name one, the channel otherwise, and `dimmable, on pwm1` where the hint goes. Never "not set", which is the single thing that is certainly wrong here.

Three things had to follow:

- **The active-low chip is reachable again.** A dimmable lamp honours `backlightInvert` exactly as a switched one does — the invert is applied to the duty — so a lamp wired the other way round needs the same control. The chip only ever appeared for a row that counted as set.
- **The row points at its pad when clicked**, since that is the only way the map says which pad the channel took.
- **The "not a pin on this processor" branch is guarded**, or a lamp with no pad in the assignment would be accused of naming a pin the camera does not have.

### The map marks that pad too

…and refuses to let a coil be assigned onto it. Judged by whether the role's own assignment holds the pad the camera reports, **not** by the role's name: the same role covers a lamp on a plain pad, where the field *is* the editable truth and the pad must stay pickable. Only where the two disagree is the pad somebody else's — and then it is exactly as unpickable as the PTZ driver's, because nothing on this page can move it and a coil assigned to it would fight the lamp for the mux.

### Verified against a live camera

Rendered against an hi3516ev200 with the lamp on `pwm1`, using the whole tree served locally with the API proxied to the camera — so the page ran one consistent tree against real data without installing anything:

- Row: **`4`**, `dimmable, on pwm1`, with its `lamp on = HIGH` polarity chip.
- Map: `<button class="mj-pin mj-pin-owned" data-pin="4" title="Pin 4 — used by the night illuminator" disabled>`
- Zero console errors on the page.

`npm test` and `tools/lint-templates.sh` green.

### Note on the pad number

The pad comes from the camera's own enumeration, which is the only place it exists. A camera that does not report one still gets the honest answer — the channel name instead of the pad — rather than "not set".

### DIFF
--- a/www/a/ircut-map.js
+++ b/www/a/ircut-map.js
@@ -138,15 +138,19 @@
 			if (FOREIGN[a.role]) owned[a.pin] = FOREIGN[a.role];
 			// A DIMMABLE lamp is configured as a PWM channel, and the pad that
 			// channel takes over is the SoC's, named in no field on this page.
-			// So the camera reports a pad here that the role's own assignment
-			// does not have — and that pad is exactly as unpickable as the PTZ
-			// driver's, for the same reason: nothing on this page can move it,
-			// and a coil assigned to it would fight the lamp for the mux.
+			// It is exactly as unpickable as the PTZ driver's pad: nothing here
+			// can move it, and a coil assigned to it would fight the lamp for
+			// the mux.
 			//
-			// Judged against the assignment rather than by role name, because
-			// the same role covers a lamp on a plain pad, where the field IS
-			// the editable truth and the pad must stay pickable.
-			if (a.role === 'backlightPin' && assign.backlightPin !== a.pin) {
+			// Told by the caller, not worked out from the assignment. The first
+			// try owned the pad only where it DIFFERED from backlightPin — an
+			// equality standing in for "the field is authoritative", which it
+			// is not: with a channel selected the camera ignores that field
+			// whatever it holds, so a stale pin that happened to equal the PWM
+			// pad handed the lamp's own pad back to the picker. The lamp being
+			// on a channel is a fact about the configuration, and the page that
+			// reads the configuration is the one that should say so.
+			if (a.role === 'backlightPin' && opts.pwmLamp) {
 				owned[a.pin] = 'used by the night illuminator';
 			}
 		});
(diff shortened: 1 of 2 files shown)


widgetii:
Both findings were one error twice, and it is the one I should have caught: I used `nightMode.backlightPin` to decide whether the lamp is dimmable — its absence in the row, its value in the map — when the camera ignores that field entirely whenever a channel is selected, and says so in the field's own hint. Its presence proves nothing about how the lamp is driven, so neither test measured what it claimed to. Fixed in 943352d.

**1. Dimmable lamps appear switched.** Correct. A lamp pin left over beside a selected channel hid the channel and drew a dimmable lamp as an ordinary switched one, naming a pad nothing drives. The row detects the channel alone now.

**2. Coils can share an active lamp pad.** Correct, and the worse of the two — the pad was owned only where it *differed* from `backlightPin`, so a stale pin that happened to equal the PWM pad handed the lamp's own pad back to the picker. That is exactly what the marking exists to prevent. The map is now **told** that a channel is driving the lamp rather than working it out: it holds no configuration, the fact lives in one key, and inferring it from the pad assignment is what went wrong.

One judgement call worth flagging, since it is not what the agent prompt suggested. The pad is shown only where the camera reports **exactly one** for the role. With a lamp pin also configured it reports two, and which of them the channel took is not something this page can tell without guessing — so the channel name is shown instead. That is still an answer, unlike "not set", which is the bug this change is about. Two entries naming the *same* pad are deduplicated first: that is one pad, not an ambiguity.

Clicking the row now points at the lamp's own pad in preference to a pin the camera ignores.

**Verified against the live camera in both shapes**, lamp on `pwm1` at pad 4:

| case | pad 4 | row |
|---|---|---|
| stale pin **= 4** (the finding's case) | `mj-pin-owned`, `disabled` — was pickable before | `dimmable, on pwm1` |
| stale pin **7** alongside | both 4 and 7 refused, matching what the camera itself refuses | `dimmable, on pwm1`, not "pin 7, switched" |
| no stale pin | owned, disabled | `dimmable, on pwm1 · 4` |

No console errors on the page. `npm test` and `tools/lint-templates.sh` green. The camera's config was put back afterwards.

Issue #370, opened 2026-09-06
GK7205+IMX307 does not switching day\night in auto mode


sergshadrap:
Problem

Automatic Day/Night switching cannot be reliably configured for a GK7205V200 + IMX307 camera with using light monitor.

The IMX307 remains usable in color at illumination levels where the IR-board CDS already switches to NIGHT. The current configuration therefore switches the camera to NIGHT too early and loses useful sensor sensitivity. That was recently updated from very old firmware , where the custom day\night script was in the play. 

The camera provides:
GPIO15: IR-board CDS/IRSTATUS (1 = light, 0 = dark)
GPIO16: IR illumination control

Additionally, the isp_again value reaches approximately 22924 in darkness, but with maxThreshold: 14000 the expected NIGHT transition does not occur, manually I can switch the modes properly. Different values of borders for day\night mode where tested . none of them lead to correct transition.

Hardware tested on
SoC: GK7205V200
Sensor: IMX307
Board: custom GK7205V200 camera
Firmware: 2.6.09.03-lite
Firmware commit: master+347335a, 2026-09-03
Majestic: Lite HiSilicon (GK7205V200)
Majestic commit: master+f096923, 2026-09-03

openipc-ai:
@sergshadrap — thank you, this turned out to be two separate things, and the second one is a bug you found for us.

## Why your thresholds do nothing

With **Automatic day/night** on, a wired daylight sensor takes the decision **before** the thresholds are looked at — the camera stops there and never reads them. Your GPIO15 CDS is that sensor, so `maxThreshold: 14000` is inert, and that is why no value of it changed anything:

> Different values of borders for day\night mode where tested . none of them lead to correct transition.

The camera will confirm it directly:

```
curl -u root:PASS http://<camera>/metrics/night | grep night_mode_source
```

`1` means the daylight sensor has the wheel, `2` means the thresholds do. On your setup it will read `1`.

So the IMX307 is not being let down by a missing feature — the camera is configured to let the CDS decide, and the CDS trips at a light level the sensor is still comfortable at. That matches your description exactly.

## Two ways forward, both starting by clearing the sensor pin

Clear **GPIO pin for light sensor**, then either:

- **Set both thresholds** — `minThreshold` *and* `maxThreshold` — to decide by gain, or
[...]

PR #388, opened 2026-09-07
Day / Night: automatic mode says what night is waiting for

widgetii:
Uncalibrated, the camera does not decide by gain at all. It waits for the automatic exposure to run out of shutter and gain, **and** for the picture to be dark with it.

The panel said the first half in words — *"night comes when the exposure runs out"* — and then plotted the gain, which answers neither.

So a camera sitting in day with its gain at 22x looks like one ignoring an obvious night, and the page has nothing further to offer. The plot it does draw is structurally blind to the direction being waited on, and the two readings that actually decide sit on the Dashboard's luminance panel rather than beside the sentence that names them. That is [#370](https://github.com/OpenIPC/majestic-webui/issues/370), where the reporter posted a full ISP dump asking why night never comes and the answer was not in it.

### The sentence finishes itself now

| state | what it says |
|---|---|
| not there yet | `The exposure has not run out yet.` |
| both terms true | `The exposure has run out and the picture is dark with it.` |
| run out, vetoed | `The exposure has run out, but the picture is still too bright to call it night (200 of 255).` |

The third carries the luminance deliberately: it is the state nothing on the page could distinguish from "nothing is happening", and the number is what tells its owner whether the scene or the reading is the odd one.

### What it does not do

- **Silent where the camera does not publish the reading.** A missing answer is not a "no", and *"the exposure has not run out"* about a camera that never said would be the confident wrong answer this panel exists to avoid.
- **An explicit night gain keeps the old sentence.** Gain is then genuinely the rule, and describing a mechanism that is not running would be worse than saying less.
- **Decides nothing.** The `128` is the daemon's own veto threshold, mirrored here to word a sentence.

### Tests

Five cases in `tests/ircut-check.test.js` — each of the three states, the unpublished-reading case, and the explicit-night-gain case that must keep the old wording. Full suite and `tools/lint-templates.sh` green.

`monitorView` is a pure function and its `line` goes straight to `#mj-ircut-mon-line`, an element already rendering on hardware, so this adds no DOM and no new dependency.

### DIFF
--- a/tests/ircut-check.test.js
+++ b/tests/ircut-check.test.js
@@ -519,6 +519,37 @@ function runRest() {
 				isp_exposureismax: 1, isp_avelum: 200 }).line;
 		check('an explicit night gain keeps the old sentence',
 			!/run out/.test(byGain), byGain);
+
+		// In night the camera waits for the OTHER door. Saying "night comes
+		// when the exposure runs out" under the word "Night." reads as a
+		// switch that has not happened on a camera where it has, and the
+		// appendix made it an outright contradiction.
+		const atNight = (extra) => ic.monitorView({ lightMonitor: true },
+			Object.assign({ night_mode_source: 4, night_enabled: 1,
+				night_auto_gain_milli: 1400 }, extra)).line;
+
+		const settled = atNight({ isp_exposureismax: 0, isp_avelum: 40 });
+		check('night is not told how night arrives',
+			/^Night\./.test(settled) && !/night comes when/.test(settled), settled);
+		check('...it is told how day does',
+			/day comes when the gain settles back down/.test(settled), settled);
+		check('...and nothing is said about a ceiling it is off',
+			!/still at its ceiling/.test(settled), settled);
+
+		// The half the plot cannot show: gain under the day mark, and day
+		// still will not come, because the exposure is pinned.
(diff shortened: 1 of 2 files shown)


widgetii:
Right, and my own hardware render should have shown me this — it read **"Night.** Watching the sensor gain; night comes when the exposure runs out." The old sentence named the night door whichever side the camera was on, which was merely odd; the appendix turned it into a camera plainly in night being told night has not arrived. Fixed in 60d1284.

Rather than suppress the appendix in night, the sentence now follows the door actually being waited at:

- **Day** → night is what is waited for, appendix as before.
- **Night** → `day comes when the gain settles back down.`
- **Direction unknown** (`night_enabled` absent) → stops at the mechanism and claims neither. Naming the wrong door is the fault being fixed, so it does not guess one.

That surfaced the mirror of the gap this PR is about. Coming back to day needs the gain to settle **and** the exposure to be off its ceiling, and the plot only shows the gain — so a lamp that has brought the gain down under the day mark can sit there looking like a switch refusing to happen. It now says `The exposure is still at its ceiling, so day waits whatever the gain does.` — only while the ceiling is what is in the way. On every other tick it would be noise beside a chart already showing the thing being waited for.

Five more tests: night is not told how night arrives, is told how day does, is silent about a ceiling it is off, names a pinned one, and neither appendix is spoken with no reported side. Ten in the group, all green, plus the lint.

Issue #317, opened 2026-09-04
Live page loads without video in Opera for Android


usa-:
Opera for Android 101.2.5178.89973 (latest) has an issue with the initial Live page load.

When opening a camera for the first time, entering the password and saving it, then closing and reopening Opera, the Live page loads without video. If you go to Dashboard and then return to Live, the video starts playing normally.

The same camera works correctly in Chrome for Android, so this seems to be specific to Opera.

<img width="432" height="876" alt="Image" src="https://github.com/user-attachments/assets/10b5fe00-b120-4f27-a738-4097e624570a" />

openipc-ai:
@usa- — thanks for the clear steps. I tried to reproduce this here and couldn't: it is specific to Opera (Chrome for Android is fine, which matches what you saw), and Opera would not run in our automated Android test setup, so I can't watch it from this end. Three quick checks from you would pin down the cause decisively, because each one rules a different explanation in or out.

1. **Tap the video.** On the reopened Live page that shows no video, tap once directly on the black picture area (not a button). Does the video start?
   - Starts on tap → Opera is waiting for a tap before it will play, and the fix is on our side (start muted, or re-arm on the first touch).
   - Nothing happens → the page did not re-connect the stream after Opera restored the tab, which is a different fix.

2. **Opera settings.** In Opera → Settings, is **"Block media autoplay"** turned on, and is **Data savings / Turbo** on? (Opera's autoplay and proxy behaviour differ from Chrome's, and either could gate the first play.)

3. **Cold restart vs already-open.** Does it happen **only** after fully closing and reopening Opera (the tab restored on a cold start), or **also** if Opera is already open and you just switch back to the Live tab or pull-to-refresh it? This separates "Opera restored a frozen tab" from a general reconnect problem.

One more, if it is easy: on that broken load, what does the small chip in the **top-right corner of the video** say — `H265 …`, `MJPEG`, or is it not there at all? That tells me whether anything connected before the picture went missing.

No rush — whichever of these you can grab is enough to point the fix the right way.

usa-:
1. When I tap the black video area, the controls appear for a few seconds in several rows, just as they do when the video is working normally. Also, from time to time I get the bitrate adjustment message automatically, again as I would expect during normal playback.

[...]

PR #396, opened 2026-09-08
Live: retry a refused muted WebRTC start on a gesture, whatever the error is named

openipc-ai:
## What

Follow-up to #391. On Opera for Android the Live picture is missing on a fresh WebRTC load; the reporter confirmed the video is decoding and laid out but `paused=true`. #391 retried `play()` on the first user gesture — but only when the refusal was a `NotAllowedError`. After `updatewebui` the reporter said nothing changed, so the rejection Opera raises for a **muted** MediaStream refusal is not named `NotAllowedError`, and the retry never armed.

Chrome always allows muted autoplay, so this refusal **cannot be reproduced there** (verified: even `--autoplay-policy=user-gesture-required` still plays the muted stream). The `NotAllowedError` gate was a guess, and too narrow.

## The change

In the muted branch, drop the error-name check: a muted element left paused after a `play()` attempt is, on a fresh load, autoplay being refused, and the only actionable fact is that a gesture will start it. No audio track is requested while muted (`ontrack` fires once), so this is not the audio-track `AbortError` the unmuted branch guards. Transparent where autoplay is allowed — the `catch` never runs (lab WebRTC unchanged, 185 frames / 0 dropped).

## Tests

`tests/talkback.test.js` now rejects the muted `play()` with `AbortError` to prove the retry arms regardless of the error name (and still re-arms per attempt, does not stack).

Addresses the WebRTC half of #317. Pending the reporter's confirmation on Opera before merge, since it can't be verified against Chrome.

### DIFF
--- a/tests/talkback.test.js
+++ b/tests/talkback.test.js
@@ -274,10 +274,12 @@ async function playRetriesOnGesture() {
 	// load; the picture is ready but paused. The player must retry play() on the
 	// first user gesture rather than leave it paused for ever.
 	const env = makeEnv();
-	let plays = 0, reject = true;
+	let plays = 0, reject = true, rejectName = 'AbortError';
 	const video = {
 		muted: true, volume: 1, srcObject: null,
-		play() { plays++; return reject ? Promise.reject({ name: 'NotAllowedError' }) : Promise.resolve(); },
+		// Not NotAllowedError: browsers name a refused muted autoplay
+		// differently, and the retry must arm regardless (#317).
+		play() { plays++; return reject ? Promise.reject({ name: rejectName }) : Promise.resolve(); },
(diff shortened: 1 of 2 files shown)


openipc-ai:
Superseded by #414 (merged).

This PR's approach — arming the gesture-retry on a muted `play()` *rejection* — was the right instinct but the wrong trigger: the reporter's test showed it behaved identically to `main`, which proved `play()` doesn't reject on Opera for Android (it resolves without starting playback). #414 fixes that properly by arming on the observed *paused* state instead, keeps this PR's valid broadening (retry on any muted rejection, for browsers that do reject), and also fixes the same freeze in the MSE player. Closing in favour of it.

PR #397, opened 2026-09-08
Live adjustments: reflect the served WebRTC channel and say why, like the Live page does

### LINKED ISSUE #252: Live adjustments: reflect the served WebRTC channel and say why, like the Live page does
Follow-up to #240, requested by @usa- in https://github.com/OpenIPC/majestic-webui/issues/240#issuecomment-5475196202.

The Camera → Settings → Live adjustments preview has its own Main/Sub radios (`renderLive()` / `attachLivePreview()` in `mj-settings.js`) but none of the served-channel handling the Live page gained in #249: no chip, no message container, and the radios keep showing an unserved selection when WebRTC falls back.

Wanted, mirroring the Live page's behaviour on a camera whose majestic sends the `served` signalling reply:

1. move the radios to the channel actually served (without re-entering the stream switch or overwriting the remembered `chosenStream('live')` preference), and
2. show a dismissible message explaining why the selected channel could not be served.

Same compatibility rule as #249: on an older majestic nothing arrives and the panel behaves exactly as today.

openipc-ai:
Closes #252.

## What

The Live View page follows the channel a WebRTC session actually serves and explains a fallback (#240/#249). The **Live adjustments** preview did not: the picker, chip and fallback chain moved into the shared `MajesticPreview` component (`mj-preview.js`), and it never hooked the camera's `served` signalling reply. So on a camera that couldn't serve the requested channel over WebRTC, the panel showed the *other* channel with the radios still on the unserved selection and nothing saying why.

(The issue text points at `renderLive()`/`attachLivePreview()`, but the radios have since moved into `mj-preview.js`, so the fix goes there — and benefits any consumer of the component, not only the Live adjustments panel.)

## The change

Hook `onServed` in the component's attach and, on a mismatch, move the radios to the served channel and show a dismissible message. Mirroring the Live page:

- radios moved by writing `.checked` — no change event, so `goToStream` doesn't re-enter and cut the very session that reported this;
- `chooseStream` left untouched — a daemon fallback isn't the viewer's choice, and the next mount asks for their channel again;
- the reason is said **once**, not on every reconnect;
- a reply that arrives while its session is still a trial is applied only once that session is on screen;
- on any **non-WebRTC** transport the served channel is moot (MSE and the software rung subscribe exactly), so a prior mismatch's message clears; a deliberate channel change resets it too.

There's no chip here to also correct (that's the Live page's), so this is the radios and the message alone. Same compatibility rule as #249: an older majestic sends no reply, `onServed` never runs, the panel behaves as before.

## Tests

The fiddly part — mismatch detection, say-once, and leaving a standing explanation up when an internal reopen is answered with the adopted channel while the viewer's ask is still unmet — is a pure `servedDecision()` the caller applies, covered by `tests/served-channel.test.js`. It's tested as a pure function because the stage builds itself from `innerHTML` and can't be mounted without a real DOM.

Verified on a lab hi3516ev300 that the Live adjustments preview still mounts and plays over WebRTC with the change (339 frames, 0 dropped, no console errors). The mismatch path itself needs a camera that refuses a channel over WebRTC, which Chrome does not reproduce.


openipc-ai:
Follow-up commit: the served-channel rule is now a shared module (`preview-served.js` / `window.MajesticServed`), not a copy.

The first commit added the handling to the shared preview component by copying the Live View page's — which would have made it a *fourth* place the same rule is kept right (next to the fallback walk and the wasm-retry ladder that already live in both `preview-page.js` and `mj-preview.js`). This lifts it out:

- `decide()` — the pure rule (mismatch, say-once, leave-a-standing-message-up when a reopen matches the adopted channel while the ask is unmet).
- `make()` — a small applier holding the say-once state, driving each page's own radios/message through callbacks, with the Live page's Auto exception.

The settings preview uses `make()`; the Live View page keeps its own state vars and resets and now words its one decision from the shared `decide()` instead of a second copy. `tests/staging.test.js` and `tests/auto-source.test.js` already assert the Live page's served chip/radios/message and still pass (that's what proves the page is unchanged), and `tests/served-channel.test.js` covers the module. Verified on a lab hi3516ev300 that both pages still play over WebRTC with no console errors.

(The pre-existing duplication of `nextRung` and the wasm-retry ladder between the two files — the one already double-fixed in #309/#342 — is a separate, larger refactor of the chain; I've left it out of this PR, happy to do it next if you'd like.)

openipc-ai:
Closing in favour of #398, which is the same work opened cleanly for review: a behaviour-preserving commit that extracts the served-channel rule into a shared module (`preview-served.js`), then the #252 feature built on it. #398 supersedes this.

PR #398, opened 2026-09-08
Live adjustments: reflect the served WebRTC channel, via a shared served-channel module

### LINKED ISSUE #252: Live adjustments: reflect the served WebRTC channel and say why, like the Live page does

openipc-ai:
Closes #252. Replaces #397 (opened clean, in two reviewable commits).

## Why

The Live View page follows the channel a WebRTC session actually serves and explains a fallback (#240/#249). The **Live adjustments** preview did not. The obvious way to add it — copy the page's logic into the shared component — would make it a *fourth* place the same rule is kept right, next to the fallback walk and the wasm-retry ladder that already live in both `preview-page.js` and `mj-preview.js`. So this does it the other way round: extract the rule, then build the feature on it.

Two commits, each green on its own:

## 1. `Live: extract the served-channel rule into a shared module`

A **behaviour-preserving** refactor. `preview-served.js` (`window.MajesticServed`) becomes the single home for the rule:

- `decide(info, wanted, shownKey)` — the pure part: mismatch detection, the say-once key, and leaving a standing explanation up when an internal reopen is answered with the adopted channel while the viewer's ask is unmet.
- `make(opts)` — a small applier holding the say-once state, driving a caller's own radios/message through `adopt`/`show`/`hide` callbacks, with the Live page's `auto()` exception (a mismatch in Auto is disclosed by the chip, not by moving the radios).

`preview-page.js` keeps its own state vars, resets, chip and Auto handling, and now words its one decision from `decide()` instead of a second copy.

**How the no-behaviour-change is proven:** `tests/staging.test.js` and `tests/auto-source.test.js` already drive `preview-page.js` and assert the served chip, radios and message — they load the new module and still pass. `tests/served-channel.test.js` covers the module directly (decision + applier + Auto).

## 2. `Live adjustments: reflect the served WebRTC channel and say why (#252)`

The feature, built on the shared module. Hook `onServed` in the shared preview component; on a mismatch move the radios to the served channel (by `.checked`, so the stream-switch does not re-enter and cut the session that reported this) and show a dismissible toast; `chooseStream` untouched. A reply that arrives while its session is still a trial is applied only once that session is on screen. Any non-WebRTC transport clears the message (MSE and the software rung subscribe exactly); a channel change resets it. No chip here (that's the Live page's), and no Auto (this picker has none). Older majestic sends no reply → unchanged, same as #249.

## Verification

Full suite green except the two pre-existing container WebCrypto gaps (`mjcrypto`/`reckeys`, which pass in CI). On a lab hi3516ev300, **both** the Live page (`live.cgi`) and the Live adjustments preview (`camera.cgi?tab=Image`) still play over WebRTC with 0 dropped and no console errors.

## Not in this PR

The pre-existing duplication of `nextRung` and the wasm-retry ladder between the two files (already double-fixed in #309 and #342) is a larger, riskier refactor of the fallback chain; happy to do it separately.


openipc-ai:
Addressed the four review findings (head `83be507`):

**1 — Preview lifecycle went untested.** The held/flush machinery (a served
reply is kept against its staged attachment until that attachment is the one on
screen, then applied once) lived inline in `mj-preview.js`, which builds its own
DOM and so is not loadable by the node harness. Extracted it into
`MajesticServed.holder(applier)` next to the decision and the applier, and
covered it in `tests/served-channel.test.js`: held until its own id goes live,
dropped when a trial fails or the channel is reset, and replaced by a newer
reply. `mj-preview.js` now wires the holder and supplies only its on-screen test
(`swap.isLive`).

**2 — Channel stated without a request.** `decide()` treated `requested: null`
(the parser's mapping of an omitted field) as a match and returned `hide: true`,
while the WebRTC player had already adopted the stated channel — the radios
stayed on the old selection. It now moves the picker to the served channel
silently for a request-less reply: no mismatch can be told and there is no
"instead of X" to word, so any standing message and its key are left as they
are. New test: *a channel stated without a request moves the radios silently*.

**3 — Dead preview kept a stale reason.** The transport ladder could exhaust
after WebRTC reported a fallback, leaving the "showing the other channel" toast
on an empty stage beside the no-preview alert. The exhaustion terminal now calls
`resetServed()`, dropping the state and hiding the toast. (The Live page never
needed this — its fallback reuses the single message slot — but this preview
owns a toast of its own.)

**4 — Private tracker reference.** Removed the private id from the module header
comment; the observable explanation of the fallback stays.

Full node suite green. Both players verified in a real browser against a lab
camera after the refactor — the Live page (1142 frames, 0 element rebuilds) and
the Live adjustments stage (856 frames, 0 rebuilds).

PR #399, opened 2026-09-08
Recordings: a day's listing is one stat and one awk, not five forks a clip

widgetii:
Opening the recordings page took **12.8 s** on a lab hi3516av300 holding 680 clips.

None of it was the network, the camera's daemon, or the browser. The page HTML answers in 111 ms and every other request it makes is under 200 ms. One request is the whole delay:

```
/cgi-bin/j/recordings.cgi?day=2026-09-08   ->  12.85 s   (680 clips, ~19 ms/clip)
```

The page cannot draw its timeline until that returns, so that number is the page load.

## Why

The per-day loop forks about five busybox processes for every clip it reports: `stat`, `basename`, and `json_str`'s own `printf | tr | sed`. Measured on that camera, 680 clips:

| loop body | time |
| --- | --- |
| glob + `[ -f ]` tests only | 0.04 s |
| + `stat` fork | 3.33 s |
| + `basename` fork | 6.34 s |
| + `json_str` pipeline | **11.82 s** |

So 11.8 s of the 12.85 s is process startup — some 3,400 forks to answer one request. `count_clips` in the same file was already written fork-free, with a comment saying why; the per-day loop was not.

It also gets worse on its own. `records.split` is counted in minutes, so a full day is 1440 clips. The camera above was half a day in at 680, and the same page would have taken about 27 s by midnight.

## The change

Stat the whole day at once and let awk write the JSON. `xargs` does the batching, so the argument list stays inside `ARG_MAX` however many clips a day holds, and it preserves input order — the glob's, already chronological because these names are the clock. awk sees one uninterrupted stream, so the commas never have to work out where a batch ended.

Measured after, same camera: **0.29 s** for 460 clips and **0.35 s** for 688. Full page-open sequence went from ~13.5 s to 0.82 s.

## Two things that quietly cost a listing

Both were found by diffing the old implementation against the new one over a directory of adversarial names, not by reading:

- busybox calls a zero-byte file a **`regular empty file`**, not `regular file`. A type test for the exact string silently drops the clip being recorded right now — the newest one, on the day you are most likely looking at.
- a backslash in a `gsub` **replacement** is reinterpreted by `gsub`, and getting that wrong is silent. The escaping is done a character at a time so it never puts one there.

## Verification

Old and new run back to back over the same directory, on the camera. Byte-identical for names containing quotes, backslashes, pipes, spaces, `$`, `*`, `'`, tabs (control characters stripped, as before) and UTF-8; a directory named `*.mp4` still excluded; a dangling symlink still excluded; a zero-byte clip still listed; an empty day still `[]`.

One deliberate difference: a symlinked clip now reports the size and mtime of the clip rather than of the link. `-L` is what reproduces the `[ -f "$f" ]` test this replaces — a link to a clip is a clip, and a dangling one fails stat and drops out — and the target's numbers are what the caller was always asking for.

Also checked that the shipped form survives packaging: comment-only and blank lines stripped out of the file still parse and still return the identical listing on the camera.

`npm test` and `tools/lint-templates.sh` both pass.

### DIFF
--- a/www/cgi-bin/j/recordings.cgi
+++ b/www/cgi-bin/j/recordings.cgi
@@ -114,13 +114,61 @@ if [ ! -d "$dir" ]; then
 fi
 
 printf '{"path":"%s","clips":[' "$(json_str "$dir")"
-first=1
-for f in "$dir"/*.mp4; do
-	[ -f "$f" ] || continue
-	info=$(stat -c '%s|%Y' "$f" 2>/dev/null) || continue
-	sz=${info%%|*}; mt=${info#*|}
-	[ "$first" = 1 ] || printf ','
-	first=0
-	printf '{"name":"%s","size":%s,"mtime":%s}' "$(json_str "$(basename "$f")")" "$sz" "$mt"
-done
+
+# One stat and one awk for the whole day, not five processes per clip.
+#
+# The obvious loop -- stat the file, basename it, json_str the name -- forks
+# `stat`, `basename`, and json_str's own printf|tr|sed, for every clip.
+# records.split is counted in minutes, so a full day is 1440 clips, and that
+# loop is some seven thousand busybox forks to answer one request. Measured on
+# an armv7 camera: 11.8 s for 680 clips, all of it process startup -- the glob
+# and the tests together cost 0.04 s. The page cannot draw its timeline until
+# this request finishes, so that number is the page load. The pipeline below
+# emits the same bytes in 0.07 s.
+#
+# xargs does the batching, so the argument list stays inside ARG_MAX no matter
+# how many clips a day holds, and it hands them on in the order it got them --
+# the glob's, which is already chronological because these names are the clock.
+# awk sees one uninterrupted stream, so the commas never have to work out where
+# one batch ended and the next began.
+#
+# -L asks about what a symlink points at, which is what the `-f` test this
+# replaces asked: a link to a clip is a clip, and a dangling one fails stat and
+# drops out, as it did before. Size and mtime now describe the clip rather than
+# the link, which is the answer the caller was always after.
+set -- "$dir"/*.mp4
+if [ -e "$1" ]; then
+	printf '%s\0' "$@" |
+	xargs -0 -r stat -L -c '%s|%Y|%F|%n' 2>/dev/null |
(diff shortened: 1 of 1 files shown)


widgetii:
Both findings were real. Fixed in 67805bc.

**1. One broken link empties the day** — correct, and worse than the report suggests. The `[ -e "$1" ]` was only ever meant to spot an unmatched glob, but it gates the whole pipeline, so any first entry that fails it returns `[]` for a day that is full of footage. The trigger is not just a dangling symlink: this directory is pruned while it is being read, and I watched eleven clips disappear from a day during one browsing session on the lab camera — the first clip vanishing between the glob and the test is ordinary. On a vfat card, which is what these record to, the prune race is in fact the *only* way to hit it, since vfat has no symlinks at all.

The guard is gone. Every entry is rejected on its own, as it was before the change. The unmatched glob needs no guard either, exactly as you say: it stays literal, `stat` says no such file, and the day comes back as the `[]` it should be.

Verified on the camera with the deployed script: a dangling symlink sorting first now lists every other clip, a clip deleted between the glob and the stat no longer truncates the day, and an empty day is still `[]`.

**2. Newline names corrupt the clip list** — also fixed, though it is worth saying it was not a regression: the old code squashed the newline out with `tr -d`, which produced `ab.mp4` for a file called `a\nb.mp4` — a name no file on the card answers to either. Both spellings were wrong, so I have made the record self-delimiting rather than pick between them.

The format now ends with a terminator printed *after* the name (`'%s|%Y|%F|%n|#'`), and awk requires `$NF == "#"`. A name split across two lines does not carry it and is dropped, so such a clip is omitted rather than reported under a truncated name. A filename that itself contains `|#` still round-trips correctly, because the terminator is the *last* field and the name is everything between the type and it — that case is in the test set.

Re-verified against the old implementation over a directory of adversarial names (quotes, backslashes, pipes, spaces, `$`, `*`, tabs, UTF-8, `|#`, a newline, a directory named `*.mp4`, a symlink, a dangling symlink, a zero-byte clip, an empty day). Byte-identical except the two documented differences: a symlinked clip reports the clip's size and mtime rather than the link's, and a newline name is now omitted instead of mangled.

735 clips in 0.33 s, so neither fix cost anything.

PR #409, opened 2026-09-08
Fail the build on `local` outside a function

widgetii:
The trap #407 fell into, made impossible to repeat.

## Why nothing caught it

`local` is legal in a function and a **runtime** error at top level: ash prints
`local: not in a function` and **aborts the script there**. So the page does not
misbehave, it stops — everything after that line is simply never emitted.

`sh -n` is a *syntax* check and this is valid syntax. `tools/lint-templates.sh`
runs `haserl -d` and then `sh -n` over every template, and accepts it for
exactly the same reason. #407 hit it moving a partial's shell out of a function:
Live came back **9,780 bytes instead of 31,952**, with the whole player missing
and one line on a stderr nobody reads.

What found it was rendering the page against the previous tree and comparing the
output — which is not something CI can do, because it needs a camera. So the
check has to be static.

## The rule

`local` needs a function open. Scope is brace groups, counting only a group
opened by a `name()` definition — a bare `{ }` group at top level is not a
function, and `local` in one is the same error.

It walks the same parser `strip-cgi-comments.js` strips with, which now exports
`shellLines()`: the lines of a `.cgi` that are *actually shell*, skipping
literal markup, heredoc bodies, and the continuation of a string opened on an
earlier line. The two checks ask different questions of a template and must not
disagree about which bytes are shell — that disagreement is how a comment full
of CSS got handed to `/bin/sh` in the first place.

A file with no code block is a plain shell script, so `sbin/` and `bin/` are
covered too: **85 files, 6,387 lines of shell.**

## Deliberately narrow

Like the gates in `pr_compliance_checklist.yaml`. It looks for `local` in
command position at the start of a line, which is how all 93 of the real ones
in this tree are written. A `x=1; local y` would be missed. That is worth it for
a check with no false positives, and the ones it does catch are the ones people
actually write.

## One shape needed real parsing, and the tree already had it

`${inet#* }` ends in "space brace", so split on whitespace it hands back a bare
`}` that closes whatever function is open. On its first run this check accused
`network.cgi:74` of a `local` that is properly inside `net_read()`. Parameter
expansions are consumed whole now, and both trim forms are regression cases.

## Verified

- Passes over the tree as it stands.
- **Reintroduce the exact #407 bug** — `local mj_src_labels` at the top of
  `p/player.cgi` — and it fails, naming the file and line.
- Unit cases for what must be caught (top level, after a function closes, a
  bare brace group, a plain shell script) and what must not (inside a function,
  inside an indented one, nested, in a comment, in literal markup, in a heredoc
  body, a brace inside a quoted `awk` program, `${v}`, the word in a string).
- `tools/lint-templates.sh` green; dist build and comment gate green (133
  assets).


widgetii:
All three fixed in 1f9dcc6, and all three were real. None is a shape this tree contains today, which is the point — a gate is judged by what it does to code nobody has written yet.

**1. Shell punctuation needs no whitespace around it.** `bareTokens` split on whitespace alone, so `f(){` was one word matching no function pattern with its brace hidden inside, and `};` never equalled `}`. Exactly as described: a compact function's `local` was rejected, and a top-level `local` after `};` was accepted — the check defeated without going near the semicolon form it documents as out of scope. Punctuation is its own token now, and a definition is the token sequence `name ( )` however it is spaced.

**2. Plain scripts bypassed heredoc handling — and this one is live.** `shellLines` returned a script line for line. `sbin/updatewebui`'s usage text *is* a heredoc, so 60 lines of English prose were being offered to the scanner as shell. It passes today only because none of them happens to begin with a keyword or hold a brace, which is luck rather than correctness. The state a code block needs is exactly the state a script needs, so both go through one walker now; 60 fewer lines are read as code.

**3. Only the first heredoc per line was tracked.** `cat <<A <<B` is one command with two bodies, one after the other, and B's was left looking like shell. All of them are reported now, in the order the shell consumes them, and both readers keep a queue. Worth noting this one was not confined to the lint — it fixes the **shipped stripper**, which would have stripped a comment out of B's body as though that body were code.

Verified the parser change does not move what the stripper emits: running the previous and current versions over the tree gives byte-identical output, since there is no multi-heredoc command here. That is why all three were latent rather than broken.

Regression cases on both sides for each shape: compact and spaced definitions, `};`, a subshell and a command substitution that are *not* definitions, a plain script's heredoc body, two heredocs on one command, and a quoted one. The acceptance test still holds — reintroduce #407's `local mj_src_labels` at the top of `p/player.cgi` and the check names the file and line.

PR #410, opened 2026-09-08
Storage: a card that has stopped recording says so on every page

widgetii:
An SD card that dies is the one fault on this camera with no symptom. The picture stays live, every counter stays healthy, ONVIF goes on answering, and the archive simply stops growing — so the only thing between an owner and a week of missing footage is a sentence in the interface.

Measured on a lab hi3516av300 with the card physically off the bus:

| Where you'd be looking | What it said |
|---|---|
| Recordings | *"There is no SD card in the camera — nothing is being recorded."* ✅ |
| SD card | correct ✅ |
| Dashboard | *"No SD card detected"* — grey, x-small, inside a panel |
| Live | nothing |
| Every other page | nothing |

Finding out required already suspecting it. Somebody with the Live page open — the page people leave open — was told nothing.

## The Dashboard was worse than silent

Its storage row is built from `df`, and a card the kernel has dropped to read-only still reports the free space it had that morning — so the row drew a **green "SD" badge with capacity** over a card that had recorded nothing since lunchtime. That is the ordinary way an SD card dies.

`df` is not wrong; it is answering a different question. `j/sdcard.cgi`'s `health` verdict exists precisely because none of these failures is visible from the filesystem, and the badge takes that verdict now. Screenshot in the thread: the row still says `/mnt/mmcblk0p1 — 0 / 64.0M (0%)` and the badge beside it is red, with the reason.

## What this adds

A banner on every page, in the header where the other four live, filled client-side so it appears **while it is true** rather than on the next navigation — a card dies while somebody is looking at a page, not between two of them.

The wording is **not** the banner's. `storage-verdict.js` now holds all nine sentences and both callers take them: the Recordings page the long form it already showed, the banner a short one. Two vocabularies for one fault is how a camera comes to describe its card one way in a banner and another way on the page the banner points at — so the action is that same page, in the same words it already used.

Only while `records.enabled`, checked before anything is polled: a camera not recording to a card has no business being nagged about one on every page, and there it costs nothing at all.

**One request added, and only one.** `records_state` and the drop counter are already in the top-level `/metrics` the heartbeat fetches every two seconds, so the recorder half rides that. Only the card half needs `j/sdcard.cgi`, at 30 s — measured at 10 ms on the lab camera.

## Tests

`tests/storage-verdict.test.js` pins both ways this goes silently wrong.

The verdict half walks every card state and every recorder state and fails if one goes quiet — including the trap this was measured on: **a camera with no card reports `records_state 0`, "ok"**, because the recorder never started and the state never moved off its initial value. Anything reading that half alone is looking straight at a dead camera and seeing health.

The banner half runs `storage-check.js` against a stubbed page and fails if the banner never appears: a wrong gate on `records.enabled`, a heartbeat sample it cannot read, or an older majestic with no recorder metrics at all, where the card half is the whole answer and must not be lost.

The three vm harnesses that assemble a Recordings page by hand load the new module too — `recordings.js` takes it at module scope the way it takes the index and the crypto, and `p/header.cgi` is what loads it in a browser. A harness without it builds a page that throws inside `cardWritable()`, which is what four failures in `recordings-pump.test.js` said, correctly, before that line existed.

Whole suite green (2002 ok, 0 fail), template lint clean, and both screenshots are the branch running on the lab camera with its card actually gone.

## Not covered

`sdcard.cgi` keys off the block device, so a tmpfs cannot stand in for a `readonly` card and that verdict is not exercised on real hardware here — only through the test and through the same df-says-storage / health-says-none contradiction the screenshot shows.

And a banner only reaches somebody who has the interface open. Pushing this to a phone through the existing ntfy/telegram extensions is the piece that would actually mean "immediately", and it is deliberately not in this PR.


widgetii:
All four addressed, and the branch is rebased onto master (the conflict was the test list in `package.json` — #408 and #409 added theirs beside this one).

They were all the same fault in different places: **a positive statement made where nothing had established it** — which is the rule this branch exists to enforce, so finding it broken in my own code four times is fair.

**1. The badge had two colours and four states to draw.** Everything that was not `danger` came out green: the moment before the card endpoint answers, a fetch that failed, and both verdicts that say footage is being *lost*. It has four now — danger, warning, neutral while nothing is known, and green only where `writable()` says both halves agree. The server-rendered badge starts neutral rather than green, because `df` alone establishes nothing.

**2. A failed poll kept the last good answer** — and the comment beside it claimed this was "exactly as the Recordings page does", which is the opposite of true: `recordings.js` clears `state.card` on a failed fetch for precisely this reason. A badge could stay green through an hour of refused polls. The reading is dropped now.

That fix was incomplete on its first pass and the new test caught it: with the card unknown there is no longer a mountpoint to say *which* row to take back to neutral, so the stale green survived anyway. The mountpoint is now remembered across a reading going away — it is a property of the camera, not a claim about the card's health, and it only ever withdraws a claim.

**3. A configuration that could not be read was being read as "recording is off".** `mjConfig()` resolves `{}` on failure and deliberately does not cache it so a later call retries — but nothing here made a later call, so one refused request at page load silently disabled every storage alert for the life of the page. `/api/v1/config.json` reports keys at their defaults (checked on the camera), so `records.enabled` is a real boolean whenever the config was read at all and `undefined` means only that it was not: unknown retries, `false` is answered once and left.

**4. Repeated ids inside the row loop.** The Dashboard draws one row per filesystem `df` matched while this endpoint describes exactly one device, so the fixed id coloured whichever row came first. Rows carry their mount now. Measured on the lab camera with two tmpfs mounts standing in:

```
/mnt/mmcblk0p1       badge flex-shrink-0 mj-sd-badge text-bg-danger
/mnt/usbdrive        badge text-bg-secondary flex-shrink-0 mj-sd-badge
```

Each is pinned by a test: the badge's four colours, a second filesystem left alone while the card's row is marked, a poll that fails taking its claim back, and a config that failed being retried rather than believed. Whole suite 2113 ok, 0 fail; template lint clean.

PR #412, opened 2026-09-08
A page writes nothing to stderr

widgetii:
Asked whether the `/bin/sh: not found` lines were really gone, I measured
instead of assuming, and they were not — just not on the Live page any more.

Rendering all 47 pages on a camera left **434 bytes on majestic's stderr**, from
three places. None of it is visible under the service, because that stderr goes
nowhere: all of it is only findable by running the daemon in the foreground,
which is exactly how the six lines that started this were found.

This takes that to **zero**.

## Three sources

**`field_textedit` ran `cat "$3"` on a path that need not exist.** `editor.cgi`
passes `"$editor_file"`, which is empty until a file is chosen, so opening the
editor from the menu printed `cat: can't open ''` on every visit. The textarea
was empty either way — which is why nothing ever looked wrong. The served bytes
are identical; only the noise is gone.

**`ex` escaped stdout and left stderr to the daemon.** Its whole job is to show
what a command said, on a panel called *"Advanced — raw configuration"*. On a
camera with no tunnel, `ip link show wg0` therefore gave the operator an **empty
box** while the reason went somewhere nobody reads. It shows the reason now,
which is both the fix and the better page.

**`p/header.cgi` is a partial, and majestic execs it for anyone who asks for
`/cgi-bin/p/header.cgi`.** haserl then runs it with none of `p/common.cgi`'s
helpers defined: eleven `/bin/sh: <fn>: not found` lines and a fragment with an
empty `<title>`, from a URL nothing links to and nobody can use. Every real page
includes `common.cgi` first, so the helpers are the test — no `esc`, no page. It
answers 404 now instead of 200-with-a-broken-fragment, which also makes it
consistent with `p/motor.cgi` and `p/pages.cgi`, whose direct requests already
failed.

## Two things about that guard are measured, not assumed

The status line has to be an `HTTP/1.1` one, the way `j/fw-latest.cgi` writes
it. A CGI `Status:` header arrives as a **response header of that name** and the
reply is a 500 — that was my first attempt, and the camera said so.

And the guard lives inside the existing code block rather than one of its own,
because what follows that block's closing tag is the response's own headers, and
haserl emits the newline that follows a closing tag. A block above them puts a
blank line in front of the status line.

## Verified on an ssc30kq + imx335

- 47 pages, **434 bytes of stderr before, 0 after**, no `not found` anywhere.
- Output unchanged except where it should be: `p/header.cgi` 9,352 → 78 bytes,
  `wireguard.cgi` +56 for the error text it now shows, **`editor.cgi`
  byte-identical**, and `network.cgi`'s two bytes of live `ifconfig` counters.
- Over HTTP through majestic: every real page still 200, the partial a clean
  404 with `Not a page.`
- `tools/lint-templates.sh`, the dist build and its comment gate (133 assets)
  and the full suite all green.

## What I got wrong before

Earlier verification runs recorded `stderr A/B = 434 / 434` and I treated
*equality* as the result. That was sound for what it was doing — proving a
refactor changed nothing — but I then spoke as though stderr were clean. Equal
is not empty, and I never opened those 434 bytes until asked.

### DIFF
--- a/www/cgi-bin/editor.cgi
+++ b/www/cgi-bin/editor.cgi
@@ -36,12 +36,16 @@ else
 	if [ ! -f "$editor_file" ]; then
 		log_create "danger" "File not found!"
 	elif [ -n "$editor_file" ]; then
-		if [ "b" = "$( (cat -v "$editor_file" | grep -q "\^@") && echo "b" )" ]; then
+		# 2>/dev/null: the path is whatever GET_f said, and a regular file can
+		# pass -f and still fail to read (/proc/self/mem). The probe's answer
+		# is "not a text file" either way, so the read error has nothing to add
+		# except a line on a stderr the service sends nowhere.
+		if [ "b" = "$( (cat -v "$editor_file" 2>/dev/null | grep -q "\^@") && echo "b" )" ]; then
 			log_create "danger" "Not a text file!"
-		elif [ "$(stat -c%s $editor_file)" -gt "102400" ]; then
+		elif [ "$(stat -c%s "$editor_file" 2>/dev/null)" -gt "102400" ]; then
 			log_create "danger" "Uploaded file is too large!"
 		else
-			editor_text="$(cat $editor_file)"
+			editor_text="$(cat "$editor_file" 2>/dev/null)"
 		fi
 	fi
 fi
--- a/www/cgi-bin/p/common.cgi
+++ b/www/cgi-bin/p/common.cgi
@@ -115,9 +115,10 @@ ex() {
 	echo "<div class=\"${2:-ex}\"><h6># ${1}</h6><pre class=\"small\">"
 	# 2>&1 because this block exists to SHOW what a command said, and what a
 	# failing one says is on stderr. Without it the operator gets an empty box
-	# and the reason goes to majestic's stderr, which under the service goes
-	# nowhere: `ip link show wg0` on a camera with no tunnel printed "can't find
-	# device" there and nothing here, on a panel whose whole job is to report.
+	# while the reason goes to a stderr the service sends nowhere -- on a panel
+	# whose whole job is to report. wireguard.cgi's `ip link show wg0` is the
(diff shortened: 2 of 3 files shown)


widgetii:
Both addressed in 1a9a4ab (now 64f4668 after a rebase onto master).

**1. `[ -r "$3" ]` is true of a directory — a real bug in the fix.** Confirmed on hardware before touching it: `?f=/tmp` reached the helper, passed the readability test, and put a read error back on the daemon's stderr — recreating through a query parameter exactly what this change exists to remove. `field_textedit` now tests for a regular file.

Chasing that with hostile paths found **two more reads on the same page**, both reachable the same way and neither exercised when the only path I had tried was an empty one: `editor.cgi`'s own binary-file probe, and the read that fills the textarea. A regular file can pass `-f` and `-r` and still fail to read — `/proc/self/mem` is the reachable example — so the reads are redirected as well as guarded. The guard states what the helper accepts; the redirect covers what a predicate cannot. Nothing is concealed by it: the operator's answer is the same empty textarea, and `editor.cgi` writes the banner that explains why.

The textarea read and the `stat` beside it were also **unquoted**, so a filename containing a space fed `cat` several arguments and failed. Quoted now, which removes a third source of the same noise and makes such a file open instead of erroring.

Tested with what a query string can actually carry — empty, a directory, a regular file, an absent path, `/dev/null`, `/dev/zero`, `/proc/kcore`, `/proc/self/mem`, and a filename with a space: **every one zero bytes of stderr**, alongside the 47-page sweep.

**2. Comments quoting captured error text.** My reading is that those strings were generic tool output naming no device, and that the symptom in the description came from the reporter's camera rather than lab infrastructure — but the rule is written binary, and the comments lose nothing by describing the failure instead of pasting it. Done. The commands stay: `ip link show wg0` is this tree's own source, not a capture.

Re-verified after the rebase: 47 pages at zero stderr, `lint-templates.sh`, the dist build with its comment gate (135 assets) and the full suite all green.

Issue #348, opened 2026-09-05
New firmware available notification is shown when no new firmware is available


usa-:
The notification about a new firmware version is displayed on all pages, even when there is no new firmware available.

On the **System → Update** page, it says that there is no new firmware, while the notification still says that a new version is available. This is confusing.

<img width="666" height="687" alt="Image" src="https://github.com/user-attachments/assets/29dd13f8-eb43-4560-9b24-ff3cb652b37a" />

widgetii:
Fixed — and thank you for pairing it with the Update page, which is what made the cause obvious.

Two things were wrong. The list the notice compares against had an entry for a build that was never actually released, so cameras running the newest available build were told they were behind something that did not exist. More fundamentally, the notice looked only at the camera software, while the button under it flashes a whole firmware image — and those two do not become available at the same moment. So it could say you were behind while the Firmware page correctly offered you nothing to install.

The notice now asks the camera whether an image it can actually install exists, through the same query the Firmware page makes, so the two cannot disagree. It stays silent unless there is something to install — including when the updater cannot be reached, because not knowing is not a reason to nag.

Nothing to do on your camera. If the notice appears again, there really is an image for your board, and the Firmware page will offer you the same one.

PR #419, opened 2026-09-09
Update: one decision at the top, and the banner it answers stops repeating

widgetii:
The Firmware page had four faults and they were one fault: nothing on it said which of the things on offer was the thing to do.

A full-width card holding four short rows of `dl`, a 32rem card of switches floating beside 870px of nothing, and then **two equal cards each with a filled button at the very bottom** — with the site-wide "N builds behind" banner still above all of it, after its own button was what brought you here. Measured on a 13" MacBook Air, Install was a small button at y=902 of a 956px viewport, beside a wider `Upload & install` that took the eye first.

### The hero

What is on offer, what it is worth, and the only filled button on the page, first thing, at y=237. A card rather than a banner — it borrows `.mj-notice`'s severity rule and 28px mark, because a banner is a thing you close and this is the page.

Below it two columns of real content: the changes on the left, the camera's facts and the write switches in a rail on the right. **From a file** drops to the last card in that rail, one line and a way in. It is not the second half of a pair of equal choices — almost nobody arrives here with a `.tgz` — and presenting it as one is what made the download route hard to find. It grows into the wide column in the one state where it *is* the route: no connection.

### The banner stops repeating

`p/header.cgi` renders no `update-notice` slot on this page. `update-check.js` needs no guard for it; a missing slot was already its "say nothing" path.

`update.cgi` says what that banner said, as its own subject: the counts beside Install, the changes in a card of their own, opened rather than collapsed, with the notes that carry this camera's vendor marked as such. The counting moved to `www/a/fw-changes.js` so both surfaces read one copy of it — two answers to "how far behind is this camera" that could disagree on the same day is #348 arriving by a different road.

Group headings deliberately carry no count. The list is routinely shorter than the totals in the pills, because a change whose sentence was withheld is still counted and simply not described, and a "3" beside a group while the pills say five is how a page points at a gap it is not supposed to discuss.

### Three states, decided server-side

`available` / `current` / `offline`, using `j/fw-latest.cgi`'s own prefix comparison in both directions so this page and that endpoint cannot contradict each other. `sysupgrade --list-builds` is still asked live on every draw, never from that endpoint's cache: the notice on every other page can afford a six-hour-old answer, and this page — where somebody has come to act — cannot.

Where the changelog is absent, which is a browser that cannot reach the public feed as well as a camera that is current, a `:has()` rule widens the rail and deals its cards across it, rather than leaving the empty two-thirds this layout exists to remove. A browser without `:has()` keeps the narrow rail, which is no worse than the page was before.

### While it writes

There is nothing else to do on this page, so there is nothing else on it: `#fw-controls` goes and the phase strip, the log and the caution take its place. The log used to open *below* two install cards that stayed live under a flash already under way.

Every marker the strip reads is one `update.js` already had to recognise for another reason, so it can only ever say what the transcript already said. The bar shows only a percentage `sysupgrade` itself printed, and hides again once the download is behind it — a bar that invents its own progress is worse than none, because it is believed.

### Tests

`tests/update-check.test.js` passes unchanged through the extraction; it now loads both files into one context, in `p/header.cgi`'s load order.

`tests/update-page.test.js` is new. The card fails the way the banner fails — a card that never unhides looks exactly like a camera whose browser cannot reach the feed, which is a state this page is *required* to render as nothing, so "nothing appeared" is both the bug and the correct answer. It pins the counts, the grouping, that no group heading carries a number, every silence, and that this page's own grouping does not put back a note `fw-changes.js` dropped for belonging to another vendor.

The phase strip is deliberately not pinned: it fails in front of somebody watching it, with `sysupgrade`'s own transcript underneath saying what actually happened.

### Verified

On an hi3516ev200, against the live feed, in all three states and at 390px. `npm test` and `tools/lint-templates.sh` pass; `www/a/bootstrap.min.css` regenerated.


widgetii:
Addressed the review — four of the six were real, and two of those were the same mistake on opposite sides of the same comparison.

**1. False network diagnosis** — fixed. `fw_state` collapsed every empty `latest_build()` into `offline` and the hero then asserted no route to the internet. The reason is its own variable now: `noroute` only when there is genuinely no default route (provable), `nocheck` for everything else — "Couldn't check for new firmware", naming no cause, and no "Check network settings" button pointing at a subsystem that may be fine.

**3. Available invented from an unreadable revision** — fixed, same way. The image is still offered (it remains installable, and that is what this page has always done), but `fw_cmp=unknown` presents it as *Build available* / "Firmware from … is available" rather than as an update, and says the camera reports no revision to compare. "X is ready" is a claim about two builds; with one unreadable there is nothing to claim.

Severity now follows from those rather than from the state — `.mj-sev-warn/-ok/-info`, named for what they mean, since an image that cannot be compared is not a warning and a check that did not happen is not a fault.

**4. Current cameras cannot reflash** — fixed, and a good catch. The `current` state now carries a **Reinstall** button, and it implies the force its own word promises: without it sysupgrade answers "same version, nothing to update" and writes nothing, which is a button that does nothing by another road. Its confirmation is asked in `update.js`'s own click handler, not through `main.js`'s `.confirm` hook — that one wires on `load` while this listener is registered during parsing, so it would have run first and asked *after* the flash began.

**6. Malformed feed revisions** — fixed. Feed revisions must now match `/^[0-9a-f]{7,40}$/`, the same shape `parseBuild` demands of the camera's own, so the two ends of the prefix match agree on what a revision is. Five malformed-revision cases added to `tests/update-page.test.js`. (Pre-existing rather than introduced here — the function moved verbatim — but the reviewer is right that remote input should be validated.)

**2 and 5. Disclosure** — fixed. The source comment no longer quotes a real `GITHUB_VERSION`, and the test fixture is synthetic and no longer worded as a camera under test. Same coverage.

Verified on an hi3516ev200: all five hero variants and both buttons.

PR #423, opened 2026-09-10
update: reattach to an upgrade already in progress on load

widgetii:
## Why

The Firmware page opened `/ws/upgrade` only when the user pressed **Install**, and the camera keeps a single upgrade session. So a browser reloaded mid-flash — or a second device — landed on the ordinary Update card with no sign that the flash was under way, and the running session answered a second connection with **503**.

That is the road to a camera powered off mid-write. Reported on a **gk7205v200 (Tiandy TC-C321N)**: the owner watched the download bar freeze at 83.7 %, refreshed, found the WebUI and SSH both gone (they are, for the ~50 s the kernel+rootfs flash takes) and pulled the plug.

The camera side now attaches a returning `/ws/upgrade` client to the running upgrade, replays the transcript, and leads it with a *"do not power off"* banner (landed separately). **This PR depends on that behaviour** — against an older firmware the reattach open is refused with 503, which this handles by keeping the warning and watching for the reboot.

## What

- **`update.cgi`** sets `data-active` on a new, always-present `#fw-inflight` element from the camera's upgrade-in-progress marker file (`/tmp/majestic-upgrade-owner`, kept for the life of an upgrade). While it is set the network-touching build check (`sysupgrade --list-builds`) is skipped: the page goes straight to the progress view, so there is nothing to offer, and nothing to add load to a camera mid-flash.

- **`update.js`**, on load, reads the flag and *attaches*: shows the progress view and opens `/ws/upgrade` **without sending a start frame** (sending one is what begins an upgrade, and one is already running). The socket open, log follow, reboot watch and error handling are now shared with the Install path through a `connectUpgrade()` helper.

- A run that attached rather than started is marked `reattached`, so `confirmUpgrade` does not judge the version the camera comes back on as pass/fail — it never saw whether the run was an update, a reinstall or a reset — and reports what is running instead. If the socket opens but nothing arrives (the upgrade ended between render and connect), the page reloads to a fresh render. If the socket is **refused** (an older single-session firmware answering 503), the "do not power off" warning is kept and the page watches for the reboot instead of reporting "could not start".

## Hardware tested on

- **gk7205v200** lab board (Goke, 8 MB NOR) — the *before* behaviour this fixes: with a `/ws/upgrade` upgrade in progress, a second connection returns `HTTP/1.1 503 Upgrade already in progress`. That is the dead-end a reload hit.

## Evidence

`npm test` — full suite green, including a new `tests/update-reattach.test.js`:

```
a page loaded mid-flash attaches to the running upgrade
  ok   the socket to /ws/upgrade is opened at load
  ok   the progress view replaces the controls
  ok   the headline names no build it never saw
  ok   and the status warns not to power off
  ok   opening the socket sends no start frame
  ok   the streamed banner reaches the transcript
an older firmware that refuses the second connection keeps the warning
  ok   a failed attach does not report "could not start"
  ok   and the do-not-power-off warning is kept
a page loaded when nothing is flashing does not attach
  ok   no socket is opened when the flag is unset
```

`tools/lint-templates.sh` — `ok: 56 haserl templates, 25 shell scripts`.


widgetii:
Addressed the review in 1bede11:

- **Older cameras lose the safety warning** (correctness): `connectUpgrade` now takes an `onError` override. On the attach path a refused open (an older single-session firmware answering 503) keeps the "do not power off" warning and watches for the reboot, instead of overwriting it with "could not start". New test case *"an older firmware that refuses the second connection keeps the warning"* pins it.
- **Comments cite a pull request** (rule 1): removed the companion-PR references from all five comments; the self-contained behavioural explanations stay.
- **A comment exposes daemon source details** (rule 8): reworded `update.cgi` to the observable contract — the camera keeps `/tmp/majestic-upgrade-owner` present for the life of an upgrade — with no source-file or internal-symbol reference.

Full `npm test` green and `lint-templates.sh` clean.

PR #424, opened 2026-09-10
Day/Night: say what the lamp-down check found

widgetii:
A camera whose night illuminator is the only light in the room reads
under the day threshold on every tick — the lamp is most of what the
gain is measuring. majestic answers that by dropping the lamp for three
seconds before it believes a day, and refusing when the scene turns out
to be night after all. That refusal is CORRECT, and this page could not
show it: the chart went on plotting a gain under the day mark, the
countdown went on promising a switch, and neither ever arrived. It reads
as a switch that is broken. It is what happened in #370, and the answer
had to be given by hand — go to Logs, filter for `lamp down`, read the
daemon's own sentence.

majestic publishes the verdict now, so the panel can say it:

    Night. Watching the sensor gain; day comes when the gain settles
    back down. Lamp at 100%. The last check dropped the lamp 2 min ago
    and found 16x — the light in view is the camera's own, so night
    holds; it looks again in 5 min.

The number is the whole of what a refusal has to say to an operator: it
is what the scene needs without the camera's own light, so it is both
the reason day was refused and the value "Gain multiple that means day
again" would have to reach for that room to count as day.

Three smaller things follow from having the verdict:

- While a refusal stands, a day countdown says "checking again in 40 s"
  rather than "switching in 40 s". What happens at the end of the dwell
  is another lamp-down check; whether it ends the night is the scene's
  business. The night countdown is untouched — night needs no permission
  from the check.
- A standing refusal silences the "the exposure is still at its ceiling"
  clause. Both explain a held night; the refusal explains it in terms of
  what is in view, which is the one the operator can act on.
- The legacy threshold pair gets the same sentence. It goes through the
  same check, and it is the door #370 was reported from.

Absent gauges say nothing at all. A camera with no lamp never runs a
check, and an older majestic publishes no verdict; neither is a refusal.
A confirmation is spoken only while it is FRESH and only while it can be
dated — it explains a switch that already happened, and by mid-afternoon
"the last check found daylight" is an old sentence holding the line, so
a verdict arriving without its age cannot be known to be worth saying. A
refusal survives the same gap, because it explains the night the camera
is in now whenever it was reached.

These two readings are used as the camera gives them, and deliberately
not read forward the way the countdown beside them is. That projection
measures time since the MONITOR's gauges last changed — the streak moves
only on the camera's own tick, slower than the poll, so it has to be
projected — and on a settled night those gauges do not change for
minutes at a stretch. The probe's age and wait are recomputed on every
scrape and are already current; adding that number to them would age a
check by however long the monitor happened to sit still and floor "looks
again in" to nothing while a stand-off was still running. Both phrases
are coarse enough that a poll period does not show.

Tests: sixteen checks in tests/ircut-check.test.js — the refusal and its
three parts, the countdown's wording in each direction, the ceiling
clause silenced and otherwise not, a camera that has never run a check,
a confirmation fresh and stale, an unreadable check, a refusal whose
gain the camera could not state, a verdict with no age at all on both
doors, a settled monitor not ageing the check with it, and the legacy
threshold pair.

### DIFF
--- a/tests/ircut-check.test.js
+++ b/tests/ircut-check.test.js
@@ -560,6 +560,136 @@ function runRest() {
 				!/day comes when/.test(unknown), unknown);
 	}
 
+	group('monitorView: the lamp-down check has a verdict, and it is said');
+	{
+		// The state the whole day probe exists for, and the one the page could
+		// not show: a room lit by the camera's own illuminator reads under the
+		// day threshold on every tick, the camera checks with the lamp down
+		// and correctly refuses, and every gauge on this page goes on
+		// promising a day that never comes (#370). The verdict is a gauge now,
+		// so the sentence can say what happened instead of leaving the
+		// countdown to be believed.
+		const lit = (extra) => ic.monitorView({ lightMonitor: true },
+			Object.assign({ night_mode_source: 4, night_enabled: 1,
+				night_auto_gain_milli: 1620 }, extra)).line;
+
+		const refused = lit({ night_probe_verdict: 0,
+			night_probe_gain_milli: 15500, night_probe_age_seconds: 120,
+			night_probe_wait_seconds: 300 });
+		check('a standing refusal is said, with what the scene needed',
+			/dropped the lamp 2 min ago and found 16x/.test(refused), refused);
+		check('...and named as the camera\'s own light',
+			/the light in view is the camera's own, so night holds/.test(refused),
+			refused);
+		check('...and says when it will look again',
+			/it looks again in 5 min/.test(refused), refused);
+
+		// A countdown under a standing refusal is a promise already broken
+		// once. The number is honest; what happens at the end of it is a
+		// check, not a switch.
+		const counting = lit({ night_auto_pending: 2,
+			night_auto_dwell_seconds: 60, night_auto_streak_seconds: 20,
+			night_probe_verdict: 0, night_probe_gain_milli: 15500,
+			night_probe_age_seconds: 30, night_probe_wait_seconds: 0 });
+		check('the countdown says what the camera will actually do',
+			/checking again in 40 s/.test(counting) &&
+				!/switching in/.test(counting), counting);
+
+		// Two explanations for one held night, and the refusal is the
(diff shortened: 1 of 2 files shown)


widgetii:
Both findings were right, and both are fixed in `31a2c3c`.

**A confirmed day was spoken without a date.** `probeNote` built the
sentence from `p.age` and left the time phrase out when the camera gave
no age — so a confirmation with no age printed "The last check dropped
the lamp and found 3.2x: real daylight", which is exactly the sentence
the freshness rule exists to prevent. A confirmation only earns its
place while it is recent, and a verdict that cannot be dated cannot be
known to be recent. It is now dropped, the same as a stale one:

```js
if (p.verdict === PROBE_CONFIRMED) {
    if (p.age === null || p.age > PROBE_FRESH_S) return '';
    return ' The last check dropped the lamp' + when + found + ': real daylight.';
}
```

A refusal still survives the same gap — it explains the night the camera
is in now, whenever it was reached — and the time phrase is built as
`when` so an undated refusal reads "dropped the lamp and found 16x"
rather than with a hole in it.

**The between-poll projection was the wrong number to add.** I had run
the probe's age and wait through the same projection as the countdown
beside them. That projection anchors on `ageS`, which `projector()`
measures from the last time the *monitor's* gauges changed — streak,
pending, dwell — because those move on the camera's own tick, slower
than the poll. The probe's age and wait are not that: majestic
recomputes both from `CLOCK_MONOTONIC` on every scrape, so they arrive
current. On a settled night the monitor's gauges hold still for minutes,
and adding that number would have aged a check by however long the
monitor sat and floored "looks again in" to nothing while a stand-off
was still running. `probeState` no longer takes `ageS`.

Both phrases are coarse enough — minutes, and "a moment ago" under 90
seconds — that a poll period does not show.

Tests: sixteen checks now. New: a settled monitor (`ageS` 240) that does
not age a 120 s check with it and does not eat a 300 s wait, an undated
refusal on both doors, and an undated confirmation that is not called
fresh.

PR #432, opened 2026-09-11
settings: the frame rate control follows the chosen resolution

widgetii:
The Frame rate control read the same bound whichever resolution was picked,
because it read only the schema's `maximum`.

On a camera whose sensor has a binning mode good for 64 fps and whose encoder
manages 26 at full resolution, one number is wrong at both ends: it offers rates
the camera cannot reach at the large sizes, and it hides the fast ones behind a
resolution nobody would think to try. An owner setting 1080p saw the same
ceiling as at 352x288 and reasonably concluded the camera had one.

The daemon now publishes `x-fps-caps`, a rate per resolution it offers. The
control uses the per-size answer where there is one and `maximum` where there is
not, and retunes when the resolution changes.

**No explanatory note.** The control is simply right, which is the whole job — a
bound nobody has to read about is the one that works.

## Two things that made it wrong while still looking like a working page

**Lowering the max of an `<input type="range">` makes the browser clamp `.value`
on the spot.** A "did this need clamping?" test asked afterwards therefore always
answers no, the `input` event is never fired, and the readout beside the slider —
which repaints from that event and nothing else — goes on showing the old
number. A 2560x1440 row read **64** with the slider sitting at its 26 maximum.
The value is captured before the max moves now, and the event fires either way.

**Empty is a *value* for a resolution, not a missing one.** The picker's first
entry is "Auto · sensor native", which stores nothing so the camera uses the
sensor's own size. Reading that as "unset, consult the saved config" left
choosing Auto bounded by the resolution the operator had just navigated away
from — 55 fps offered for a stream that would run at 26. Empty now resolves
through the schema, `default` then `x-native`.

## Verified in a browser against a camera

Every entry in the picker:

| resolution | slider max | value shown |
|---|---|---|
| Auto (native 2592×1520) | 26 | 26 |
| 2592×1520 | 26 | 26 |
| 2560×1440 | 26 | 26 |
| 2304×1296 | 32 | 26 |
| 1920×1080 | 55 | 26 |
| 1280×720 | 64 | 26 |
| 640×360 | 64 | 26 |

The value stays where it was clamped rather than rising with the bound — raising
someone's frame rate because they changed resolution is not the page's call.

And a full round trip: picking 1280x720, pushing the rate to the offered 64 and
saving brings the stream up at 64 with 63 delivered.

## Tests

`mj-fps.js` is its own file and pure, so `tests/fps-caps.test.js` can ask it — a
control bounded wrongly is still a control, a slider that stops too early looks
exactly like a slider, and reproducing either needs a camera with more than one
sensor mode.

A daemon that publishes neither key leaves the page exactly as it was.

Needs the matching majestic change to publish `x-fps-caps`; without it this is
inert and the page behaves as before.


widgetii:
Pairs with a matching daemon-side change that publishes `x-fps-caps`. Inert without it — the page falls back to `maximum` exactly as before.

PR #441, opened 2026-09-12
nav: offer Restart camera in the System menu (#436)

### LINKED ISSUE #436: Feature request: Add reboot button
Feature request: Reboot button

I would prefer a button instead of the reboot command in console

Thank you

openipc-ai:
Fixes #436.

## The capability existed; the discoverability did not

`restart.cgi` reboots the camera when opened, and **Restart camera** buttons
already sit in the pending-changes banner, in the notice `camera.cgi` raises
when majestic is not running, on the WFB page, and behind banners in
`recordings.js` and `video-check.js`.

Every one of those is *contextual* — you are offered it because something else
brought you there. The only standing control was on a page called **Backup &
Restore**, which is not where anybody looks for a reboot. Hence the report:
the reporter was reaching for the console instead.

So this is placement, not a new mechanism.

## What changed

One row in `p/pages.cgi` and one entry in the System menu, under **Maintenance**
beside Update and Backup & Restore — the group it belongs to, on a bar that is
on every page:

```
System ▾
  Setup         Network · Time · Access
  Maintenance   Update · Backup & Restore · Restart camera
  Diagnostics   Logs · Console · File Manager · SD Card
```

It asks the same question the banner and the Backup page already ask, so all
three now word a reboot identically:

> Restart the camera now?
>
> Settings are kept. Video and recording stop for about half a minute while it
> comes back.

**The entry is coloured**, and that is the only part here that is a judgement
rather than a move. Every other item in the bar is navigation: following it
shows you a page, and following it by accident costs a click back. This one
reboots the camera, and a menu should not hide that among twenty entries it
looks exactly like. It is the bar's version of a rule this tree already states
for the buttons in notices — the class that asks first is spent only where
pressing acts on the camera rather than going somewhere.

## Verified on hardware

Driven in a real browser against a camera:

* the menu renders with **Restart camera** third under Maintenance, in
  `rgb(220, 53, 69)` against `rgb(230, 232, 238)` for every neighbour;
* **declining** the prompt leaves the page where it was and the camera
  untouched, with the intended sentence shown;
* **accepting** navigates to `restart.cgi` and really reboots — uptime reset,
  back in about fifteen seconds with the daemon running.

## Tests

`tests/nav-confirm.test.js` reads the bar's own entries and holds two things
`notice.test.js` holds for banners but cannot hold here, because it does not
read the bar:

* an entry whose page **acts** on the camera carries the class `main.js` hangs
  `confirm()` off, and one that merely navigates does not wear it;
* the acting entry's prompt is its own sentence naming the cost, rather than
  the generic "Are you sure?" that #160 found people clicking through.

It also refuses to pass on a scan that matched nothing, which is the one way a
file like this goes green while checking nothing.

It earns a file because the failure is silent in every direction: drop the
class and the markup still parses, the item still renders, the menu still looks
right, the lint passes and the suite goes green. The first sign of it is
somebody's camera rebooting without being asked — and nobody clicks a reboot to
check that it warns first. Both mutations were confirmed to fail the test.

## One thing worth a second opinion

`factory-reset.cgi` is deliberately kept **out** of the bar — `p/pages.cgi`
files it under "Reached from a page rather than from the bar". A reboot is not
in that class (settings are kept, the cost is half a minute), so I do not think
that precedent forbids this, but it is a reasonable place to disagree.


openipc-ai:
Addressed the review finding, and it was a real one — thanks.

I measured the mechanism in a browser against a camera before changing anything. A middle click on a menu item delivers `auxclick` and never `click`, so the `confirm()` `main.js` attaches never runs; "Open link in new tab" delivers no event at all. As an anchor this entry would have rebooted the camera unasked for anyone who opened a menu item the way they open every other menu item — from every page, since this bar is on all of them.

The entry is now a form that POSTs rather than an anchor: no URL to middle-click, to prefetch, or to restore with a tab. Verified on a camera — it renders identically to the anchors beside it (same width, offset, font and padding), a plain click still asks and declining does nothing, accepting still reboots, and a middle click now does nothing at all. The test gained the half a `confirm()` cannot cover: an acting entry must not be an anchor, and the scan reads buttons too, since reading only `<a>` would have made it blind to exactly the entry it exists for.

What I did **not** do here is change `restart.cgi` itself. It has always rebooted on a plain GET, five other call sites still link it as an anchor, and making it POST-only moves `tests/notice.test.js` and the notice component with it. That is wider than this issue, so it is filed as #442 rather than bundled in.

Issue #335, opened 2026-09-05
Live: Video flashes every 2 seconds with Main + MSE enabled


usa-:
On the Live page, when both the Main and MSE buttons are enabled, the video flashes every 2 seconds: video, then a black frame, then video again.

There are no JavaScript errors in the console.

**SOC**: gk7205v300 (gk7205v200)
**SENSOR**: imx335
**FIRMWARE**: 2.6.09.04-lite
**BUILD**: master+2a0d5ca, 2026-09-04
**MAJESTIC**: Lite HiSilicon (gk7205v200), master+707ff6d, 2026-09-04 17:31

Browser: Safari Version 26.6 (20624.4.5.18.1)

[majestic.yaml](https://github.com/user-attachments/files/31858256/majestic.yaml)

openipc-ai:
@usa- — thanks for the report and the config. I tried to reproduce this and couldn't yet, and the reason is useful to know: I don't have a camera that matches yours closely enough. Your Main is H.265 at **30 fps** on a gk7205v300; the hi3516ev200-class parts I have here top out at 25 fps for this sensor, and at 20–25 fps the H.265 stream comes through clean — one init segment, no periodic reset. And the flash is in Safari, which I also can't run here. So a couple of targeted questions from you would pin it down much faster than me guessing:

1. **Stream:** does it flash on the **Sub** stream (H.264) too, or only on **Main** (H.265)?
2. **Transport:** only on **MSE**, or also on **WebRTC**?
3. **Browser:** only in **Safari**, or does it also happen in **Chrome/Chromium** on the same machine?

Those three answers separate the likely causes cleanly — whether it's the camera re-sending the video's init segment (which resets the player and would show as a black frame), the page nudging playback to the live edge, or something specific to Safari's MSE decoding.

Two more, if it's easy:

4. Anything in the browser **console** at the moment of a flash — even a warning, not just errors (Safari: Develop → Show Web Inspector → Console)?
5. Does it change if you drop Main to **25 or 20 fps**, or if you turn the **OSD clock off** for a moment? (Just to see whether it's tied to framerate or to the per-second overlay redraw — no need to keep those settings.)

No rush, and I know the camera goes away Monday — whatever you can grab before then helps. If it turns out to be Safari-only on MSE, that narrows it to a handful of lines on our side.

usa-:
I checked this as requested.

1. The flickering happens only with the Main stream.

[...]

PR #456, opened 2026-09-13
preview: reset the video element in place, don't clone it, for Safari HEVC (#335)

openipc-ai:
## What / why

On the Live page with **Main + MSE** (H.265) in **Safari**, the picture flashes repeatedly and then drops to the software/MJPEG rung — OpenIPC/majestic-webui#335.

Reproduced on a **real Mac** (macOS 15, **Safari 26.6.1**, hardware VideoToolbox decode) against a **gk7205v300 + imx335** camera — the reporter's exact configuration — driving the actual Live page:

- `MEDIA_ERR_DECODE` (code 3) fires within the first GOP (~0.7 s),
- the `DECODE_MAX` rebuild loop turns that into the ~2 s flash,
- two strikes fall through to the software/canvas decoder.

## Root cause

`freshVideo()` replaced the `<video>` element with a fresh `cloneNode()` on every (re)connect. **A freshly cloned element attached to a `MediaSource` faults Safari's hardware HEVC decoder.** The very same fMP4 bytes, appended to the element **reused in place**, decode cleanly.

Isolated by swapping only that one step on the reporter's platform, all else held constant:

| element handling | result on Safari 26.6.1 (HW decode) |
|---|---|
| clone + replace (before) | repeated `err=3` → software fallback (the flash) |
| reuse in place (this PR) | plays steadily |

The appended fMP4 is valid either way — it decodes clean in a bare MSE player — and the other suspects were each ruled out on the same Mac: append batching (`APPEND_BATCH`), `sb.mode`, the live-edge seek (`syncLive`), the stage DOM/CSS, fps (25 and 30 both failed before), and WebRTC/wasm decoder contention.

It never showed on the `safari-hevc-qa` CI runners because those decode HEVC in **software**, where a cloned element is harmless; only a real Mac's **hardware** decoder trips on it.

## The change

`freshVideo()` now clears `src`/`srcObject` and calls `load()` on the existing element and re-applies `muted`/`volume`, instead of cloning and replacing it. Same clean reset for the next session, no new element for the decoder to choke on. One function, +16/−13.

## Validation

- **Reporter's platform** (lab gk7205v300+imx335, Safari 26.6.1, HW decode), real Live page: before — repeated `err=3` → software fallback; after — steady playback (`err` clears, `currentTime` advances through 20 s+), no fallback. Verified over multiple runs.
- **Unit tests**: `preview-socket`, `preview-live-edge`, `preview-datachannel`, `mj-preview`, `preview-chain`, `preview-stats`, `preview-hero`, `video-check` all pass. Change doesn't touch the append path the software-decode runners exercise.

## Notes for review

- Under the safaridriver harness I still see **one** transient `err=3` at cold start that immediately recovers (single `DECODE` strike, no software fallback) before steady play — a large improvement over the perpetual loop, but not provably zero. @usa- — could you confirm on your Mac whether Main + MSE now holds steady (a brief blip at startup is fine; what matters is it no longer flashes on)?
- Follow-up worth doing: `safari-hevc-qa`'s recorded stream is all-keyframe 2592×1520, which can't exercise this mid-GOP hardware-decode fault; a representative 1080p30 recording (normal GOP with P-frames) would regression-protect this.


openipc-ai:
**Addressed the review finding** ("Tests still simulate detached videos") in 2174315.

`freshVideo()` reuses the element in place now, so fixtures that modeled clone-and-replace were validating a detached-node path production no longer takes:

- `preview-socket` and `preview-live-edge`: the fake `<video>` drops `cloneNode()`/`parentNode.replaceChild` — one node reused across reconnects, with an `srcObject` to clear like the real reset. (The review flagged `preview-socket`; `preview-live-edge` carried the same fixture, so I fixed it too.)
- `staging`: the MSE stub no longer swaps in a fresh node on attach; the now-unused `replaceNode` helper is removed, and the group that asserted the swap survives a replaced node now reads as keeping the element in place.
- `auto-source`: dropped the stale "MSE replaces the video nodes → detached" rationale from the observe-the-stage test; the reason that stands is the viewport clamp resizing the stage.

Full test suite green.

Issue #416, opened 2026-09-08
Day & Night: slider shows 33 when GAIN MULTIPLE THAT MEANS NIGHT is unset


usa-:
After upgrading a camera from firmware that used the old image-based day/night detection algorithm, I open **Camera → Settings → Image → Day & Night** and disable **LEGACY SETTINGS**.

The **GAIN MULTIPLE THAT MEANS NIGHT** slider moves to the position corresponding to **33**, but the value itself remains unset. The only indication that the value is actually unset is that **33** is not displayed to the right of the slider. The user can easily assume that the current value is 33 until they move the slider back and forth, at which point the value appears.

Would it be more appropriate to set **GAIN MULTIPLE THAT MEANS NIGHT = 33** when disabling **LEGACY SETTINGS**, rather than leaving the value unset while displaying the slider at position 33?

Or maybe this is the desired behavior?

openipc-ai:
Thanks for this — it turned out to be two faults, and the answer to your question is no: **33 should not be written into the field.**

### Why not

An empty **Gain multiple that means night** is not a missing number, it is a different way of deciding day and night. With a value set, gain alone decides in both directions — night above the multiple you chose, day at or below the day multiple. With the field empty, the camera uses its calibration-free trigger instead: night begins when the automatic exposure can no longer brighten the picture, and the return to day consults that same state. So filling the field in on your behalf would quietly move your camera onto the other mechanism, and 33 is not even a number anybody chose — see below.

### What was actually wrong

An HTML slider has no empty state. Given no value the browser parks the thumb in the **middle of the track** and then reports that position as the value: for a 1–64 field that is 32.5, rounded up to 33. The page did know the field was unset — that is why no number appeared beside the slider and why saving did not write anything — but a blank readout is an absence, not a statement, and you read the handle instead. Quite reasonably.

Your report also found a second one nobody had noticed: **Reference enhancement layers** on the Main stream page (0–3, also with no default) drew its handle at 2 of 3, two thirds of the way along, looking very much like a deliberate setting.

### What changed

An unset slider now says so. The readout beside the track prints **Unset**, and the handle steps off the track until you hover it, focus it, or start to drag — so there is nothing sitting at a number nobody picked. The moment you move it, it becomes an ordinary value and the save bar appears as usual.

[...]

PR #472, opened 2026-09-13
settings: say which settings are off stock, and what a reset will do

widgetii:
The section head prints "4 of 17 off stock" — true, provable, and unusable: every row carried the same muted reset arrow whether or not it had anything to undo, so the only way to find the four was to open the config file, which is the thing this page exists to replace.

### The arrow is the mark

It is the undo for exactly this fact, so it now carries the fact:

- **lit, in the amber the count is printed in** — off stock, press to undo;
- **hidden until the row is hovered or focused** — everything else: a row at its default, where majestic keeps no key at all (it writes back only what differs from the build's defaults) so the press provably changes nothing, and a row whose key the schema gives no default for, where the glyph is a *clear* rather than an undo and the page cannot show it has anything to clear. Still reachable, because clearing is the only route back to unset on a control that cannot be emptied by hand (#416).

So a glyph on screen means one thing: this row is off stock. The row's own left border could not carry it — amber there already means unsaved, and off stock is a different fact that survives Save.

`paintStock()` also no longer returns before the rows when a leaf has no head to print into, so the Overlay, Motion detection and Live adjustments leaves mark their rows too.

### A reset that switches something off says so

Pressing ↺ on Main stream's **Enable** took the main stream down on a bench hi3516ev300 — the encoder stopped, and the page then honestly showed a switch nobody had touched sitting at off.

Neither question this button asked said that was coming. "Clear this setting and leave it to the camera" reads as a promise that the camera has a sensible answer of its own; on a switch with no declared default it has exactly one, and it is off, because `config_get_boolean` answers false for a key that is not there and every generated boolean accessor funnels through it. "Reset to its default" is no better on a switch whose default is false: it names an action and never the value.

A press about to turn a switch that is currently **on** off now says so, in the question and in the title — judged at the press, since a switch already off has nothing to be warned about. Every other clear keeps today's words.

### A rate of zero is the sensor's rate

`video<N>.fps` defaults to 0, which the camera reads as "run at whatever the sensor mode gives" and publishes as `x-fps-sensor`. The slider printed the number, and while the schema's floor was 1 the control rounded it up and said a camera delivering 25 frames was delivering 1. Where the camera names the sensor's rate, 0 now reads `Auto · 25` in the value box and in what the reset arrow offers to put back. Other ranges are untouched: `sliceUnits` is 0 because it is disabled.

An empty default prints as a word rather than as a sentence stopping at its colon.

### Checks

`npm test` and `tools/lint-templates.sh` green. Rendered against a bench hi3516ev300 at 1440px in both themes, and on the Main stream, ISP, RTSP, Day/Night, Overlay and Motion detection leaves: marks match the head's count, hovering an at-stock row reveals its arrow, and typing a value away from its default lights that row's arrow and moves the count.

The daemon-side half — the defaults these rows describe — is a separate change to majestic.


widgetii:
Both review findings addressed in 7fafbb6.

**The rate control starts where its default is.** The bug is real and I had the daemon-side fix masking it: against a camera whose schema still puts the fps floor at 1 while declaring 0 as the default — every build in the field until that fix lands — the browser sanitises 0 up to the floor, so the Auto reading never ran, the row showed 1 and counted as off stock. The track now starts at 0 wherever the camera names the sensor's rate beside the field, whatever floor the schema carries. Checked by serving the page a schema with `minimum: 1`: the control reads 0, shows `Auto · 25` and is not marked off stock.

**The rationale names behaviour, not internals.** Reworded to what is observable from outside the camera: a switch whose key is absent reads off, measured by clearing the main stream's Enable and watching the encoder stop.

PR #473, opened 2026-09-13
delivery: send the recording itself, and watch both senders do it

openipc-ai:
The camera can now run a command when a recording is finished, handing it the clip's path, why it ended and roughly how long it is. This makes the two senders consumers of that: the recording itself gets delivered, with the run-up before the event that a snapshot can never have.

### The senders take a file

Called with a path, `telegram` and `ntfy.sh` send **that file** instead of taking a picture — the clip is what the event was, and a still taken afterwards shows what the camera can see now rather than what set the recorder off. Called with nothing, both behave exactly as before: the cron line and the `?send=` webhooks are untouched.

`sbin/record.sh` is what the camera is pointed at. It forwards **motion clips only** — a camera recording continuously closes a clip every few minutes whatever is in front of it, and a chat that receives one around the clock is a chat nobody reads. Each sender decides for itself in its own config, so an image that does not ship one (the FPV build leaves out the Telegram sender) simply has nothing to call.

The two pages get a *Send motion clips* switch. Turning either on points the camera at the dispatcher **only if nothing else is there**: an operator who has already pointed it at a script of their own keeps it and is told, rather than having it overwritten, and turning both off clears only ever our own value.

### The harness, and why it earns its place

`tests/delivery.test.js` landed before either sender changed. Everything else about these scripts is checked by reading them — `sh -n` has their syntax, the template lint has the pages that write their config — and reading cannot see what a send does, which is exactly where both of them failed **silently**:

- `bin/ntfy.sh` ended in a bare `exit 0`. A rejected topic, a wrong password and a delivered notification were one answer to the operator, and the page's test button printed OK for all three.
- It also `echo`ed its own command line, which carries `-u user:pass`.

So: one zero-dependency server playing the camera, the Bot API and ntfy at once; the real scripts run against it; assertions about the bytes that arrived, the endpoint chosen, the caption, the filename and the exit status. The scripts are **rewritten rather than re-typed** — the config path and the endpoints are substituted, and a substitution that no longer matches fails the test rather than skipping it — for the reason `notice.test.js` cuts its functions out of the real file: a test carrying its own copy of the subject stops describing it the first time somebody edits the original.

Against the scripts as they were, **14 of its 30 checks failed**: every one about a clip, the rejected upload, the password on stdout, and the disabled integration reporting success. They pass now.

### What ntfy.sh gained along the way

What the harness measured missing: an exit status from `%{http_code}`, a private temp directory with a trap instead of a predictable `/tmp` name (this runs as root, from a webhook anyone who can reach the camera can fire — `sbin/telegram` carries the same note), no more printing the credentials, and a proxy block that reads the keys the proxy page actually writes rather than two spellings that have never existed anywhere.

Both scripts now use `.` rather than `source`. busybox takes either; dash — what `/bin/sh` is on the machines the test suite runs on — has only the POSIX spelling.

### Verified

`npm test` green, including the new file. Deployed to a camera recording on motion: a real motion clip reached the receiving end one second after the recorder closed it, 665748 bytes, `Content-Type: video/mp4`, named after the clip.

A camera whose build cannot run a command when a recording finishes refuses the setting, and the page says so rather than reporting a switch that did nothing.


openipc-ai:
All seven were real; fixed in e388a15.

**3. Clip names can execute shell commands.** The sharp one. The path is caller-supplied now and both senders `eval` their command string, so a single quote in it closed the quoting — and the path comes from a strftime pattern an operator types, so that is a name somebody can have. Escaped now (close, escape, reopen), as are the caption and the `Filename` header, which travel the same road and where an apostrophe is ordinary rather than exotic. The harness gained the case, and it can fail: against the old interpolation it errors with ``unexpected EOF while looking for matching `'``.

**4. Clip sends expose the bot token.** Correct, and the test missing it was the worse half. `--verbose` is gone — the response body is printed either way and the exit status is the answer — and both secret checks now read stderr as well as stdout.

**1 + 5. Failed reads can replace custom hooks / owners see a false success.** Right, and collapsing the two failure states is exactly what `mj_cfg`'s three-way contract exists to prevent — the empty branch *writes*, so a camera mid-restart or a build without the setting could have had an operator's own command replaced on the strength of a request that never landed. Kept apart now, and a read that failed changes nothing. The message reaches the page too: `set_error_flag` wrote into the flash log and the success redirect that followed overwrote it, so the helper now hands the sentence back and each page picks its own.

**2. Hook updates can be falsely confirmed.** Writes are read back now rather than trusted — this API can accept a value and keep none, and a page reporting success from a status code would state the opposite of what the camera is doing.

**7. Disabling clips leaves a camera hook set.** `mj_clear` sends a null leaf, which the config API takes as a removal, instead of storing an empty string where there had been no setting at all.

**6. Failed deliveries still report success.** `record.sh` now exits with the worst status. The camera does not read it, but a person running it by hand does, and "the clip went nowhere" must not look like "nothing to do".

PR #479, opened 2026-09-14
PTZ: the missing-driver sentence is a toast, not a cell in the pad

widgetii:
## What it looked like

On a camera that declares a motorized lens but has no driver to serve it, the Live page's zoom/focus pad swelled to **1181×250** on a 1280px page — 27% of the picture — with the sentence on its own line and the second column empty. On a phone it was 39%.

## Why

The reason line was emitted inside the pad:

```html
<p class="small" role="status"><%= $ptz_reason %></p>
```

with the note *"`small` only — a class the stylesheet already carries, so the pad gains no new CSS."*

The class was never the problem; the parent was. `#mj-ptz-fn` is `display:grid; grid-template-columns: repeat(2, 1fr)`, so the paragraph became a grid **item**. `1fr` is `minmax(auto, 1fr)`, the pad is positioned over the stage with nothing constraining its width, and so both columns sized themselves to the sentence's max-content — 585px each. It also occupied the first cell alone, leaving the second empty and pushing the ZOOM caption down a row.

Reading the diff could not have shown this, and `tools/check-layout.mjs` does not catch it either: nothing was clipped, nothing overlapped, nothing spilled. The pad was simply the wrong size. (A rule for "an overlay must not swallow the picture" looks worth having, but it is not this change.)

## Why not just span both columns

Measured. `grid-column: 1 / -1` plus `max-width: 13rem` on the pad gives **208×305** — still twice the pad's size, because 45 words inside a 134px control wrap into a taller column instead of a wider one. On a phone that is 20.7% of the stage against 9.2%. The message is prose and the pad is a grid of 40px buttons; they do not belong in the same box at any width.

## What it does now

The sentence goes to `#mj-toasts`, where the page already puts standing explanations — the stack bounds its own width, hangs under the chip, and already carries the served-channel message, which is the same shape (true for the whole session, nothing to time out). `preview-ptz.js` relocates it there the way it already relocates the pads.

It is now emitted for the pelco backend as a whole rather than inside the zoom/focus group, which fixes a second case: a camera whose `ptz_caps` names only `pan tilt` drew a pad and no explanation at all.

Nothing is lost if the relocation never runs — pressing any button answers with the same text from `j/ptz.cgi`.

One new rule, because this toast is the only one with no `×` to dismiss (the driver is missing until somebody installs it; there is no state to clear): it gives back the room the corner button reserves and stops offering a pointer for a click that does nothing.

## Measured

On an hi3516ev300 with the driver removed, at 390 / 1280 / 2560:

| | pad, before | pad, after | toast |
| --- | --- | --- | --- |
| phone 390×844 | 382×313 (39.0%) | 134×211 (9.2%) | 293×75 |
| laptop 1280×900 | 1181×250 (27.4%) | 134×211 (2.6%) | 352×54 |
| monitor 2560×1440 | 1181×250 (8.4%) | 134×211 (0.8%) | 352×54 |

134×211 is what the pad measures when the driver is present, so the two states now render identically apart from the toast. No overlap, no horizontal scroll.

`node tools/check-layout.mjs` — clean at all three widths. `npm test` — all passed. `tools/lint-templates.sh` — ok, 58 templates, 24 shell scripts. `tools/regen-bootstrap-css.sh` leaves `www/a/bootstrap.min.css` unchanged.


widgetii:
Both findings addressed in eac81fc.

**1. The warning blocks picture taps — confirmed and fixed.** The finding is right and the mechanism is exactly as described: `.mj-toasts > *` sets `pointer-events: auto`, and `preview-hero.js` returns early for any touch whose target is inside `#mj-toasts`. That is correct for the two toasts a press does something to — tapping either dismisses it — and wrong for this one, which answers nothing.

Measured before the fix, hit-testing the toast's centre and both far corners:

| viewport | toast | `elementFromPoint` | tap on the picture |
| --- | --- | --- | --- |
| 390×844 | 293×75 | the paragraph | swallowed |
| 1280×900 | 352×54 | the paragraph | swallowed |

After `pointer-events: none`, all six points return the video element and the tap reaches the stage. `cursor: default` went with it — an element that is not hit-tested never supplies a cursor, so the rule was doing nothing; the picture underneath supplies it, which is the truth about what is under the finger.

**2. The motor template embeds a postmortem — taken, partly.** The pixel sizes and the pad's grid dimensions are facts the stylesheet owns and will move without telling the comment, so they are gone. What stays is the part that cannot go stale and is the reason the paragraph is not where a reader would expect it: prose put in either pad is a grid item and sizes the track it lands in. The measurements live in the commit message and in this PR, where they are dated.

`check-layout.mjs` clean at 390/1280/2560, `lint-templates.sh` ok, `build:dist` ok.

PR #480, opened 2026-09-14
sysinfo: a cached answer dies with the majestic that gave it

widgetii:
## What it looked like

Install the motor driver package on a running camera, restart majestic, reload the Live page — and it went on telling the operator to install the package they had just installed. The camera had PTZ, `GET /ptz` said so, and the UI disagreed until somebody rebooted it.

## Why

`update_caminfo` caches to `/tmp/webui/sysinfo.txt`, and `common.cgi` regenerated that file only when it was **absent**:

```sh
[ ! -f $sysinfo_file ] && update_caminfo
```

Most of what it holds is a fact about the camera — the SoC, the flash size, the MAC. But some of it is not a fact at all, it is an answer the daemon gave: `mj_version`, and the whole PTZ/autofocus verdict, which asks majestic whether it can drive the motor wire. Those are good for exactly as long as the majestic instance that gave them, and nothing said so.

The file lives in `/tmp`, so a reboot cleared it, and that covered the only case anyone had hit. It is not the only case. Nor is there a hook to invalidate from, the way a save on a page has one — `update_caminfo` is called from `access.cgi`, `network.cgi` and `time.cgi`, all places where *this* UI changed something. Here what changed happened outside it entirely, so the cache has to notice by itself.

## Noticing for free

Measured on an hi3516ev300: a page load reusing the cache takes about 100 ms, and one that rebuilds it about 500 ms. That 400 ms is the whole reason the cache exists, so the staleness check has to be cheap or it eats the thing it is protecting. A check that forks `pidof` costs about 12 ms — a tenth of the saving, on every page of the UI.

So `update_caminfo` records who answered, and the check is builtins only: `read` from `/proc/<pid>/comm` and `/proc/<pid>/stat`, no subshell. A thousand calls come to a second on that board, against zero for an empty loop of the same length; a thousand `pidof` forks come to twelve seconds.

The recorded identity is the pid **and** the task's start time (field 22 of `stat`), not the pid alone. A pid is unique only while its process lives, so a name check by itself would accept a later majestic that landed on the same number. The start time is monotonic, which also keeps this indifferent to the clock being stepped — a file-mtime comparison would have gone permanently stale if the clock ever stepped backward between the daemon starting and the cache being written.

One path still forks: a cache with nothing recorded — written before this existed, or written while majestic was down. It asks once. A daemon that is *still* down leaves the record alone rather than paying 400 ms on every page of the UI somebody is using to go and fix it, and "could not ask" is not a better answer than the one already on file.

## Verified

On an hi3516ev300 with a motorized lens, driver removed and restored by hand:

| step | result |
| --- | --- |
| driver absent, majestic restarted, page loaded | cache records pid, start time and the missing-driver verdict |
| five more page loads, nothing restarted | cache mtime unchanged — reused, not rebuilt |
| driver installed, majestic restarted, **no reboot, nothing deleted** | `GET /ptz` → 200 |
| next page load | verdict regenerated by itself, message gone, new identity recorded |

Every branch of `caminfo_stale` exercised directly on the camera: the same instance reads fresh; the recorded start time moved by one — which is what pid reuse looks like — reads stale; a pid that has gone, a pid now belonging to `init`, a cache with a pid but no start time, and a cache with neither all read stale. Globbing is intact after the field split. The end-to-end run was then repeated against the stripped build that actually ships.

`npm test` — all passed. `tools/lint-templates.sh` — ok, 58 templates, 24 shell scripts, including `no 'local' outside a function anywhere`. `npm run build:dist` — ok, 145 assets carry no comments.


widgetii:
Both findings addressed.

**1. A lab transcript becomes public — taken.** The description had the benchmark numbers formatted as a three-line block, which reads as pasted console output whatever its provenance. Rewritten as prose: hardware class, then the timings in a sentence. The code comment the finding cites is a measurement attributed to a hardware class, which is what the rule asks for, so it stands as written.

**2. Daemon restarts can preserve stale data — fixed in d654094.** The finding is right about the claim being wider than the test: a pid identifies a process only while that process is alive, so checking the name at the recorded pid accepts a later majestic that landed on the same number. Unlikely on a camera — pids climb, and the counter has to wrap the whole space to collide — but the check said "the same instance" and tested "something called majestic".

So the cache records field 22 of `/proc/<pid>/stat` as well, the task's start time in jiffies since boot. That also settles a question the summary's alternative 2 raises: it is monotonic, so unlike a file mtime it is indifferent to the clock being stepped. A backward step between the daemon starting and the cache being written would have made an mtime comparison read "stale" on every page from then on, which is a persistent 400 ms tax rather than a one-off.

It stays fork-free — `read` and `set --` are builtins. On the same board a thousand `caminfo_stale` calls come to a second against an empty loop's zero, where a thousand `pidof` forks come to twelve. `set -f` is on for the one field split, because field 2 of that line is the command name in parentheses; the name is checked first, which is what makes the split safe, and globbing is verified intact afterwards.

Every branch exercised on the camera: same instance → fresh; recorded start time moved by one (what pid reuse looks like) → stale; pid gone → stale; pid now belonging to `init` → stale; a cache with a pid but no start time → stale; a cache with neither → stale. Then the end-to-end run again — driver installed, majestic restarted, no reboot — against the stripped build that actually ships.

widgetii:
On the remaining rule violation: it re-ran against the previous description. The three-line block it quotes — `caminfo_stale x1000` / `pidof fork x1000` / `empty loop x1000` — is no longer there. The description now gives those timings as a sentence, with the hardware class and no output-shaped block anywhere in it.

The code comment the finding cites is a measurement attributed to a hardware class, which is what Rule 6 asks verification to look like, so that stays as written.

Rebased onto master now that #479 has landed, so CI and the camera both see the two changes together. Verified on an hi3516ev300 against the stripped build that ships: with the driver removed and majestic restarted, the explanation is in the toast stack and nothing is inside the pad; with the driver installed and majestic restarted — no reboot, nothing deleted — the next page load drops the toast by itself, the pad is still drawn, and a move reaches the motor.

Issue #478, opened 2026-09-14
An upgrade socket that closes before any flash is reported as a reboot, then waits eight minutes


openipc-ai:
Found while investigating #474. Independent of #477, and it still bites once that one is fixed — a camera can die without a word, and then there is nothing to decode.

## What happens

`www/a/update.js` ends the transcript with one sentence whatever happened:

```js
function endLog() {
    if (logClosed) return;
    logClosed = true;
    term.note('--- connection to the camera ended here; it is rebooting ---');
}
```

`ws.onclose` calls it, and then `beginPollBack()`, which says `Waiting for the camera to reboot — do not power off…` and starts pinging. A camera that never went anywhere answers every ping, so `downSeen` stays false and the run costs `DOWN_TRIES = 160` ticks of 3s — about **eight minutes** — before the page finally says `The camera never rebooted — the update may have failed.`

"it is rebooting" is a claim the page has no evidence for. The file already knows the difference: `sawFlash` is set by the markers the updater prints at its point of no return (`Protected: flashing…`, `Kernel updated`, `RootFS updated`), and `rebootMarker` is set when the updater announces the reboot outright. With **neither** of those seen, nothing has been written and nothing has said it is going, so the page is guessing — and in #474 it guessed wrong and told the reporter so in the one line he had to go on.

## What it should say

- End the transcript on what is known. With a flash marker or a reboot announcement behind it, `--- connection to the camera ended here; it is rebooting ---` is correct. Without either, `--- connection to the camera ended here ---` is the whole of what the page can honestly say.
- Recognise the camera's own "the run is over" statements the same way `Aborting.` is already recognised, so a failed upgrade ends immediately with the reason instead of after eight minutes of waiting for a reboot the camera has explicitly said is not coming. (Those statements only become readable once #477 lands — the two changes are worth making together.)

The "do not power off" warning should stay on any path where a flash may genuinely be under way; the point here is not to drop caution, it is to stop asserting a reboot that nothing observed.

Refs #474, #477.

Issue #474, opened 2026-09-13
Firmware upgrade fails on ssc337de


usa-:
Firmware upgrade stops after the first line of output. The camera remains on the current firmware version and does not reboot. SSH access remains available.

SOC: ssc337de (infinity6b0)
SENSOR: gc4653
FIRMWARE: 2.6.09.13-lite
BUILD: master+910e724, 2026-09-13
MAJESTIC: Lite SigmaStar (infinity6b0), master+98b8049, 2026-09-13 17:29
WEBUI: master+4dae29a, 2026-09-13 20:19
U-BOOT: U-Boot 2015.01 (Aug 20 2024 - 15:01:25)

<img width="740" height="530" alt="Image" src="https://github.com/user-attachments/assets/17dd47ec-76bd-4798-a889-c6c53501f7dc" />

usa-:
This model is also affected:

**SOC**: ssc378de (infinity6c)
**SENSOR**: imx415
**FIRMWARE**: 2.6.09.13-lite
**BUILD**: master+910e724, 2026-09-13
**MAJESTIC**: Lite SigmaStar (infinity6c), master+98b8049, 2026-09-13 17:28
**WEBUI**: master+4dae29a, 2026-09-13 20:19
**U-BOOT**: U-Boot 2015.01 (Oct 03 2024 - 22:45:04)

openipc-ai:
Thanks for both boards. There are two things tangled together here, and the first one is why you were left with nothing to go on.

### The page was throwing your camera's answer away

`/ws/upgrade` carries the updater's transcript as **binary** frames, but the camera says everything *about* the upgrade — its refusals, and the line it sends when the updater exits without rebooting — in **text** frames. The page decoded every frame as bytes, and `new Uint8Array(aString)` is a zero-length array, so all of those sentences rendered as nothing at all. Measured on a camera here:

```
<<< BINARY [133] 'Firmware\nFile /tmp/nosuch.tgz not found Aborting.\n…'
<<< TEXT    [80] '\r\nUpgrade did not complete. Video has been restarted; the camera is unchanged.\r\n'
<<< CLOSE
```

The binary lines are what you normally see in the Progress pane; the text line is the one that says it is over, and it never reached a browser. When the camera refuses *before* the updater is started there is no transcript at all, so that invisible frame is the entire message — which is exactly your screenshot: the upload line, then nothing, then the page's own guess that the camera was rebooting.

[...]

PR #481, opened 2026-09-14
Render text diagnostics from upgrade WebSocket

### LINKED ISSUE #477: Upgrade page discards every diagnostic the camera sends about a failed upgrade
Found while investigating #474, and it is why that report cannot be diagnosed by the person who filed it.

## What happens

`www/a/update.js` decodes every frame arriving on `/ws/upgrade` as bytes:

```js
ws.onmessage = e => append(dec.decode(new Uint8Array(e.data), { stream: true }));
```

For a **binary** frame that is right, and the sysupgrade transcript is binary, so the log renders.

For a **text** frame `e.data` is a `DOMString`, and `new Uint8Array(aString)` is a **zero-length array** — the typed-array constructor takes the iterable path only for an `Object`, and a primitive string goes down the length path, where `ToIndex("ERROR: …")` is `NaN` and therefore `0`. `binaryType = 'arraybuffer'` does not help; it governs binary frames only.

So every text frame is rendered as the empty string. And text is exactly what the camera uses for the sentences that say what went wrong: the four `ERROR: …` refusals it answers a bad or unstartable upgrade with, and the two endings it sends when the updater exits without rebooting —

```
Upgrade did not complete. Video has been restarted; the camera is unchanged.
Upgrade did not complete, and video could not be restarted. Reboot the camera.
```

None of those has ever reached a browser. What the operator gets instead is a blank progress pane, then the page's own note `--- connection to the camera ended here; it is rebooting ---`, then an eight-minute wait for a reboot that is not coming (that second half is #478).

## Evidence

Measured against a camera with a raw WebSocket client that prints each frame's opcode.

A refused upgrade — one frame, and it is text:

```
>>> {"source":"/etc/passwd","kernel":false,"rootfs":false}
<<< TEXT   [35] 'ERROR: invalid upgrade parameters\r\n'
<<< CLOSE  1000
```

A real upgrade that aborts before the flash (`source` naming a file that is not there) — the whole transcript arrives as binary and is rendered, and then the one sentence that says it is over arrives as text and is not:

```
<<< BINARY [299] 'OpenIPC System Updater v1.0.63 … Stop services, sync files, free up memory …'
<<< BINARY  [24] 'OK\nStopping syslogd: OK\n'
<<< BINARY [974] 'Uptime: … Memory: … Processes: … Unmounting SD card …'
<<< BINARY [133] 'Firmware\nFile /tmp/nosuch.tgz not found Aborting.\nRestarting the services stopped for the upgrade\n'
<<< BINARY  [49] 'majestic is restoring itself; leaving it running\n'
<<< TEXT    [80] '\r\nUpgrade did not complete. Video has been restarted; the camera is unchanged.\r\n'
<<< CLOSE  1000
```

And the decode, in isolation:

```js
new Uint8Array('ERROR: invalid upgrade parameters\r\n').length   // 0
```

That abort case happens to survive because the page also watches for `Aborting.` in the binary log. Nothing rescues the cases where the camera refuses **before** the updater is spawned — there is no transcript at all then, and the refusal is the only thing there was to show.

## The fix

Decode by frame type rather than by assumption:

```js
ws.onmessage = e => append(typeof e.data === 'string'
    ? e.data
    : dec.decode(new Uint8Array(e.data), { stream: true }));
```

The camera should also be sending these on the one channel it already streams the transcript on, so that the sentences about the log arrive through the code path that is known to work; that half is filed separately against the daemon. Both are worth doing — the page must not be the reason a camera's own words are lost, whichever frame type they turn up in.

Refs #474.

iibaranov-IG:
Fixes #477.

The upgrade socket handler now appends WebSocket text frames directly and keeps the existing streaming `TextDecoder` path for binary frames. Camera refusals and final failure explanations therefore reach the terminal instead of becoming an empty typed array.

Validation:
- the new text-frame assertion fails on the pinned baseline and passes here
- the existing binary replay assertion still passes
- focused `update-reattach.test.js`: 10 checks passed
- both changed JavaScript files pass `node --check`

The full local `npm test` reached the unrelated Telegram/ntfy still-capture tests, where nine assertions depend on camera-side still retrieval and failed in this non-camera environment. No firmware was flashed locally; the owner check remains a refused/aborted upgrade on a camera.


iibaranov-IG:
Addressed the mixed-frame decoder-state finding in 32a32e9. Before a WebSocket text frame is appended, the streaming TextDecoder is now flushed so pending binary UTF-8 bytes cannot cross the frame-type boundary. The focused regression covers a partial binary multibyte sequence, a text diagnostic, and a subsequent valid binary character in exact frame order. The focused test and full npm test suite pass.

openipc-ai (approved):
Approving.

The change is the fix #477 prescribes — decode by frame type instead of assuming binary — and it is the idiom this tree already uses in `www/a/preview.js` for the `/ws/video` signalling line, so it is consistent rather than novel.

Verified rather than read:

- both new assertions are real regression tests: each fails on the commit before it and passes after — the text-frame one against the pre-fix handler, the flush one against the first commit;
- `npm test`, `npm run build:dist` (terser + `node --check`), the haserl template lint and the regenerated purged bootstrap subset are all clean locally. CI had not run on this branch at all — a fork pull request waits for workflow approval — so I approved the run; all four jobs pass;
- measured on an hi3516ev300 + imx335: a raw WebSocket client confirms the camera really does send these diagnostics as text frames on a current build, and the page itself — driven headless through a genuinely failing upgrade, the kind the updater refuses before it touches flash — now ends the transcript with the camera's "Upgrade did not complete…" sentence, where the shipped build showed nothing at all after the updater's last line.

Three notes, none of them blocking, all cheap to pick up later:

1. The type check carries no comment, in a file where every non-obvious branch has a paragraph. The trap is invisible at the call site: `binaryType = 'arraybuffer'` two lines above reads like a guarantee that `e.data` is an ArrayBuffer, and `new Uint8Array(aString)` is a silent zero-length array rather than a throw. This is the line a future reader simplifies back. The comment that was added explains the flush, which is the smaller half.

2. The second commit's scenario is not reachable on this endpoint. On every path this page can reach, a text frame is the last thing the camera sends before it closes, so there is no later binary data to inherit a stale prefix. The flush is harmless and I have kept it — but the symmetric case that *is* reachable, a stream ending with a partial sequence and no text frame after it, is still unflushed at `onclose`.

3. The test's payload, `ERROR: invalid upgrade parameters`, is a refusal this page cannot elicit: it validates the kernel/rootfs selection client-side and always names its own source path. The text that does arrive in practice is the pair of "Upgrade did not complete…" endings, which is what I exercised on hardware.

The other half of what the reporter of #474 saw — the page announcing a reboot that is not coming, then waiting eight minutes — is #478 and stays open. This fix is what makes the camera's contradicting sentence visible right above that note.

PR #486, opened 2026-09-14
raw: a capture page that works without the editor

widgetii:
Cameras that serve `/image.dng` have had no way to look at what comes back. This
adds the page — and fetches the editor from a CDN rather than carrying it in the
image.

## Why the editor is not in here

Only some HiSilicon and Goke parts serve raw at all, and this archive ships to
every board, so an editor in the image would be dead weight on most of them. It
lives in [OpenIPC/raw-editor](https://github.com/OpenIPC/raw-editor), pinned at
`v0.1.0`.

Same rule as the HEVC decoder, xterm.js and CodeMirror, and the same shape: a
pinned tag on jsDelivr, an `MJ_RAW_BASE` override, an 8s timeout, and a session
latch so a camera with no route out pays that timeout once rather than on every
press. Imported rather than run from a blob — a module can be imported
cross-origin, and the editor makes its own worker.

## What ships is useful on its own

A capture page: take a frame, keep it in the tab, hand it to the browser as a
file. The editor is an extra on top. So a camera with no way out shows an
ordinary working page and one sentence about the editor, rather than something
broken or visibly waiting.

Frames are never written to the camera — a raw frame is several megabytes and
the flash it would land on is the one holding the firmware.

## Four answers, not three

A build with no key for it; a build with it switched off; a build serving frames;
and **a request that never arrived**. The first version of this branch collapsed
the last into the first — `mjConfig()` answers `{}` for a failed fetch, so an
absent key and a dead request were the same object, and the page would have told
an operator their camera cannot do something it can. Review caught it; the check
now asks for the config directly so it can see a failure.

On unknown the page claims nothing and leaves capture offered: pressing it asks
the camera, which is a better answer than any guess made here.
`tests/raw-loader.test.js` pins all four, plus the load latch, the retry that
clears it, and the pin itself.

## No new CSS

The frame strip is built from classes the stylesheet already carries.
`./tools/regen-bootstrap-css.sh` produces no diff — a class invented here would
have needed the whole sheet rebuilt to survive the purge.

## Verified

`npm test` and `tools/lint-templates.sh` pass, and the purge regeneration is
clean.

Then end to end on a lab gk7205v300 with an IMX335, driving a real browser
against the running camera:

```
  ok  raw mode was read from the camera: slow
  ok  a frame was captured: 1.34 s
  ok  it appears in the session strip: 4.7 MB
  ok  the editor button became available
  ok  the editor loaded from the CDN and developed the frame: 1296x760
```

That camera runs firmware from before the DNG metadata work, so its files name
only the chip vendor rather than the sensor — which makes it the older-file path,
and it renders correctly.


widgetii:
Three of the four are real and are fixed in 423cec6. The fourth I think is a
misreading, and I would rather argue it than quietly ignore it.

## 1. Failed config fetch hides raw capture — right, and it undid a claim I made

This is the sharpest of the four. `mjConfig()` answers `{}` for a fetch that
failed, so `mjGet()` returned `undefined` whether the key was absent or the
request died, and the page would have rendered *"this firmware does not serve raw
frames"* at a camera that does — over one bad request on a slow link. That is the
flow the checklist names outright, and the commit that introduced it claimed in
its own message that the page kept those answers apart.

There are four states now: no key, switched off, on, and **nobody answered**. The
check asks for the config itself so it can see a failure, rather than going
through a helper whose whole job is to paper over one. On unknown the page claims
nothing and leaves capture offered — pressing it asks the camera directly.

The test deserves its share of the blame and has been rewritten. It stubbed
`mjConfig`, so the two cases it existed to separate were identical by
construction and it passed. It stubs `fetch` now: a rejection, a 500 and an
unparseable body are each distinct from a body that parsed with no key in it.

## 3. Unknown cameras can start capture — right

It shipped enabled and was clickable before anything was known. Disabled in the
markup, offered once there is an answer.

## 4. The retry button never retries — right, and neatly circular

The latch that stops a camera with no route out paying a timeout on every attempt
also defeated the one control whose entire purpose is to try again. An explicit
press clears it; an automatic path still does not.

## 2. Operators need a CDN to edit captures — dismissing, with reasons

The gate's three substantive tests all pass, and its own wording is what decides
two of them:

- *"a fetch of anything **the page needs in order to render**"* — the page renders
  completely without the editor. It is a capture page: take a frame, download it.
  That is the whole point of the shape.
- *"an endpoint **the project does not publish**"* — the gate answers this itself:
  *"'The project publishes it' is a question of who controls the bytes, not of
  whose name is in the hostname."* jsDelivr mirrors `OpenIPC/raw-editor` from its
  git tag. OpenIPC controls the bytes.
- *"leaves the page incomplete, broken or visibly waiting: an offline camera must
  look normal, not degraded"* — it looks normal. Capture and download work; one
  notice says the editor could not be fetched and why.

And the precedent is not hypothetical: `www/a/preview-wasm.js` does exactly this
today, for the HEVC decoder, and the gate was written with that file in the tree.
`console.cgi` and `files.js` do it for xterm.js and CodeMirror.

The recommended fix — commit the editor under `www/a/` — is the thing this design
exists to avoid. Only some HiSilicon and Goke parts serve `/image.dng` at all, and
this archive ships to every board, so a raw editor in the image would be dead
weight on most of them. The objective at the top of that very gate is *"the camera
serves every byte off its own flash, which on the smallest boards is the
difference between fitting and not"* — which is the argument **for** keeping it
out, not for pulling it in.

What the finding is right about is the consequence: editing does not work on a
closed network. That is deliberate and the page says so plainly rather than
hiding it, and `MJ_RAW_BASE` lets an operator host the module themselves.

PR #488, opened 2026-09-14
pins: a page that draws the chip and says which wire goes where

openipc-ai:
The page half of the new `/api/v1/pinmux` endpoint. A **Pins** leaf under System, for somebody who can solder to the board: it draws the part they are holding the iron to — the package, its tracks, its pads and the dot that says which way round it goes — from `/api/v1/pinmux`.

## Not the Day / Night pin map

`ircut-map.js` draws the banks the kernel reports as rows of tiles, and says at the top why it is not a package drawing:

> a package drawing would have to re-pitch itself per SoC, and on a BGA its pin numbers would be a fiction anyway

Both halves are true and both are answered rather than ignored:

- **Re-pitching.** One CSS custom property — the lead pitch — sets every other size, and `mj-pins.js` solves for it from the *measured column*, not a media query. The settings page insets its content behind the rail, so a viewport breakpoint sized the package for a width it never had: at a 1680px window the chip overflowed its column by about 200px. Measuring is the only thing right on every screen and on a column the chip does not choose.
- **The positions are a fiction**, so the page says so under the chip instead of implying a footprint. What this page has that the other never did is the pad *table*, so it draws the pads the part actually has, in the order it has them — including the ones an owner must never touch, blank, because a chip with pins missing is not a chip.

## Which wire

A family is not solderable. "Serial, 8 pins" does not say TX or RX; SDA is not SCL; and `I2C1_SDA` paired with `I2C2_SCL` is not a bus at all. So every choice is one signal and reads as **Serial 3 · RX**, with a line saying what that wire does, and each category names the buses, ports, slots or channels the chip has — because the pins of one are not interchangeable with the pins of another. The words are the ones printed on the module being wired up, never `UART1_RXD` or `SPI0_SDI`.

## Try it

Holds the change for thirty seconds; the page keeps it alive with a heartbeat so nobody watching sees it expire. Leaving the section puts the pads back — a window nobody is watching is exactly what it exists to end.

## One change outside the page

`mj-tree.js` gains a `custom` option: a section whose leaf draws something other than its fields. Rule 4 drops a section with nothing to draw, which is the right rule, and `pins` — one key, hidden — is the exception. Named rather than special-cased, so the leaf cannot silently disappear the day the daemon stops sending a visible key.

## Tested

`tests/pins.test.js` covers the two things that fail silently: dealing any pad count round four sides without losing one, and the pitch arithmetic that keeps the package inside its column (the first version used the top pad's dimensions for the side pads, which are turned end-on — that clipped the right-hand numbers off the drawing and nothing reported it). Plus the label composition, including a check that no composed label carries a datasheet spelling.

Sized and read on the lab hi3516ev300: 93 pads drawn, 50 offered, the sensor clock and the core voltage regulator among the blanks.

---

### Review, second pass

Four of the five findings were real and are fixed; each is in the commit message.

- **A private source path was named in a comment.** This repository is public and majestic is not, so the page now describes what the camera does behind `/api/v1/pinmux` rather than where it does it. The same reference has been taken out of this description.
- **Bare `fetch()` meant no session handling.** Now `apiFetch()`, like every other page: a 401 becomes the login redirect instead of being parsed as pin data, and the header the camera's cross-site check looks for is carried.
- **No rejection path.** `post()` parsed every reply as JSON without looking at the status and attached no catch, so a network failure left the countdown running and the bar claiming a trial that was not there.
- **A late trial could outlive the page.** `destroy()` could not cancel a try already in flight, and its callback started the heartbeat unconditionally — pads held open with nobody watching, which is the one state the window exists to end. A teardown flag is checked by every callback that could start or keep a window open, and the undo on teardown is now unconditional, because a try still in flight has not set `trying` yet.

The fifth — that pin edits bypass the settings form's dirty tracking, save and reset — is the design rather than a defect. A pin change is applied to the hardware and held open for thirty seconds before it is kept, which no settings form can express, so the leaf carries its own Undo / Try it / Keep it. The form's toolbar is `d-none` on a leaf with no dirty controls, so there is no dead Save button and no false unsaved-changes prompt. What the finding did reach is the empty state, which was a dead end; it now says what is still readable when the module cannot mount.


openipc-ai:
## Second pass

Four of the five findings were real and are fixed — each is in its commit message.

- **A private source path was named in a comment.** Gone, along with the private tracker reference in this description. The page now describes what the camera does behind `/api/v1/pinmux` rather than where it does it. I swept the whole diff, both commit messages and this description against the disclosure rule's failure criteria; nothing matches.
- **Bare `fetch()` meant no session handling.** Through `apiFetch()` now, like every other page: a 401 becomes the login redirect instead of being parsed as pin data, and it carries the header the camera's cross-site check looks for.
- **No rejection path.** `post()` parsed every reply as JSON without looking at the status and attached no catch, so a network failure left the countdown running and the bar claiming a trial that was not there.
- **A late trial could outlive the page.** `destroy()` could not cancel a try already in flight, and its callback started the heartbeat unconditionally — pads held open with nobody watching, which is the one state the window exists to end. A teardown flag is checked by every callback that could start or keep a window open, and the undo on teardown is unconditional now, because a try still in flight has not set `trying` yet.

**Declined: "Pin edits bypass settings safeguards."** Its evidence cites "Rules 24 and 25" of this repository's compliance checklist, requiring specialized controls to retain the replaced field as a hidden real field. `pr_compliance_checklist.yaml` has ten rules and neither of those is among them, so there is no such requirement to meet.

On the substance: a pin change is applied to the hardware and held open for thirty seconds before it is kept, which no settings form can express — so the leaf carries its own Undo / Try it / Keep it, which is a stronger safeguard than Save, not a weaker one. A hidden input mirroring `pins.assignments` would let the form's Save write the list without touching a register, leaving the file and the pads disagreeing: exactly the divergence the endpoint goes to some length to prevent. The form's toolbar is `d-none` on a leaf with no dirty controls, so there is no dead Save button and no false unsaved-changes prompt.

What the finding did reach is the empty state, which was a dead end; it now says what is still readable when the module cannot mount.

PR #490, opened 2026-09-14
pins: hovering a category shows where its pins are, which wire each is, and which bus

openipc-ai:
The Pins page could say what one pin is, one pin at a time. "Where do I solder my I²C sensor?" meant opening pads until you found one. Hovering a category now answers it on the drawing.

## What it does

Hover — or focus with the keyboard — a row in **What you can connect**, and the pins that could be it light up while the rest dim. A click holds it so the chip can be read with the pointer elsewhere, which is also the whole of the interaction on a touch screen; Escape lets go.

Three things then say what you need to solder:

- **Each lit pad shows the wire it would be** — `TX`, `SDA`, `MOSI`, `DATA3` — in place of its number.
- **The ring takes the colour of its bus.** On this chip's three I²C buses that is the only question that matters: eight pads are lit, and SDA on one bus wired to SCL on another is not a bus at all. Pins 3, 4, 48 and 49 all ring teal, so the pair on the left column and the pair on the top row are visibly one bus.
- **A legend at the foot of the category list, in the same colours** — one row per bus with each of its wires and the pins that offer it:

```
WHICH BUS, AND ITS WIRES
■ bus 0    SDA 46         SCL 47
■ bus 1    SDA 4 or 48    SCL 3 or 49
■ bus 2    SDA 50         SCL 51
```

## Three things that were harder than they look

**Nothing on the page may move while the pointer runs down the list.** The first version swapped the text of one paragraph above the chip, and the resting sentence wraps to two lines where most lit ones fit on one — so every row the pointer crossed shifted the drawing it was pointing at. That line now holds both sentences in one grid cell, the resting one hidden and carrying the height at whatever width the column has. Measured at 1680px: the package's top edge was at 221px lit and 241px at rest before; it is 260px in both now.

**"4 or 48", not "SDA 4, SDA 48".** `UART1_TXD` and `I2C1_SDA` each appear on two different pads. Listed per occurrence, port 1 looked like it had four wires where the chip has two — it is a choice of *pad*, and it reads that way.

**Every label stands off its pad.** A label flush against its pad is drawn over by any ring the pad wears: the SD card's `CLK` read as `CL` and `DATA3` as `TA3`, and — found while fixing that — the *selected* pin's number was struck through by its own selection ring, on every visit to this page, with nothing lit at all. One clearance constant now serves the number and the wire name, and the package is drawn smaller to make the room rather than overflowing into it.

## Not repainting

Each lead carries a class per family it can be and its bus's colour as a custom property, so lighting a category is one attribute on the wrapper and no lead is touched. A pointer crossing six rows would otherwise rebuild ninety-three buttons six times over, and a chip that answers *at a glance* cannot afford that.

## Tested

`padsFor`, `wiresFor`, `litSentence`, `busColour` and `pitchFor` are pure and exported, because a highlight that lights the wrong pins looks exactly like one that lights the right ones. 50 assertions cover which pins a family lights, the bus grouping including one wire on two pads, the colour cycle, the label composition, and the two arithmetic properties the no-jump layout depends on — that no lit sentence is longer than the resting one, and that the package always fits the width it is given.

2990 assertions green. Read on a lab hi3516ev300 at desktop width with every family in turn.


openipc-ai:
## Review — three fixed, and the first is the most serious thing in the feature

**"Shared pins show the wrong wire" — confirmed, and worse than the finding says.** It is not only that the bus was overwritten; on a real hi3516ev300's SD card the two offers are **different wires**:

| pin | slot 0 | slot 1 | what the drawing showed |
|---|---|---|---|
| 34 | `DATA0` | `DATA3` | **DATA3** |
| 35 | `DATA1` | `DATA2` | **DATA2** |
| 36 | `DATA2` | `DATA1` | **DATA1** |
| 37 | `DATA3` | `DATA0` | **DATA0** |

Anyone wiring slot 0 would have joined their card's DATA0 to the pad marked DATA3 and got nothing back, with nothing on screen to say why. Six pads on this part, and not making that mistake possible is the whole point of the page.

The drawing now claims only what is true of a pad **whatever choice is made**: the wire where every offer agrees on it, the bus colour where every offer agrees on that, and nothing where they do not. On the lab board, lighting SD card now gives:

- pads 32 and 33 keep `CLK` and `CMD` — the same wire on either slot — but take a neutral ring instead of one slot's colour, because the slot is what is in doubt;
- pads 34 to 37 keep their **numbers**;
- the rest, which offer one slot only, keep their wire and their slot's colour.

An ambiguous pad is resolved in the legend beside the chip, which lists every choice under its own bus — and is found there by the number the pad is still showing.

**"Missing pin data appears as zero"** — a family the camera reports no pins for is no longer lit at all. It was rendering "0 pins on this chip. They are the bright ones." with every lead dimmed behind it, which is a sentence about nothing.

**"Focused rows can lose their highlight"** — the pointer and the keyboard now have a slot each, so a pointer entering and leaving a second row no longer clears the highlight off a row that still has focus.

`padWire()` is pure and exported; eight assertions cover the pad that is two wires, the pad that is one wire on two buses, the ordinary pad, and the older parts' `I2C_SDA`, which has a wire and no bus at all. 2998 assertions green.

PR #491, opened 2026-09-14
checklist: name the five components that may come from a CDN

widgetii:
Rule 7 forbade any third-party asset outright — and **four things in the tree had always broken it**: the HEVC decoder, xterm.js, CodeMirror and the webfonts. The raw editor made five, and its review made the gap impossible to keep ignoring: the finding was right about the rule, and its suggested fix was to commit the editor into the image, which is exactly the cost a separate module exists to avoid.

So the exemption gets written down — as a **closed list of five named components**, not a shape. "Anything optional may come from a CDN" is a different rule and not this one.

| | component | loaded by | without it |
|---|---|---|---|
| 1 | HEVC decoder (`OpenIPC/hevc-wasm`) | `www/a/preview-wasm.js` | it is itself the last rung of the preview chain, reached only after the browser says `undecodable`; the chain ends one rung earlier |
| 2 | terminal (xterm.js) | `cgi-bin/console.cgi` | the one-command-at-a-time runner it ships with |
| 3 | file editor (CodeMirror) | `www/a/files.js` | a plain textarea, which still opens, edits and saves |
| 4 | raw editor (`OpenIPC/raw-editor`) | `www/a/raw-loader.js` | still captures a frame and hands it over as a file |
| 5 | webfonts (Montserrat, PT Mono) | `cgi-bin/p/header.cgi` | the system stack the CSS already names |

### The conditions are the part with teeth

Pinned to an exact version, never a moving tag. Fetched only when someone opens the thing that needs it. Never holding first paint. And the page must stay **usable** without it — degraded is allowed, blank and broken are not, and it has to say which function is missing rather than leave a control that silently does nothing.

A sixth fails the gate. Widening the list is an edit to this file, argued on its own.

### Checked, not assumed

Every claim in that table was read out of the code: `console.cgi`'s local runner, `files.js`'s `textarea` fallback, `raw.js`'s capture-and-download path, `header.cgi`'s `media="print"` trick that keeps fonts off the render path, and `preview-wasm.js`'s position in `preview-chain.js`.

Suite green, templates lint clean. No behaviour change — this is the checklist only.


widgetii:
Both findings on this head are declined, with reasons.

**1. "Webfonts load globally and can drift"** — this reads the file as it was before `93fbbf2`, which is the commit under review. The success criteria no longer apply pinning and feature-triggered loading to every exempt component; they apply to *the first four*, and (5) is carved out explicitly:

> (5) is a different shape and is held to a different condition, because it cannot meet those: the fonts are wanted on every page, and Google serves css2 as a moving resource by design, so there is no revision to pin. What stands in for pinning is that they can never block.

That was exactly this finding's earlier form, and it was correct then — the conditions I first wrote were ones the fonts could never meet. It is fixed.

**2. "Reviews now permit remote dependencies"** — this is an objection to the PR's purpose rather than to its execution. Rule 7 forbade any third-party asset, and four things in the tree had always broken it; this makes the exemption explicit, closed and conditional instead of leaving it undocumented. The alternative the finding recommends — bundle everything in firmware — is assessed in this bot's own summary on this PR as: *"Consumes scarce flash capacity on every camera. Ships large features to devices and operators that never use them. Undermines separately delivered optional modules such as the raw editor,"* with the recommendation *"Keep the closed, named allowlist."*

The narrowing that answers the underlying worry is already in the diff: a sixth component fails the gate, and widening the list is an edit to this file argued on its own.

PR #493, opened 2026-09-14
pins: say what the camera should do with an on/off pin, and what it is doing

openipc-ai:
Telling the camera a pad is an on/off signal wires it up and stops there.
That is half of what somebody soldering a module to the board needs: a
WiFi card's power switch is held off by a pull-up and comes on when the
pin is pulled **low**, and getting it there has meant a shell and a line
in `rc.local` that the next reflash takes away.

So the choice gains three sub-choices — **Don't drive it**, **Hold it
low**, **Hold it high** — and the page shows what each pin is actually at.

Needs a camera that offers levels. One that does not sends no list of
them, and then the page shows no control, reserves no space and claims
nothing.

## Two facts, two vocabularies

The control is imperative (*"Hold it high"*). The reading is declarative
(*"Right now it is high."*). They agree while the camera is holding a pin
and diverge when it is not — a pin left floating is whatever is soldered
to it.

If both rendered the same word with the same weight in the same 16rem
column, people would read the control as a readout and conclude the camera
was ignoring them. For the same reason the inert choice is not called
"Input": that single word is the whole difference between the two.

No volts anywhere — the page cannot know the rail — and no "pull-down",
"open-drain" or "tri-state". *"The camera pulls this pin down and holds it
there"* is what somebody says with an iron in their hand.

## Nothing is claimed that is not being carried

The whole first state arrives with the request the page already makes, and
changes arrive on a socket afterwards. The moment that socket closes,
every mark comes off the drawing and the line goes back to *"The camera is
not saying what this pin is doing right now."* A pin shown as low that is
actually high is a wiring decision made on a lie.

A reconnect is given a whole state rather than a delta against a list it
never received, and a level off the socket is vetted before it becomes a
DOM attribute.

## State model

`pending` entries become one object rather than a signal here and a level
in a second map beside it. Two maps break three things: a level-only edit
is absent from `pending` entirely, so the Keep button stays grey over a
real change; a pin taken to a bus and back leaves a stale level that Keep
would then drive onto a pad somebody has a relay on; and "set it back to
nothing" is one `null` here with no partner over there.

The live level is never written into `pending`. That map is what the owner
asked for; the pushed value is what the pin is. Keeping them apart makes
"a pushed level armed the Keep button" impossible rather than unlikely.

## Three details that are load-bearing

**The control does not call `paint()`.** The choice rows above it do, which
destroys the clicked button and drops focus to `<body>` — survivable for a
button, fatal for a control you navigate with the arrow keys, where every
press would rebuild the radios and the keyboard user could move exactly
once. It gets a narrow in-place refresh in the `applyLit()` spirit.

**The level mark rides `.mj-pin-trace`** — the one element in the drawing
carrying no state of its own — as colour and cross-axis thickness only.
`--mj-trace` is a term in the lead-stack sum `tests/pins.test.js`
reproduces by hand to check the package fits its column, so changing it
would let that test go on passing while the real chip overflowed.

**Stacked, not horizontal.** Three horizontal labels came to roughly 250px
against a 16rem detail column, and 15rem at the 1299px breakpoint. One
`.mj-seg-stack` modifier beside the existing rules: same radios, same
focus ring, laid on its side.

## Two pre-existing faults fixed on the way past

Named rather than folded in silently, because a reviewer would otherwise
read them as introduced here.

`.mj-pins-drives` said *"The camera drives this pin itself"* of **every**
on/off pin, when one the camera merely wires up is an input. The level is
what makes it visibly wrong.

And `used` covers two different things that were treated alike: a pad the
camera drives, and a pad the camera reports because **this page** claimed
it. Keeping a level made the pane undo itself — the heading reverted and
the control vanished, at the moment somebody is most likely to want it
again.

## Verified

191 lines of new tests, and the whole suite green. The pure layer is
exported so the rules are held without a DOM: a level belongs to an on/off
pin and nothing else; a level-only edit arms the bar; a bus row carries no
level whatever the edit says; an on/off row carries its level even when it
is the inert one; and the drawing claims nothing unless the feed is live.

Checked in a browser against a real hi3516ev300 at 390 / 1280 / 2560 —
`tools/check-layout.mjs` clean at all three, its self-test run first.
Confirmed by hand that picking a level keeps focus on the radio, that the
bar arms, that the mark appears on the drawing, and that the readout says
what the pin is.


openipc-ai:
Both taken. Pushed as 466a9fa.

**#1 was a functional break, not just a display one.** The page sends the
whole list on every save and the camera refuses a level word it does not
recognise — so one unrecognised value sitting in a row nobody was editing
would have the *entire* save refused, and the owner could not then change
any pin at all. What goes back is now always a level this build knows,
which is what the camera does with an absent one anyway. Tested, including
the camera-offers-no-levels case, where no level member is sent at all.

**#2** is right about the direction of the race: `refresh()` is called
after a keep and after a reconnect, and a delta landing between the
request going out and the reply coming back is *newer* than the reply. The
older snapshot overwrote it and those pins showed a stale level with no
later delta bound to correct them. The socket is the authority once it is
live; the snapshot only seeds when it is not.

openipc-ai:
Right to re-raise it — I fixed the half about what gets **sent** and left
the half about what gets **shown**. Pushed.

`normLevel()` coerced a saved level this page cannot offer to the inert
one, so a pin a newer camera was doing something else with rendered with
"Don't drive it" ticked and "The camera reads this pin and drives nothing
onto it" underneath. That is the page claiming to know a setting it does
not, and somebody could wire to a pad on the strength of it. An
unrecognised level is now carried through untouched, nothing is ticked,
and the pane says the camera is doing something this page does not know
about.

One half I'm declining, with reasoning rather than silence: **a legacy row
keeping Keep disabled is correct.** An absent level and an explicit
`float` are not two states — the camera reads a row with no level exactly
as "don't drive it", and applies the same thing for both. So ticking
"Don't drive it" on a legacy row genuinely changes nothing, and arming
the bar would offer a save that would alter no behaviour. The row becomes
explicit anyway the next time anything on that page is saved.

The related claim that unsupported values are "sent back unchanged on
unrelated saves" is also no longer true as stated: a value the camera
advertises is deliberately preserved, and only something neither side
knows is rewritten — otherwise one unrecognised value in a row nobody was
editing would have the whole save refused.

PR #494, opened 2026-09-14
raw: let Calibrate write to the camera, and take it back

widgetii:
The editor gained a Calibrate mode that measures a colour chart and solves both matrices (pinned here at **raw-editor v0.5.0**). This is the half that reaches the camera.

`isp.colorMatrix` and `isp.dngColorMatrix` are written **together**, because one measurement produces both and neither is derived from the other — one drives the live picture, the other is what a RAW snapshot carries.

### What makes it safe is reading the keys before writing them

A matrix that ruins the picture also ruins the view you would use to notice it, and the setting survives a reboot, so the camera would come back still wrong. The editor runs the clock and asks the question; this supplies the verbs and remembers what to go back to.

- **Reverting is the default**, confirming is the exception.
- `keep()` is what stands the unload handler down. Without it, confirming left it armed and closing the page later undid a calibration that had been deliberately kept.

### Verified on a lab gk7205v300, from the camera's side

Polling `/etc/majestic.yaml` across a full cycle:

```
 6  colorMatrix: ""
10  colorMatrix: 1.3449 0.2089 -0.5539 -0.8092 2.1432 -0.3340 ...
 1  colorMatrix: ""
```

Applied, held for the countdown, and put back with nobody touching it. A confirmed run instead leaves the matrix in place and reports "Kept".

### One limitation, stated rather than papered over

A tab that disappears mid-countdown gets a best-effort `keepalive` request that outlives the page. Nothing guarantees it arrives. That is why the editor asks for confirmation rather than treating this as the safety net — and why a camera-side lease would be the real fix if this ever needs to be guaranteed.


widgetii:
Three findings left; one taken, two declined with evidence.

**"Readers inherit a transient review link"** — fair. The description cited a pull request to say where Calibrate came from, which is a review artifact rather than a durable fact. Rewritten to name the released version (v0.5.0) instead.

**"Paired matrices can save separately"** — not so, and I checked on the camera rather than reasoning about it. `/api/v1/set` takes both keys in one request and applies both:

```
after a two-key request:
    colorMatrix: 1 0 0 0 1 0 0 0 1
    dngColorMatrix: 2 0 0 0 2 0 0 0 2
```

**"Revert leaves absent settings configured"** — true as stated, and it makes no difference. Restoring an unset key writes `colorMatrix: ""` rather than removing the line. majestic treats the two identically — `src/hisi/scene.c` reads `spec == NULL || *spec == 0` and returns the ISP to automatic for either — so the camera ends in the state it started in. The residue is a line in the YAML, not a setting.

The one that mattered, **"Confirmed calibration is undone on exit"**, was real and is fixed on both sides: `keep()` in raw-editor v0.5.0, and here it clears the remembered previous value, which is the single thing the unload handler reads.

widgetii:
Correcting myself: I declined findings 1 and 3 and was wrong about both. This repo documents the answer and I had not read it.

**"Paired matrices can save separately"** — right. `POST /api/v1/config` is the batch write; CLAUDE.md: *"walks and sets every leaf, then reloads the SDK and writes the file back exactly once. Aborts on first rejected leaf... no persistence partial-credit"*, while `/api/v1/set` is the single-key variant *"not currently called by the WebUI"*. My defence — that a two-key query string works on the camera — tested that it **functions**, not that it follows the contract. Two matrices that must agree should not go as two writes.

**"Revert leaves absent settings configured"** — also right, and it is not cosmetic the way I claimed. CLAUDE.md spends a paragraph on it: *"A `null` leaf REMOVES its key... which is the only way to put an optional setting back the way it was found"*, with the incident behind it — a cleared integer storing 0, which is a real GPIO. Restoring a setting the way it was found is this feature's entire safety net, so the distinction is the point.

Both fixed: one `POST /api/v1/config` carrying both leaves, `readKeys` answering `null` for an absent key so a revert removes it again, and a re-read afterwards because older majestic answers 202 and ignores nulls.

Verified on the camera — both keys land in one POST, confirmed in the file and in `config.json`, and a null pair removes them and the `isp:` section with them.

PR #497, opened 2026-09-15
ircut: the picture may not speak from a frame with no light in it (#492)

### LINKED ISSUE #492: Dashboard: IR-cut warning cannot be dismissed
The camera has its illumination disabled. At night, the image is completely black.

However, the Dashboard permanently displays the following warning, with no close/dismiss button:

> ⚠️ **The picture looks like an open IR-cut filter** — Recent frames have the magenta cast of infrared reaching the sensor while the camera is in day mode. The pins are configured, so run the IR-cut test to see whether the filter actually moves. A scene lit by an IR lamp, or by magenta light, looks the same from here. [ **Open day/night** ]

In this configuration, the warning is not necessarily actionable. The camera can simply be operating without illumination, and the user may not want to run an IR-cut test or change the day/night state.

Should this warning really be non-dismissible?

I would expect such a diagnostic notification to have a close/dismiss option, so that the user can acknowledge it without being forced to keep it permanently visible on the Dashboard.

**SOC**: ssc378de (infinity6c)
**SENSOR**: imx415
**FIRMWARE**: 2.6.09.14-lite
**BUILD**: master+d71a47b, 2026-09-14
**MAJESTIC**: Lite SigmaStar (infinity6c), master+98b8049, 2026-09-13 17:28
**WEBUI**: master+8c6a6e3, 2026-09-14 20:16
**U-BOOT**: U-Boot 2015.01 (Oct 03 2024 - 22:45:04)

openipc-ai:
Closes #492.

A camera with its illumination disabled shows a black picture at night, and the Dashboard warned — all night, with nothing to press — that the IR-cut filter looked open. The reporter asked for a dismiss button. The warning was wrong instead.

### The banner is not stuck

Reproduced the true positive first, on an hi3516ev300 in daylight: filter closed reads `gmin 0.016 / mex25 -0.086`, and the Dashboard shows nothing; drive the filter open on the same scene and it reads `gmin 1.000 / mex25 +0.512`, and the banner appears within 20 s. So it is being told something.

### Both statistics drift toward "open filter" as the light goes away

* The magenta excess is normalised by each pixel's own brightness, so a **fixed** channel offset — black-level residue with the white-balance gains on top of it — says something different at each end of the range. Three counts between `(R+B)/2` and green clear the 0.15 bar at a luminance of 20; at 80 it takes twelve. The test is easiest to trip exactly where it is least trustworthy.
* "Green is the minimum channel" needs only red and blue to carry the larger white-balance gain, which in a frame made of amplified noise is every pixel. Measured on one daylight frame from that camera **with the filter closed**: green is the valley in 0.001 of the pixels above luminance 96, and in **0.426** of those below 16 — the same frame, either end of its own histogram.

### The floor that was supposed to stop this was a count

`n >= 200` usable pixels, against a sample of 160×90 = **14,400**. So 1.4 % of a frame could speak for all of it — the thing the floor's own comment said must never happen. The test that pinned it used a 20×20 frame, where 200 was half of it, and the mismatch was invisible from there.

It is a share of the frame now: **half the sample has to be genuinely lit** before either statistic may answer.

### Why this is not the brightness gate this file refuses to make

That one would ask the scene how dark it is, and auto-exposure makes it unanswerable — a correctly exposed midnight frame and a correctly exposed noon frame have the same mean by construction. Auto-exposure is exactly what makes the inverse question answerable: it lifts any scene it can reach to its target, so a frame *still* under the floor is one the camera could not lift, and that is the frame whose colour is its own noise. Measured while covering a lab camera's lens: the gain pinned at its maximum, `isp_exposureismax 1`, and the picture came back at mean luminance 118 with a lit share of 1.00. Light it can reach, it reaches.

### What is unchanged

Lit share measured across daylight and indoor scenes on three lab cameras: 0.62, 0.89, 0.92, 0.99 — and 0.99 for an open filter in daylight. Verified end to end through a real browser on the camera: filter closed, no banner; driven open in daylight, the banner within 20 s at `gmin 1.000 / mex25 +0.496`; restored, gone.

The pin sweep is untouched — it reads `gmin` deltas off the raw statistics and never asked `look()`. What the passive check now refuses is what the active **Test** has refused from the start: *"Not enough light to tell"*.

### Tests

`tests/ircut-check.test.js` pins the regression with a noise-floor pixel, `[24, 18, 22]` — `gmin 1.000`, `mex25 +0.234`, an open filter under the old rule and nothing at all under the new one — plus the share floor at a tenth of the frame (refused) and at half (answered). The floor test it replaces asserted a rule that could not hold at the sample size it ships with.

### Confirmed by the reporter

Asked on the issue whether the banner also stands in daylight, the answer was: *"In complete darkness, the image does have a slight purple/magenta cast. The warning does not appear during the day."* That is this fault exactly — a cast that is real, measured in the one condition where the statistics cannot support a conclusion from it, and absent in the condition where they can. His filter is working.

### Out of scope, filed separately

The case a lit scene can still produce — a working filter under an IR floodlight, where the picture really is magenta in daylight and the warning is correct and useless — is not this change. The IR-cut **Test** adjudicates precisely that question and its verdict is in-page state, so the Dashboard goes on accusing a camera that has just been measured and cleared. Filed as #498.


openipc-ai:
qodo — pushing back on the Rule 6 finding.

Rule 6's success criterion is *"Verification is described by hardware class and configuration — 'measured on an hi3516ev300 + imx335, main stream 1280x720' — which is what makes a result reproducible"*, and its failure criterion lists a hostname or domain, a private IP, a credential, an SSH alias, a private branch or commit id, or pasted console output and screenshots.

The flagged comment is the first of those, not the last: a hardware class, a configuration (filter closed), and the number the threshold was chosen against, in prose. It carries no address, no credential and nothing that identifies a machine, and the measurement is exactly what makes the constant reviewable — a reader who disagrees with the floor needs to know what a real frame does at either end of its histogram, or the number is a magic one.

It is also the file's established form. `tests/ircut-check.test.js` has recorded *"gmin 1.000 open against 0.03-0.07 closed, magenta excess p25 +0.40 against -0.11"* since the feature landed, and says in the same breath why the frames themselves are not committed: *"the pixels are a picture of somebody's window and this repository is public."* That is the line this rule is drawing, and the change keeps to it — numbers about a scene, never the scene.

The PR description is the same content in the same form: measured values named by what was measured, with no host, no credential and no pasted log.

Leaving the comment as written. If the rule is meant to reach measured values as well, that is a change to the rule rather than to this diff, and it would take the existing comments in this file with it.

PR #499, opened 2026-09-15
ircut: a test answers the picture, and the Dashboard hears the verdict (#498)

### LINKED ISSUE #498: The IR-cut test adjudicates the picture, and the Dashboard never hears the verdict
Split out of #492, where it is named but deliberately not fixed.

The Dashboard's IR-cut finding says, of a day-mode frame with a magenta cast:

> The pins are configured, so run the IR-cut test to see whether the filter actually moves. A scene lit by an IR lamp, or by magenta light, looks the same from here.

So it sends the operator to the Test, and the Test answers the question properly — it drives the filter and compares the two pictures. But its verdict lives in page state on the Day & Night section (`state.ircutTestedOn`, dropped by `syncVerdict()` as soon as the pin assignment it was measured against changes) and is gone on the next page load. The Dashboard never sees it.

The result is a loop with no exit for a camera whose filter is fine and whose scene is genuinely magenta — an IR floodlight in view is the ordinary way to get there:

1. The banner says to run the Test.
2. The Test says *"The IR-cut filter is wired correctly"*.
3. The Dashboard keeps the banner, because it recomputes from the picture every five seconds and knows nothing about step 2.

There is no dismissal, and that is right — #367 removed the one that existed because it stored a claim about the camera where only one banner could read it, and nothing could invalidate it at the moment the wiring changed. The answer here is not to put that back. It is that a measurement the operator has already made should settle the accusation the picture made, and be dropped again by exactly the thing that makes it stale — the wiring it was measured against, which is the rule `syncVerdict()` already implements in the one page that has the verdict.

What needs deciding is where a verdict lives so both pages can read it, and what witnesses its staleness from the Dashboard's side. Not urgent: it is unreachable for a camera whose filter is genuinely open (there the banner is correct and the fix is the wiring), and the reporter of #492 confirmed his own case is the night misfire, which #497 fixes.

openipc-ai:
Closes #498.

The Dashboard's picture finding says, of a magenta day-mode frame, that the pins are configured and the IR-cut **Test** is what settles it. The Test does settle it — it drives the filter and compares the two pictures — and then the verdict lives in page state on the settings section and is gone on the next load. The Dashboard never hears it.

So a camera whose filter demonstrably moves and whose scene is genuinely magenta is told to run a test, runs it, passes it, and keeps the banner. That loop with no exit is what sent the reporter of #492 looking for a dismiss button.

### The rule

A verdict of *"the filter moves correctly"* stops the picture being asked about, while the wiring it was measured against still stands. Only that verdict: one that found a fault leaves the finding alone, because the fault is the news and it is not the picture rule's to report.

### Why this is not the dismissal #367 removed

The difference is where the invalidation lives. That one recorded an **opinion** — "no filter here" — which had to be *dropped* when the wiring contradicted it: an event, needing a witness, on a page where the event never happens. This is a **measurement carrying the wiring it was taken against**, compared against the camera's own configuration every time it is read. There is no moment to catch and nothing to keep in step; a record that no longer describes the camera stops counting, on the next paint, in whatever page is reading it.

The four settings it is stamped with are the four the test's own blocker already guards — the two pads, the single-pad invert, and whether the daemon drives the filter at all. Moving the daylight sensor or the lamp changes nothing about what a pulse does, and a key that fires on them throws away a good measurement for an unrelated edit.

### Why it does not expire

An argument rather than an omission. The case it silences is a picture that looks open on a camera whose filter demonstrably moves — magenta light in the scene — and there the picture has *already* stopped being able to tell a later fault from the scene. Expiry would buy no detection back, only a question re-asked on a timer.

### Two refusals to remember

* A run that **could not finish** is half a measurement, and half a measurement is not evidence about the filter.
* A run whose **restore failed** is not an answer whatever it found about the wiring: the filter is sitting where the test left it, so if that is open, the picture is magenta because the filter really *is* open — and remembering "fine" would silence the one accusation that is true.

### Scope

Per browser, which is the honest limit of it: the fact is a measurement made there, and the only camera-side place to put it would be a second store of the kind #367 deleted. Unreadable storage reads as "never tested", so a browser that cannot remember goes on asking rather than going quiet. A record of the wrong shape — a hand edit, an older key — is refused rather than compared field by field against `undefined`.

### Verified end to end

On an hi3516ev300 in daylight, one browser, filter actually moving:

| step | result |
|---|---|
| filter driven open, no test on record | banner up within 35 s |
| Test pressed on the settings page | *"The IR-cut filter is wired correctly"*, recorded as `{"id":"ok","wiring":"11/10/0/1"}` |
| back to the Dashboard | silent |
| record rewritten to describe other wiring | banner straight back |

Unit tests cover the rule itself: each of the four settings invalidating a verdict, the daylight sensor not, every fault verdict failing to silence, malformed records, and that a settled verdict silences the picture finding **only** — a camera that cannot move its filter at all still gets its missing-pin finding.


openipc-ai:
Both findings taken, and both were real. Fixed in 3be7956.

**An interrupted test can hide a fault.** Right: the store was cleared in the completion handlers, so between the first toggle and the last there was a window in which the record described a filter that was being driven — and if the page went away inside it, neither handler ran and a "fine" survived a run that may have left the filter open. It is forgotten up front now, before anything moves, and written again at the end if the end agrees.

**Dashboard keeps a stale test verdict.** Also right, and it undercut this change's own argument, which is that the record is compared against the camera's configuration *every time it is read*. `mjConfig()` resolves once and is cached for the life of the page. That is fine for a finding that merely describes a snapshot and not fine for one a record can silence. The wiring is now re-read on a slow beat of its own — and only while a record exists, since a browser that has never tested has nothing that could be silenced and should pay nothing.

One deliberate difference from the recommended fix: a refresh that **fails** leaves the last reading standing rather than making the finding visible again. A failed fetch is not a fact — it is not evidence that the wiring changed — and a Dashboard that cannot reach the camera is already saying so through its stale banner. Treating it as a change would let a flaky link resurrect the banner on a camera nobody has touched.

Verified on an hi3516ev300, both paths:

```
A. before pressing Test: {"id":"ok","wiring":"11/10/0/1","at":1}
   during the run:        null
   verdict: The IR-cut filter is wired correctly …
   after the run:         {"id":"ok","wiring":"11/10/0/1","at":1789464624}

B. banner with a matching verdict: false (expect false)
   swapping the coils through /api/v1/config, Dashboard left open…
   banner after the swap:          true  (expect true)
```

PR #500, opened 2026-09-15
sdcard: say whether the card is fast enough, not only whether it is there

widgetii:
The other half of OpenIPC/firmware#1747 — a card delivering a fraction of what
the vendor firmware managed, open since March 2025 with two guesses on it and
not one number, because nothing on the camera would produce one. The answer it
got was "have you tried another sdcard?", which is the only answer available
when the camera can say nothing about the card it has.

Every check this page already made asks whether the card is **there** — mounted,
readable, writable, with room on it. A card that answers yes to all four can
still be discarding clips because it cannot take the video in time, and for that
camera the page was green.

## Three things say so now

**What the card says it is.** The SD Status register carries the speed class,
UHS grade, video class and app-performance class, and the kernel has already
read it at probe time. This matters more than it looks: C10, U1 and V30 are all
measured on one long uninterrupted write, and **A1/A2 is the only mark that
promises anything about the small scattered I/O a recorder does while deleting
old clips underneath itself**. So a card can carry every sequential mark there is
and still be the reason footage is being lost — and nothing the camera could show
its owner separated the two, which is what made "use a faster card" advice nobody
could act on.

Every field is checked against the set the spec defines and the whole register is
dropped if any one falls outside it. A rating decoded wrongly and printed as fact
is worse than no rating.

**What it is actually doing.** Achieved write rate, longest flush, footage
discarded, write errors and queue depth — all off the `/metrics` heartbeat every
page already runs, so this adds **no request**. Each figure is gated on having
been read: a counter this majestic does not publish is left out rather than drawn
as a zero, because a zero here reads as "nothing has gone wrong".

**What it can do, for a camera that is not recording yet.** Those counters are
silent until recording has worked at least once, which is not the state somebody
is in when they come looking — it is exactly #1747's state. *Measure this card*
writes 32 MB in flushed chunks, times them, and reads them back cold.

Bounded three ways, because each bound is a failure somebody would otherwise wait
through: a chunk count caps the work, a 20-second budget stops a slow card
between chunks, and a ten-second per-chunk timeout stops a wedged one *inside* a
chunk. It refuses an unmounted or read-only card; refuses to run when under 64 MB
is free rather than make room by deleting footage; uses a per-run filename so a
file an operator keeps on the card is never what gets deleted; and holds a
camera-wide lock released on every exit path including a kill.

## The banner could already say this and never did

`storage-verdict.js` has carried the sentence *"cannot keep up — footage is being
lost"* since it was written. Only the Recordings page ever passed the evidence for
it; `storage-check.js` passed `''`. So on every other page a card that was
mounted, writable and reporting itself healthy while silently discarding clips
drew nothing at all. One argument.

It is also gated on the recorder still writing. The verdict is a claim in the
present tense, and recording switched off while the page is open would otherwise
leave a loss on screen that is no longer happening, with nothing able to clear it.

One formatter moves into `storage-verdict.js` with it. The timeline's rounds to
whole seconds, so a window of sub-second losses printed "0 s of video has been
dropped" — a sentence that reports a loss and then denies it.

## Verified on four boards, four cards, three SoC vendors

| hardware | kernel | SD Status register | card reports | measured write |
| --- | --- | --- | --- | --- |
| hi3516av300 + imx415, 4K main stream | 4.9.37 | present | C10 · U1 · V10 · no A | 9.0–18.6 MB/s |
| hi3518ev200 + jxf22 | 4.9.37 | present | C10 only | 4.7 MB/s |
| Ingenic T31 + sc2332 | 3.10.14 | **absent** | *this camera cannot read them* | 8.0 MB/s |
| hi3516ev200 + sc2315e | 4.9.37 | n/a — empty slot | — | refused, as it should |

Two of those boards changed the code.

**The Ingenic T31 has no such register.** Its MMC driver exposes `cid`, `csd` and
`scr` but not the SD Status, and `scr` is no substitute — speed class and A1/A2
live only there. It is not a kernel-version split either: both HiSilicon boards
are 4.9.37 and one Ingenic 3.10 is not. That board's slot holds a 2021 SanDisk
64 GB, a card that is almost certainly A1-rated, and the page told me *"this card
does not report its speed ratings"* — blaming the card for what only the camera
lacked. "No rating" now has three answers: the platform cannot read it, the
register would not decode, or the slot holds no SD card at all.

**The hi3518ev200 measured 4.7 MB/s against a Class 10 badge promising 10.** The
tempting conclusion is that the card is fake. It is not: a raw sequential read
straight off the block device, with no filesystem involved, manages the same
4.7 MB/s, while the CPU copies 40 MB/s. The host controller on a 440 MHz ARM926
is the ceiling. So the shortfall is pointed out and the cause is left open — the
rated floors are quoted for a card reader, and on some cameras the slot is the
slower half.

## Tests

`tests/sdcard-perf.test.js`, 37 assertions over the branch table: the rating
cases, the live figures, sub-second rounding, the three ways to have no figures
(a poll that has not landed, a build with no such counters, a camera that has
recorded nothing — identical once they reach markup, and only one may be turned
into a sentence), a duration that was never read, a read that did not finish, and
the contention caption.

Every one was checked against a deliberately broken version rather than only
against the fix. The first attempt at the "unknown recorder" case did **not** fail
when the unknown was collapsed into a known absence, which is the distinction the
whole three-state shape exists for; the assertion was rewritten until it did.

`npm test` and `tools/lint-templates.sh` pass, and a fresh
`tools/regen-bootstrap-css.sh` reproduces the committed stylesheet byte for byte.


widgetii:
## Checked on two more boards, and one of them found a bug

Verified across three SoC generations. The measurement half ported unchanged;
the **ratings** half did not, and the way it failed was worse than not working.

| hardware | kernel | SD Status register |
| --- | --- | --- |
| hi3516av300 + imx415 | 4.9.37 | present, decodes |
| Ingenic T31 + sc2332 | 3.10.14 | **does not exist** |
| hi3516ev200 + sc2315e | 4.9.37 | controller up, empty slot |

It is **not** a kernel-version split — both HiSilicon boards are 4.9.37 and one
has the attribute, the Ingenic 3.10 does not. That driver exposes `cid`, `csd`
and `scr` and no SD Status. `scr` is not a fallback: it carries SD_SPEC and bus
widths, while speed class and A1/A2 live only in the SD Status. There is no other
way to ask.

### The bug

The T31's slot holds a 2021 SanDisk 64 GB — a card that almost certainly *is*
A1-rated. The panel told me **"this card does not report its speed ratings"**,
blaming the card for something only the camera is short of. That is exactly the
kind of false accusation the comments in this endpoint already warn about, and no
amount of testing on the first board would ever have shown it.

So "no rating" now has three answers rather than one:

- **platform** — the attribute does not exist here. Says *"this camera cannot
  read card speed ratings"*, about the camera.
- **unreadable** — the attribute is there and did not decode. Blames neither.
- **notsd** — an SD structure asked of something that is not an SD card. No row
  at all.

Rendered from each board's own endpoint answer, the Rated row reads
`Class 10 · U1 · V10 · no A1/A2 rating [sequential only]` on the hi3516av300,
`this camera cannot read card speed ratings` on the T31, and is not drawn at all
on the board with an empty slot.

Three test cases cover it, and collapsing the platform case back to the old
sentence fails three assertions.

### Everything else ported unchanged

The flushed-chunk write, the monotonic clock it is timed against and the refusal
paths all behave the same under Ingenic 3.10 as under HiSilicon 4.9. The T31's
SanDisk measured 8.0 MB/s write and 15.8 MB/s read with a 0.29 s worst chunk; the
board with no card in it refused with *"the card is not mounted, so there is
nothing to measure"* rather than failing in some more interesting way.

### How the seam was checked

Rather than deploy the JS to each board, each board's own endpoint answer was fed
through `www/a/sdcard.js` in a node `vm`. That proves the backend-to-frontend
composition on three platforms with nothing installed on any of them, and every
board was left with no scratch file, no lock and no overlay copy over the
firmware's own files.

widgetii:
## A fourth board, and it wanted to accuse an innocent card

`hi3518ev200 + jxf22`, kernel 4.9.37, SD Status register present — so three of
four boards carry it, only the Ingenic 3.10 without.

Its card decoded to a shape none of the other three had: **Class 10 and nothing
else** — no UHS grade, no video class, no app class. A 2012-dated 64 GB card.
That is the first time on hardware that the omit-zero-grades path has been
exercised, and the Rated row reads `Class 10 · no A1/A2 rating [sequential only]`.

### It measured 4.7 MB/s against a badge promising 10

The obvious reading is "this card is fake". Before letting the page imply it, two
controls: a raw sequential read straight off the block device, with no filesystem
involved, manages the **same 4.7 MB/s** — cold and warm alike — while the CPU
copies 40 MB/s to `/dev/null`.

The host controller on that 440 MHz ARM926 is the ceiling, not the card. Printed
as a bare comparison, that figure sends somebody to buy a replacement and measure
exactly the same number again — the same false accusation as blaming a card for a
register its kernel does not expose, arriving from the other direction.

So the shortfall is pointed out and the cause is left open:

> That is below the 10 MB/s this card's markings promise — but the rating is
> quoted for a card reader, and on some cameras the slot itself is the slower
> half. A figure under the rating does not on its own mean the card is at fault.

Three cases cover it — shortfall, comfortably meeting the rating, and no rating to
compare against. Deleting the note fails three of them; rewording it to blame the
card fails two.

### The tests now press the real button

Worth noting because the first attempt was wrong in an instructive way: it
assigned to the page's own `speed` variable from outside the IIFE, which cannot
work — and had it worked it would have been asserting against a planted fixture
rather than against the code. The element stub now keeps its delegated listeners
and the POST stub answers with the measurement, so the page's own handler is what
runs.

widgetii:
Eight of the nine addressed in `439f496`; one rejected with evidence.

**Fixed**

- *Clock failures report infinite speed* — the endpoint no longer substitutes the
  opening timestamp for a closing read it could not make, and the page gates
  every figure on the pair it is made of rather than trusting the endpoint to
  have done so. A test plants a zero duration and fails if anything but silence
  comes out; reverting the guard reproduces `Infinity MB/s` exactly as described.
- *Failed reads show a card speed* — `dd`'s status is checked now, and a
  read-back that did not finish reports that instead of being divided into the
  full byte count. Reverting that guard prints `-32000.0 MB/s`. A write that
  fails part-way is also no longer indistinguishable from the budget expiring.
- *Slow cards can hang the measurement* — each chunk now carries its own
  ten-second timeout, so the bound no longer waits for a flush to return. A card
  that cannot take 2 MB in ten seconds is reported as having stopped accepting
  data, which is a different answer from "slow" and worth saying as one.
- *Interrupted tests block future tests* — released on every exit path via a
  trap, and the lock carries the time it was taken so a run that died is
  reclaimed rather than disabling the feature for good.
- *Existing card files can be destroyed* — the scratch file is named per run.
- *Speed tests fork an avoidable relay* — correct, and it was a compliance
  failure as well as a cost. Both `mj_cfg` calls and the `recording` field are
  gone; where the clips are configured to go and whether bytes are moving are
  both already in the browser, so the answer is derived there.
- *Disabled recording still shows warnings* — the banner is gated on the
  recorder still writing. Suppressed rather than forgotten, so a recorder that
  resumes reports what it loses from then on instead of re-announcing the old
  delta.
- *Lab camera output becomes public* — the comment now names hardware class and
  configuration, which is what the checklist asks for. The description and both
  my comments have been rewritten to drop pasted device output and build
  identifiers; the commits and the diff were already clean.

**Not accepted: "Private tracker details become public"**

The finding says citing `OpenIPC/firmware#1747` exposes a sibling repository and
its private tracker number. That repository is public — `gh repo view` reports
`isPrivate: false, visibility: PUBLIC`, and the issue has public commenters on
it. Rule 1 forbids references into the *private* source repository; this is a
reference into a public tracker, which the rule's own success criteria allow and
which `CLAUDE.md` positively asks for: *"Cite the issue that reported the fault
— a document that stays readable and says what the code cannot."* Removing it
would leave four comments explaining a decision with nothing a reader could
check. The citation stays.

**One the review did not find, and hardware did**

The fix for the lock finding introduced a new bug: the trap called `rmdir`, which
refuses a directory that is not empty, and the lock holds its timestamp inside
itself. Every successful run leaked its lock and the next measurement within the
stale window was refused as "already running" — silently, since `rmdir`'s failure
was discarded. It is `rm -r` now, verified on a camera: a normal run releases, an
immediate second run succeeds, and a run killed mid-flush releases both the lock
and the scratch file.

PR #503, opened 2026-09-15
ircut: the filter can be told not to follow day/night, and the page says so

openipc-ai:
The web-interface half of the reporter's ask on #492, whose daemon half landed as majestic's `nightMode.irCutAuto` / `backlightAuto`: off leaves that actuator out of the night transition while the pad stays claimed, held, and available to a hand on the switch. Three places here assumed an actuator always follows day/night.

### The Live panel hid the whole group

While Automatic day/night is on, the runtime switches are hidden and one sentence names what has the wheel — three dead switches say less than that. Still right where it is true. Where an actuator has been told not to follow, **its switch is the only way to move it**, so hiding it would leave the configuration this feature exists for with no way to use it. The group stays, the monitor's own switches are the dead ones in it and each says why, and the sentence names what the monitor actually drives instead of reciting all three.

### The night toggle predicted the other two

It moved the IR-cut and lamp switches with the mode, because the mode used to move them. It asks the camera now — the same gauges the panel reads at mount. An actuator that does not follow does not move, and predicting that it did is a switch that lies about where the filter is.

### The Dashboard's conflict finding convicted on the new state

*"Automatic day/night says it is night, but the filter has been in the day position for 600s"* is a fault when the monitor drives both, and is the setting working when it does not — permanently, by design. It stands down for a filter that does not follow, and only it: `hunting` is untouched, because a camera flapping between day and night is flapping whatever the filter does.

Absent reads as *following* throughout — that is the default, and every camera before these keys did.

### Measured, in a browser

On an hi3516ev300. That camera's daemon predates the keys, so the config it serves is patched on the way into the page; everything after that is the real page, the real code path and the real layout.

| configuration | night | IR-cut | lamp | note |
|---|---|---|---|---|
| monitor off | live | live | live | — |
| monitor on, both follow (today) | hidden | hidden | hidden | driving night, IR‑cut and the lamp |
| monitor on, filter does not follow | greyed | **live** | greyed | driving night and the lamp |
| monitor on, neither follows | greyed | **live** | **live** | driving night |

The greyed ones carry *"Automatic day/night is driving this one."* Row two is today's behaviour, unchanged.

`tests/ircut-check.test.js` pins the finding: a non-following filter may disagree with the mode, an absent or explicitly true switch still means it follows, and the hunting finding is not caught by the same rule.

The two new switches need nothing to appear on the settings page — it is generated from the daemon's schema, and they render beside the park switches they belong with.


openipc-ai:
All five taken — each was a way for the page to state something the camera had not said. Fixed in 2bf1696.

**Manual actuator switches disappear / Manual filters trigger false alarms** (1, 2). Right, and worth noting the strict comparison was the pattern I copied from the park reads beside them — while `ircut-check.js` has carried an `on()` helper since the day it landed, for exactly this reason. Both go through one `off()` now, and I extended it to the park reads rather than leaving them as the odd pair out: the same failure, one line apart, and fixing only mine would have been a new inconsistency. It is the mirror of `on()` and not its negation, because these keys default **on** and absent has to stay distinct from off.

**A failed refresh shows actuators off** (3). Right, and pre-existing in the mount-time read — `+v > 0` turns an empty body, an error page or a 503 into "off". Putting that call after every toggle would have made it routine rather than rare. It requires an OK response and a number now, and otherwise leaves the control alone.

**Late reads restore stale switch states** (4). Right; two batches exist now where there was one. Generation-stamped.

**Manual time counts toward a fault** (5). The sharpest of the five, because standing the finding down looked like enough and was not: the tracker banked the whole manual interval, so turning following back on would hand the finding an hour it was never entitled to — and the sentence names that duration out loud. The clock is now told whether a disagreement means anything, by both pages, through one exported question. Flips are still counted either way: a camera swinging between day and night is doing that whatever the filter follows.

Re-measured on an hi3516ev300 across six configurations, the two quoted ones included:

```
monitor off                          night live     | ircut live     | lamp live     | (no note)
monitor on, both follow              all hidden                                      | driving night, IR-cut and the lamp
monitor on, filter does not follow   night disabled | ircut LIVE     | lamp disabled | driving night and the lamp
monitor on, not following, QUOTED    night disabled | ircut LIVE     | lamp disabled | driving night and the lamp
monitor on, neither follows          night disabled | ircut LIVE     | lamp LIVE     | driving night
parked filter, QUOTED                night live     | ircut disabled | lamp live     | (no note)
```

PR #508, opened 2026-09-15
telegram, ntfy: send what the camera saw, without a card

openipc-ai:
The camera can now be asked for a short clip over HTTP without a card, a
recorder or an HLS playlist — `GET /video.mp4?duration=10` holds the muxer up
for the life of one request and gives it back afterwards, and `?pre=N` asks for
the seconds before the request as well. These are its first two consumers.

Both senders could already forward a recording the camera handed them, which
needs a card. Without one they could only take a still picture of what the
camera can see *now*, rather than what set the trigger off a minute ago.

## What each sender does

Three questions, in the order the answers are allowed to win:

1. **A path argument** — a file somebody else made, which `sbin/record.sh`
   passes when a motion recording closes. It still wins outright: a fresh
   capture at that point is whatever is happening a minute later, not the
   motion.
2. **`--clip` / `--image`** — a caller that knows what it wants.
3. **The page's own switch**, which governs the schedule, where nobody is
   present to say.

Where the answer is a clip and no file is in hand, the sender fetches
`localhost/video.mp4?pre=N&duration=N` and sends that. Telegram routes it to
`sendVideo`, ntfy pushes it as `video/mp4`; both already dispatched on the
file's extension, so that part needed nothing.

## The webhook verbs name their payload

`?send=image` and `?send=clip`, on both pages. A dashboard pulling a thumbnail
every minute goes on getting one after the schedule has been switched over to
video — the setting answers the schedule, the URL answers itself.
`ntfy.cgi?send=test` stays the page's own button and sends whatever is
configured.

`ntfy.cgi` gains the **Remote send** card `telegram.cgi` already had. It has no
schedule and no webhook of its own, so its clip support would otherwise have
been reachable only from its test button.

## The run-up is usually nothing, and says so

The camera holds a run-up only while another request *that asked for one* is
open. In practice that means two sends overlapping, and nothing else — watching
the Live page does not make one, since that is WebRTC or MSE and neither goes
near this endpoint. The response states what it actually contains and the
sender prints it; where the camera said nothing, no figure is printed at all,
because an absent header is not a run-up of zero.

## A refusal is a refusal

A non-200, or a 200 carrying no bytes, ends the send. Falling back to a picture
would leave an operator believing the clips they configured are arriving.

## Verification

Measured on an **hi3516ev300 + imx335**, 4 Mbit main stream, with no recorder
and no HLS:

| | result |
|---|---|
| Schedule set to video | a ten-second clip of about 540 KB, reported as ten seconds with no run-up |
| Five-second configuration | a five-second clip, again with no run-up |
| `--image` with video configured | still picture, no clip fetched |
| Camera not answering | refused, non-zero exit, no leftover work directory |
| Plain `/video.mp4` viewer open throughout | run-up **0** — a viewer does not make one |
| Second send six seconds into a ten-second capture | run-up **4**, first send 0 |

The pages were rendered on that camera and measured in a real browser engine at
390, 1024 and 2560 px: no page overflow, no clipped `<select>`, no overlapping
boxes, with an untouched sibling page as the control for the two pre-existing
`FORM` measurements the harness reports at every width.

`tests/delivery.test.js` grows the camera's half of the conversation: the fake
server serves `/video.mp4`, records the query it was asked, and can refuse.
That query is the subject of several checks, because a clamp that passed 600
seconds to the camera would look exactly like one that worked — the fake camera
answers whatever it is asked. 45 checks to 74, all passing.

**Requires a majestic that serves `/video.mp4?duration=`**; the two ship
together, so there is no fallback path and none is written.

## Review fixes

`981e729` answers the review: a transfer cut short after a 200 is now a refusal
rather than half a video, and the pages normalise the configured length with
the senders' own bounds so the webhook card cannot promise what the send will
not do. The new test cuts the transfer for real and fails with the fix taken
back out.


openipc-ai:
Two of the four are real and are fixed in `981e729`. One is fair about the
description and is addressed there. One I am declining, with a reason.

**3 — a cut transfer is not a clip.** Right, and it is the same reasoning I
applied to the empty 200 stopping one step short. curl writes the status code
it was given the moment the headers arrive, so a transfer that stalls, resets
or runs into the deadline after a 200 leaves a partial file behind a
successful-looking status. Half a video is worse than none: it plays, up to the
point where it stops, and nothing about it says it was cut. curl's own exit
status is now the first of the three tests, and it says which one refused.

The test cuts the transfer for real — the fake camera announces more than it
writes, then destroys the socket. Taking the new guard back out makes it fail,
which is the only reason to trust it passing.

**4 — the card promised what the send would not do.** Right, and worse than the
finding says: with `600` in the file, one setting read as three numbers. The
list showed `5` (no option matched, so the browser picked the first), the card
promised 600 seconds, and the send asked the camera for 60. The pages now
normalise with the senders' own bounds — deliberately the senders' and not the
offered list's, since what the card promises has to be what the send does, and
a hand-edited length the list does not offer is still honoured. Checked on the
rendered pages with `600` in the file: the list shows 60 and the card says 60.

**1 — lab output.** Fair about the pull request description, which quoted the
senders' stdout verbatim; that is paraphrased now. I have kept the sentence in
`CLAUDE.md`, because it is the form the rule's success criteria asks for rather
than the one it prohibits: a hardware class, a configuration, and the behaviour
observed, with no device identity, address, credential or copied console line
in it. "Measured on an hi3516ev300: a plain viewer open throughout, run-up 0; a
second send six seconds into a ten-second capture, run-up 4" is what makes the
claim checkable by somebody else, and removing the figures would leave the
paragraph asserting behaviour with nothing behind it — which is the failure
mode the checklist's other rule exists to prevent.

**2 — older cameras.** Declining. This WebUI is not backward compatible with
older daemon builds, by policy: the two ship together in one firmware image,
installing this tree onto older firmware is not a supported path, and
capability probes or "older firmware" fallback prose are what that policy
exists to keep out of the tree. `CLAUDE.md` states the requirement plainly in
the paragraph this change adds, which is the honest form of the disclosure —
and the controls are not broken on a camera that cannot serve a clip: the
sender refuses, says so, and the page's test button reports the failure, which
is what finding 3 above is about.

PR #510, opened 2026-09-15
How am I supposed to find a GPIO?

openipc-ai:
## How am I supposed to find a GPIO?

The chip drawing on the Pins page tells you what each pin *is*. It could not
tell you what the thing you soldered to one is *on*, and the hunt that could
have — the one that drives pins and watches for something to change — lived on
the Day / Night page and read its answer off the picture. That finds an IR-cut
filter and can never find anything else.

So the hunt moved to the Pins page, where it belongs: what it does is drive pins
across the whole chip, and everything that makes that safe — the list of pins
that must not be driven, the range, the record that survives the reboot a pin
caused, the chip saying which pins are already carrying something — is about
GPIO in general and has nothing to do with day or night. Day / Night keeps what
is genuinely its own: which pin each coil is on, and testing the filter once
they are set.

And it grew a second way of looking, because the camera can see far more than
its own picture. The section now opens with two doors, each carrying the
sentence that says which one you want:

> **What is my pin wired to?**
> For anything that comes alive when it gets a signal — a wireless card, a card
> slot, a second network port, a relay board. The camera holds each pin in turn
> and tells you the moment something appears, or disappears.
>
> **Which pins move the day/night filter?**
> The filter is the one thing on these pins the camera can see through its own
> lens, so this one works differently: it drives pins two at a time and watches
> the picture. Needs daylight.

A find reads as a sentence — *"Pin 47, held high, switched on 802.11n WLAN
Adapter (1-1) and a wireless adapter (wlan0)."* — and points at the drawing,
where that pin can be set to hold high at every start.

`pin-sweep.js` is the pure half and is tested without a browser: which pins may
be held, in what order, and the stopping rule. It stops at the **first**
difference on purpose — whatever has just come online is still online when the
next pin is held, so every later pin would report the same arrival again or
report it leaving, and the one true answer would be buried.

## Coming back to a scan that did not finish

From a user thread: a scan that keeps taking the camera with it, restarted from
the beginning each time, with a hand-maintained exclusion list in a shell script.

Progress is remembered in the **browser** and the dangerous-pin list on the
**camera**, and the asymmetry is the point: the camera is the thing that reboots.
The pin in flight is written before the request goes out and cleared when it
comes back, which is what lets a later visit tell "the camera went down" from
"something cut the connection" — the camera journals the first for itself, and
the second leaves no trace on it at all.

Three different things can end a run and they get three different answers. The
one that matters most is the middle: a pin can take ethernet down without
troubling the camera, and those are exactly the pins somebody wants out.

## The camera's leave-alone list is visible now

Nothing here called any of the camera's `forget` verbs — zero hits across
`www/`. Two consequences:

- **"Start over" cleared this browser's progress only**, leaving the camera's own
  record of the last pins it drove standing for good. That record is what a later
  visit reads to decide whether anything went wrong. It posts `forget=progress`
  now — never `forget=all`, which would walk the hunt back onto a pin that
  stopped the camera.
- **The list itself was invisible.** It lives on the camera because it must
  outlive the page and the reboot that made it, but the only way to take anything
  off it was one pin at a time in the drawing. The resting card now says how many
  pins are being left alone and why, expands to name them with "try it again" on
  each, and carries one deliberate "Forget the whole list" that asks twice and
  says what is going — a pin the camera went down on is proof that cost a crash
  to earn, and nothing anywhere else remembers it.

## Before you press it

The confirm card says what will be touched, counted from the same facts the
sweep is built from so the number cannot drift: *"37 pins out of 80 will be
tried, at two levels each. 43 are being left alone."* Where the camera cannot
ask its chip what a pin is carrying it says so, rather than implying a clean
bill of health. A range limits which pins are in play, which is what somebody
who already knows which end of the chip to look at actually needs.

## Also

- An excluded pin fell out of the chip drawing entirely: `.mj-pin-avoid` set
  `position: relative` on a `.mj-pin` that is `position: absolute`.
- `tools/check-layout.mjs --open "Label"` presses things before measuring, so a
  card that only exists after a click can be measured at all.
- Removed what nothing called: `mount()`'s returned handle and the variable it
  was assigned to, the `assign`/`setAssign` stubs in the page's lending handle,
  and `pin-sweep`'s `describe`, `LEVELS` and `HOLD_MS` exports — the last of
  which duplicated a default the camera owns and clamps.

## Checked

Full suite green. Layout clean at 390, 1280 and 2560 with the two doors, the
confirm card and the expanded leave-alone list on screen — `--self-test` first,
so a clean result means something. No page errors, against a camera running the
matching build.


openipc-ai:
Ten of the eleven findings were real. Findings 1, 2, 3 and 8 are one defect seen from four sides, so they are answered together.

### The save path — 1, 2, 3, 8

All correct, and `CLAUDE.md` settles it: the WebUI writes through `POST /api/v1/config`, and `GET /api/v1/set` is "the single-key variant the WebUI does not use". I used the one the WebUI does not use, over a method a browser reissues on its own.

The batch write is also the answer to **8**: the server walks every leaf, aborts on the first it rejects and only then reloads and saves — *no persistence partial-credit*. `irCutPin1` and `irCutPin2` must agree, because one pin without the other is a filter that cannot move, and two requests cannot promise that. `raw.js` already says this in as many words.

And **3**: it now re-reads `/api/v1/config.json` and compares both values before saying "Saved." An accepted request is not a stored one — `CLAUDE.md` notes older majestic answers 202 and goes on to ignore what it was sent, and "Saved." over a camera that kept its old wiring is the one outcome worse than an error.

**1** — staging into the settings fields — does not apply here and is the reason the hunt moved: there is no filter form on the Pins page. A button that quietly staged something into a page you would then have to find is worse than one that writes and tells you. The substance of the finding (atomicity, and confirmation rather than assumption) is what the change above delivers.

### 10 — the navigation did nothing

Right, and embarrassing: my own comment claimed it was "the same in-page navigation the tree uses". It is not. This page routes on `?tab=` and never reads the fragment, so `location.hash = 'pins'` changed the URL and left the reader exactly where they were. It now presses the rail's own link, which keeps `wireNav()`'s unsaved-changes prompt instead of duplicating it, and falls back to a plain load when a search is filtering that section out of the rail. The "Saved." message's `#nightMode` link had the same fault and is now a real `?tab=` link.

Checked in a browser against a camera: pressing **Find them on the Pins page** on `?tab=nightMode` now lands on `?tab=pins` with the chip drawn and both hunts offered.

### 4 — a pin that cut the connection was offered straight back

Right, and it is the case the feature exists for. The in-flight pin was persisted before each request and read back by nothing, so reloading after the network dropped rebuilt `todo` with that pin still in it. There is now one card — shown both when the run ends that way and on a later visit — that names the pin and offers *never try it again* or *it was fine, carry on*. The live path and the resume path are the same code, which is how they stay the same answer.

### 6 — a resumed hunt ignored the range

Right. The range lived in page memory and the progress did not carry it, so a reload resumed a fenced-off hunt over the whole chip — driving exactly the pins the owner had ruled out. Both hunts now store the range with their progress and the stored one outranks the page's. The normaliser moved into `pin-sweep.js` where it can be tested: a stored `null`, a string bound, a `NaN` and an empty object are all *no range*, because whatever is in `localStorage` was written by an older version of this page or by hand, and half-believing a bound fences off the wrong pins.

### 9 — a stopped hunt re-drove the pin it had just finished

Right. Progress was recorded one step behind, at the start of the next step, so the last completed one was never written. The pair hunt now records a pair when it is *released*, which is the moment it is ruled out; the pin hunt records from the run's own `tried` list when it resolves. On a pin that cuts the network, re-driving the one just finished is not a wasted second — it is the failure again.

### 5 — the hunt kept driving after its Stop button was gone

Right. A hunt is hundreds of requests that each drive hardware, and its Stop button lives in a box the page throws away on every repaint. `mount()` returns a `stop()` handle again — it had one, I removed it in the dead-code pass because nothing called it, and the answer is that something should. The Pins page stops the previous hunt when it remounts and when it is destroyed.

### 7 — an older camera was offered a hunt it cannot run

Right. The endpoint ignores an unknown parameter and hands back its pad list, which this page would have read as "held it, nothing happened" — two hundred times, cheerfully, finding nothing. The inventory the hunt diffs (`attached`) arrives in the same answer, so its presence is the capability check. Without it only the filter hunt is offered, with one line saying the camera's software is too old for the other and that updating adds it.

### 11 — an error body became pin data

Right. Painting the map from a refusal's JSON would replace a real pad list with an error object — every pad quietly unknown, on the drawing somebody reads to decide what is safe to drive. `r.ok` is checked before parsing.

---

Full suite green, layout clean at 390/1280/2560 with both doors and the expanded leave-alone list on screen, no page errors against a camera running the matching daemon.

openipc-ai:
Correcting my answer on finding 1. I dismissed it and I was wrong.

`CLAUDE.md` on the IR-cut scan: *"Nothing is written to majestic behind anyone's back: the proposal is staged into the hidden fields and the ordinary save bar appears."* I read that as a convention about where a button lives, and argued it did not apply once the hunt moved to a page with no wiring fields on it.

It is not a convention. Two entries above it in the same file is the measured failure it protects: the Day / Night filter test reads the camera's wiring through those fields and refuses to run while they are dirty, because *swap the two coils on the map, do not save, press Test, and the page answered "The IR-cut filter is wired correctly" about an assignment that was backwards and had never been on the camera.* Writing the config from another page steps around the one mechanism that keeps that honest — and "there is no form here" is an argument for carrying the proposal **to** the form, not for writing instead of staging.

Fixed in eaee343: the find crosses the page boundary as a proposal in `sessionStorage` (not `localStorage` — a proposal that outlived the tab would raise a Save bar next week holding pins from a board no longer on the desk), and Day / Night reads it once on mount, removes it whatever happens next, and stages it through `pushAssign()`/`map.set()` — the same path the map's own edits take — with a line saying the pins are filled in and nothing is saved yet.

Checked against a camera: a handed-over find lands in both fields with the Save bar enabled, the note up and the proposal consumed, and `/etc/majestic.yaml` still holds the pins it held before.

That also retires the rest of my answer to 2, 3 and 8: there is no configuration write from this page at all now, so there is no method, no atomicity and no read-back to get right.

PR #512, opened 2026-09-16
ircut: one control per actuator, and the page reads the mode it names

openipc-ai:
The web half of the daemon's `nightMode.irCut` / `nightMode.backlight` — one control per actuator with the three states it really has, in place of two booleans whose four combinations covered three. They ship as one feature; neither is much use without the other.

Reported by the reporter of #492, who counted the combinations and asked why "off" needed a switch at all when clearing the pins says the same thing to the daemon. It does not say the same thing to the owner — off keeps the wiring, which is what the park switch was asked for — but the counting was right.

### The control needed nothing

The form dispatches on schema type, so a string with an `enum` arrives as a `<select>`: two checkboxes become one dropdown with no work here. Verified rather than assumed — the measurement below shows it rendering `[off manual auto]` on a real camera.

### Two reads needed thought

The two booleans were never equally load-bearing, and merging them loses that unless it is put back.

**The test blocker** refuses a verdict about wiring that is only staged, over a set of four that deliberately excluded the follow switch: a verdict is about what happens when the camera pulses the pads, and moving day/night in or out of the job does not change that — a blocker that fires on it is one that gets read past. It now watches whether the mode crosses **off**, rather than whether it changed.

**`wiringKey()`** has the same shape and higher stakes: it stamps a remembered IR-cut verdict with the wiring it was measured against, and #503's guarantee is that the verdict stands until that wiring changes. Stamping the mode itself would throw a good measurement away the moment somebody turned automatic switching on. It stamps driven-or-not.

*Worth knowing on upgrade:* the stamp's spelling changes, so every remembered verdict invalidates once. One re-test, and only for someone who has run the IR-cut test in that browser.

### And one the measurement caught

The Live panel's gate moved from "not auto" to **"manual"**. A filter set to **off** is not the operator's to move either, so keeping the group open for it showed three dead switches — the exact thing the one-sentence chip replaced. The group earns its place only when something in it is live. I only saw that by looking at the rendered panel in the off case.

### Reading rules

An unreadable mode reads as **auto** throughout — the state every camera was in before the key existed, and the one that keeps *asking* rather than standing findings down on a typo. An unquoted `off` in a hand-edited file is YAML's boolean `false` by the time it reaches a browser, so that spelling comes in too.

The **Wiring** heading became **IR-cut filter**, symmetrical with **Camera light** below it: it was named for the pin map it sits above, while the key beneath it was a switch, and it now introduces the filter's own control.

### Measured

On an hi3516ev300. That daemon predates the key, so the schema and the config are patched on the way into the page; everything after that is the real page, the real dispatch and the real layout.

| configuration | live panel | chip |
|---|---|---|
| auto — what every camera was | group hidden | driving night, IR‑cut and the lamp |
| filter manual, monitor on | **ircut live**, lamp greyed | driving night and the lamp |
| filter off, monitor on | group hidden | driving night and the lamp |
| both manual, monitor on | **ircut live, lamp live** | driving night |
| `irCut: false` (unquoted `off`) | group hidden | driving night and the lamp |

Day / Night section: `SELECT "IR-cut filter" = manual [off manual auto]`, `SELECT "Camera light" = auto [off manual auto]`, and the two old checkboxes gone.

`tests/tree.test.js` caught the grouping before I did — an unplaced key and an empty heading — which is what it is for. Local suite otherwise green; `delivery` and `motion-notify` fail identically on master in this container (they shell out), and pass in CI.


openipc-ai:
One taken, one declined.

**Absent modes appear switched off** — real, and the consequence is worse than "a fabricated state": a select with no matching option takes its **first** one, and the first option has no relationship to the key's default. `isp.slowShutter` defaults to medium and would read *disabled*; `nightMode.irCut` defaults to auto and would read **off** — a control claiming an actuator is parked, snapshotted into `state.initial` as though the camera had said so, and one Save away from parking something nobody parked. Fixed in b9094a7, in the shared enum branch, so every enum whose default is not its first option gets it.

Worth recording that it is not reachable today, rather than leaving that to luck: the daemon emits every registered key in `/api/v1/config.json`, defaults included — checked on an hi3516ev300, where `nightMode` carries all nineteen. This is about the page not depending on that. Measured with the key dropped from the config on the way in: the row reads `auto`, and nothing is marked dirty by it.

**Older cameras record false test results** — declining, because the premise is a configuration this project does not ship. The web interface and majestic travel in **one firmware image** and are not compatible across versions by policy: a page carrying `irCut` will never meet a daemon exposing `irCutEnabled`, because they are built and flashed together. That is why the repository's guidance says to state a daemon requirement outright rather than write fallback prose for older builds, and why `TEST_DEPENDS` names the key this schema has rather than carrying both spellings.

The finding is right about what *would* happen if the two were mixed — a parked filter would reach the test and record a stuck verdict. The answer to that is the same as for every other key this page reads from the schema: don't mix them. Carrying a compatibility branch here would be the first one in the file, and it would have to be carried in nine other places that read the same key.

PR #513, opened 2026-09-16
ircut: a mode with nothing wired to it says so

openipc-ai:
Follow-up to #512, from the reporter of #492 arguing that the three modes should be two.

> I am still convinced that the configuration only needs the hardware/wiring settings plus a single choice between MANUAL and AUTOMATIC operation. The absence of a configured control mechanism means OFF.

The state stays — `off` is what makes the daemon *not own the pad* while keeping the pin numbers, and "leave it in manual and don't touch it" is not the same thing, because manual exports the pad and holds a level for the life of the process. But he is right about what he was looking at: **three modes on an actuator with no pad are three ways of saying the same nothing**, and the row offered the choice anyway. He was reading a control that could not do anything and reasoning from it.

So the row says it now:

> Nothing is connected to the IR-cut filter yet — assign its coils on the map above, and this will start to mean something.

The lamp gets its own sentence and its own test, because a dimmable lamp lives on a PWM channel and may have no pad at all — "no pin" is not what unwired means there.

**Said rather than disabled**, which is the call `x-requires` already makes on this page: a mode is a legitimate thing to set before the wire is on, it is remembered, and it starts meaning something the moment a pad is assigned. Disabling would only move the surprise to whoever sets it first and wires afterwards.

It repaints from `updateDirty()` rather than its own row's events, for the reason `syncTestBtn()` hangs there: the pads are assigned by the **pin map**, and the map funnels through that and nothing else. `state.repaint` is where a row whose text depends on *other* fields registers, cleared with `state.fields` since the closures hold nodes from the section being replaced.

### Measured

On an hi3516ev300. The daemon there predates the key, so the schema and config are patched on the way into the page; everything after that is the real page.

| wiring | IR-cut filter | Camera light |
|---|---|---|
| nothing wired | says it | says it |
| filter wired, lamp not | silent | says it |
| lamp on a PWM channel, no pad | says it | silent |
| both wired | silent | silent |

Assigning a coil on the map clears the sentence **immediately, with no save**. No page errors in any of the four.


openipc-ai:
Both taken, and the first was the half of my own claim I had not tested. Fixed in 8ba7106.

**Cleared wiring still looks connected.** Right. I checked that assigning a coil takes the sentence away and never checked that clearing the last one brings it back — and it did not, because the saved value underneath went on satisfying the test. Each key resolves separately now, with a control on the page winning over the saved value: the precedence `mj-requires` uses, for the same reason — an edit that has not been saved yet is what the operator can see.

**Boolean pins look like real wiring.** Also right, and it is a trap this file's neighbour carries a paragraph about: `Number(false)` is 0 and GPIO 0 is a real pad, so `irCutPin1: false` in a hand-edited file counted as wiring. It asks `isNumish()` now — the spelling `testBlocker()` already uses. A third predicate for "is this a pad" is how the second one drifts, which is the same mistake I made with the boolean-off spellings two branches ago.

Re-measured on an hi3516ev300, with the two cases the first round lacked:

```
both wired                         filter silent | lamp silent
  -- cleared 3 pin fields by hand: filter SAYS IT | lamp SAYS IT   (no save)
pin hand-edited to false           filter SAYS IT | lamp SAYS IT
```

and the original four unchanged, no page errors.

PR #517, opened 2026-09-16
live: zooming in tells the camera to spend its bits there

widgetii:
Needs a majestic that declares `videoN.roiRect` and `videoN.roiQp` as
live-class keys. Until a camera has one, the configuration carries neither —
and that absence IS the capability test, so this file finds nothing and does
nothing rather than offering something the encoder cannot do.

live: zooming in tells the camera to spend its bits there

## What is wrong

Zooming the Live page is pure browser magnification. `preview-zoom.js`
scales the decoded `<video>`, so at 3x one stream pixel becomes nine screen
pixels, and the camera is never told that anybody wanted a closer look. For
the thing people actually zoom in for — read a face, read a number plate —
the dominant loss is not resolution but BITS: an encoder spreads its budget
over the whole frame, so a plate filling two per cent of the picture gets
about two per cent of them, and rate control smooths away exactly the
high-frequency detail the glyphs are made of. No amount of magnification
recovers it.

## The change

Nothing to press. The part of the frame on screen becomes an encoder region
at a lower quantiser than the rest; open the view back out and it goes away.
The rectangle is simply what is visible, which is the one rectangle the
viewer has already said they care about.

It needs the majestic change that adds `videoN.roiRect`/`roiQp` as live-class
keys. Without it the keys are absent from the configuration, and their
absence IS the capability test — majestic declares them only where the
backend can program a region — so on every other camera this file reads the
config, finds nothing, and does nothing for the rest of the visit.

`preview-zoom.js` grows two things and no behaviour: `view()`, what of the
frame is on screen in the stream's own pixels, and `onView()`, fired whenever
that moves. A LIST, unlike `onScale`, which is a single slot this page has
already taken — and fired on pans too, which `announce()` deliberately is
not, because the chip it repaints prints only a percentage while a pan
changes the rectangle without changing the scale at all.

## The parts that are judgement rather than plumbing

TWO thresholds, not one: it engages below 0.55 of the frame on screen and
holds until 0.72. A single threshold flaps — a zoom that settles near it
would engage, release and engage again, re-programming the encoder each time
for a hand that never stopped moving.

PUSHES ON SETTLE, 350 ms, plus a "did it actually move" test at a twentieth
of the frame. `onView` fires from inside a gesture; without this a drag is
one request per pointermove, against a camera whose whole HTTP server is one
thread.

STRENGTH FOLLOWS THE SPARE BITRATE, and the direction is the opposite of the
intuitive one. Measured on a hi3516ev300: the GAIN is the same at 4 Mbps and
at 1 Mbps (+7.2% of high-frequency detail inside the region), so what the
headroom decides is not what the region gets but what the REST of the picture
pays — the background moved -0.6% with room to spare and -5.9% at the
ceiling. So it pushes hardest where it is nearly free: -16 below 55% of the
configured bitrate, -12 to 85%, -8 above. -16 is also where the curve
flattens; -51 was worth another 0.8 of a per cent, so there is nothing past
it worth offering.

IT STOPS WHEN NOBODY IS LOOKING. A tab left zoomed in would go on re-tuning
the stream for a picture nobody is watching, so the region is given back on
`visibilitychange` and taken up again on return, and on `pagehide`.

## Honesty, with no control to hang it on

This changes the encoded stream, so it reaches every viewer, every NVR and
the recording. That is the reason to prefer it to a private view — the plate
is in the clip you download tomorrow — and it is why the status chip gains
` · detail here` while a region is in force, with the full sentence as the
chip's tooltip. A page that quietly re-tuned the recording would be the worse
trade; a banner for something nobody pressed would be the page talking about
itself.

Regions are pushed to `/api/v1/live`, which applies without writing
majestic.yaml. Nothing here is saved, and an empty value is the undo that
means "put back what is configured" rather than a list of no rectangles,
which would delete regions somebody had saved.

Coordinates go through `mj-region.js`, the map the settings page already uses
to land a rectangle from one stream on another — sensor frame in between, so
a region measured on the sub stream reaches the main encoder in the right
place. Where the camera does not describe its crops it falls back to the
plain ratio of the two frames, which is exactly right while neither is
cropped and is what the settings page falls back to.

## Test

`tests/preview-roi.test.js` drives the module in a `vm` over stubs. Every
branch of this is silent: engage too eagerly and a viewer who barely zoomed
re-tunes the recording for everybody, engage on every pointer move and a drag
is thirty encoder re-programs, get the rectangle wrong and the bits go
somewhere real but not where anybody is looking. None of it reports anything,
and reaching any of it on a camera needs an encoder with ROI, a picture
detailed enough for a few per cent to be visible, and a zoom gesture.

The file is deliberately NOT in the `SRCS` of `auto-source.test.js` or
`staging.test.js`: it is DOM-free but it is the page that must keep running
in their bare vm, and both suites passing untouched is what says the split
was made in the right place.

`npm test` green, `tools/lint-templates.sh` green. The new case was shown
able to fail before its pass was believed.

## On hardware

Lab hi3516ev300 + imx335, with the matching majestic. At rest, 89.6% of the
frame on screen and no region. Wheel-zoomed to 300% (3.8% on screen), the
chip read `H264 2592×1520 · 19 fps · 300% · detail here` and the camera's own
`/proc/umap/venc` table carried `1632x560x544x288 qp -16`. Back to Fit and
the table was empty again.


widgetii:
Addresses the seven code findings from the review on this PR, plus the
eighth: the description named a sibling repository and its tracker item,
which has been rewritten to describe the dependency by capability instead.

live: the review findings on the encoder region

Seven, from the review posted on the pull request. The first is the one that
would have been noticed last and mattered most.

## Every MJPEG frame reset the settle timer

`setFrame()` ran a layout on every call, including the ones that hand it the
size it already had — and the multipart rung fires `load` on EVERY JPEG. So a
view was published twenty times a second, the region's 350 ms debounce was
reset by each one, and on that rung the timer would have fired NEVER. A
continuous preview, silently reaching the camera not at all.

A frame that has not changed asks for no layout now.

## An unparseable 200 read as "applied"

`push()` turned a failed `json()` into `{}`, whose missing `refused` then read
as confirmation. The page would light the chip and state that the recording was
being re-tuned on the strength of a body the camera never sent. The reply has a
shape; anything else is not a yes.

## A failed clear said nothing was in force

`clear()` dropped `engaged` before the round trip, and `push()` turned every
failure into `false` without putting it back. A clear that failed while zooming
out left the camera holding the region — affecting every viewer and the
recording — with the page claiming nothing was active. That is the one thing
this feature must never get wrong, because there is no control on screen to
check it against. The claim now stands until the camera confirms it let go.

## A push landing after a clear put the region back

The settle callback marked its rectangle engaged on success with nothing to say
the world had moved on. Hide the tab or change channel after the timer fired
and the stale completion re-applied the region against a picture nobody was
watching, chip and all. Every push carries the generation it belongs to now,
and a clear moves it on.

## The first zoom of a visit could be lost

`supported` is set from a promise, and `consider()` drops everything that
arrives before it. A viewer who zoomed and then held the picture got no further
view, so nothing reached the encoder until they happened to move it again. The
capability answer replays the view as it stands.

## A channel change picked the strength from the wrong channel

`mj-stream-changed` was dispatched BEFORE `stream` was updated, so a listener
asking which channel is shown was answered with the one being left — and the
strength for the new channel was chosen from the old channel's bitrate. Fired
after the assignment.

## The crop-map fallback needed the map

`forChannel()` falls back to the ratio of two frames when the camera does not
describe its crops, but took both sizes from that same absent report. On a
camera with regions and no `/api/v1/osd`, every channel but the one on screen
was skipped — so the recording got none of the promised detail. The
configuration's own `videoN.size` is the fallback it should always have been.

## Test

`tests/preview-roi.test.js` gains a case for each, driven through the same vm
harness: a malformed 200 claims nothing, a failed clear keeps the claim
standing, a push released after a hide is dropped, a view delivered before the
capability answer is replayed, and a camera with no crop report still marks
both channels.

`npm test` green, `tools/lint-templates.sh` green.

PR #519, opened 2026-09-16
updatewebui: download the installer instead of carrying it

openipc-ai:
## Why

`gk7202v300_lite` stopped building. CI: `rootfs.squashfs: [5132KB/5120KB]`, over
by 12 KB. Last night's image was 5108 KB — it had **12.0 KB** of headroom and
this week's WebUI work spent it.

`sbin/updatewebui` is 48,779 bytes of that partition, and its only job is to
download a tree from GitHub. A camera that cannot reach GitHub could never use
it, so carrying it buys nothing.

**The download path already ships.** Every image has this in `/etc/profile`:

```sh
majestic_webui() {
	curl -s -L https://raw.githubusercontent.com/openipc/majestic-webui/master/sbin/updatewebui -o /tmp/webui.sh && sh /tmp/webui.sh
}
```

What this adds is the command still being called `updatewebui`, arguments
reaching it, and the copy being kept.

## What changed

`/usr/sbin/updatewebui` becomes `sbin/updatewebui-fetch`: it works out which
branch is being installed, downloads that branch's `sbin/updatewebui`, checks
what came back, caches it at `/etc/webui/updatewebui.sh` and `exec`s it with
the arguments it was given.

**It is also a fix, and that is worth more than the bytes.** A script that
installs itself is always one release behind: the copy running an install was
the one the *previous* install left, so a fix to the installer took effect on
the run **after** the one that fetched it — exactly backwards for the run
bringing the fix down. The answer to that was `hand_over`, comparing
`scr_version` against the downloaded tree's and `exec`ing it mid-flight behind
a one-hop `UPDATEWEBUI_HANDOVER` guard. Fetching first makes the running copy
the newest by construction, so `hand_over`, `--no-self-update` and the whole
guard go — and with them the coupling that made a change to the variant skip
list oblige a version bump, which is how a stale installer once put `wfb.cgi`
on a lite camera. Measured on the lab board at the time, not reasoned.

`--restore` is what it costs. That mode reaches no network and says so in its
own help, so a cached installer is used there **as-is** rather than refreshed;
any cache is by definition a copy that ran an install on this camera, so it
knows how to undo one. With no cache and no network it prints the `curl` line
to fetch one by hand.

The two refusals are the ones `curl` makes easy to get wrong: `-f`, without
which a 404 is written to the file and exits 0, and a content check, because a
200 carrying the wrong thing is the failure `-f` cannot see. `/etc/profile`'s
version has had neither for years — fixed in the firmware half.

## What it is worth

Measured by rebuilding the whole rootfs with the partition's own compressor
(xz, 128K blocks). The method is validated first: re-squashing the shipped
image untouched reproduces it **byte-for-byte**, so these are image sizes, not
estimates.

```
last night's image, re-squashed unchanged   5230592 B   (+0 vs the real one)

today's payload, installer carried          5296128 B
today's payload, installer fetched          5283840 B   -12288 B
```

12 KB, which is the whole of the 12 KB overflow — and none of the margin.
`CHECK_SIZE` warns below 32 KB left, so that board would go from error to
warning, not to comfortable. Whether more is wanted is a separate question;
the next lever is an excludes list or the 191 KB sitting unused in that
board's kernel partition.

The stub is 4,505 bytes with its comments and a 739-byte version saves exactly
the same 12 KB — at 128K block granularity it disappears into padding. So
keeping the command name, the arguments and the explanation costs nothing.

## Verified on hardware

An hi3516av300 + imx415, lite build. Described rather than pasted: this
repository's own rule forbids console output captured from a lab camera, and
the point of describing it by class and configuration is that you can redo it.

**The ordinary path.** With the stub put where the image puts it, asking it to
install this branch made it fetch the branch's installer and run it — the run
announced installer **v10**, resolved this branch's head commit, and planned
the install. After a real install `/usr/sbin/updatewebui` was the stub (8,186
bytes), `updatewebui-fetch` was absent, and the kept copy was the 45,691-byte
installer at `scr_version=10`. So the walk maps `sbin/updatewebui-fetch` onto
the canonical name and skips the 48 KB script, and the camera ends up with the
stub under the name people type.

**A dry run keeps nothing.** With the kept copy deleted first, a `--dry-run`
fetched the installer, ran it, printed its plan — and left no copy behind and
no state directory created.

**The archive path, with GitHub unreachable.** `raw.githubusercontent.com` and
`github.com` pointed at 127.0.0.1 in `/etc/hosts`, and no kept copy. The
control first: a plain `--branch` run **failed**, with the message naming both
recoveries — so the blackhole was real. Then the same camera, same conditions,
`--url=file:///tmp/tree.zip` against a branch archive already on it:
**succeeded**, announced installer v10, and planned the install. That is the
case finding 4 was about, and it is how a camera that must not be given a
route out is updated.

**`--restore` with GitHub unreachable.** After a real install, blackholed
again: restore ran from the kept copy with no network at all, removed what the
install had put there, emptied the manifest, and left the firmware's own
48,768-byte installer showing through from `/rom` again. The offer to remove
the emptied overlay directories correctly declined to act on a non-terminal
stdin, so nothing rebooted.

The camera was left on its firmware WebUI, with the manifest emptied and the
scratch files removed.

## Tests

`tests/updatewebui-fetch.test.js` — a fake `curl` and a fake installer, the
real stub run against them. Branch parsing in all three spellings the installer
accepts plus the four things that are *not* a branch; both refusals (an error
page arriving with a 200, and a shell script that is not this one); the offline
fallback in both directions; `--restore` not fetching when a cache exists and
falling back to fetching when none does; and the arguments arriving in order.

Shown able to fail, each in the right place:

| break | goes red |
|---|---|
| drop `-f` from the fetch | `curl is asked to fail on an HTTP error (-f)` |
| always use `master` | all four branch cases |
| remove the `--restore` shortcut | `a cached installer is used without fetching` |

`npm test` exit 0, 0 FAILs. `tools/lint-templates.sh` clean — `ok: 60 haserl
templates, 27 shell scripts`, the stub among them.

## Ordering

The firmware half is OpenIPC/firmware#2431: the package installs the stub under
the canonical name, and `/etc/profile`'s helper is fixed. It is guarded on the
stub actually being in the tarball rather than on `-f`, so **either repo can
land first** — a firmware built against a release from before this change keeps
its own installer and behaves exactly as today.


openipc-ai:
Seven of the eight are taken; one I think is wrong. Fixed in 5e17462, and the
PR description has been rewritten.

**1. Lab console output — accepted, and it was mine to know better.** The
description pasted the hardware run's transcript. This repository's own rule 6
prohibits exactly that, and I wrote that rule after doing a version of this
before. The verification is now described by hardware class and configuration,
which is also what makes it reproducible by someone who is not me. The same
edit has been made to the companion firmware PR.

**2. `majestic_webui()` as a private symbol — rejected, and the reason is
checkable.** That function is in `general/overlay/etc/profile` in
**OpenIPC/firmware**, which is public, and the companion PR edits it in the
open. The rule is titled "No references into a sibling repository's *private*
source" and its failure criteria are about the private daemon repository: a
path inside its tree, its name or URL, a reference into its tracker, or a
C-style identifier attributed to the daemon. A shell function in a public
overlay is none of those, and a reader can follow the pointer — which is the
objective the rule states. I have left the one remaining mention, in the test
that explains why the `-f` check exists, because it is true and useful.

**3. First update fails with no state directory — accepted, real, and my test
was hiding it.** The download was staged at `$CACHE.new`, inside `/etc/webui`,
which a camera that has never run an install does not have; curl could not open
its output file, so the very first update failed before it started. Everything
is staged under `/tmp` now and `$STATE_DIR` is created only by a run that is
going to keep a copy. The fixture no longer creates its stand-in state
directory up front, which is what makes the case visible — with the old staging
restored, five assertions go red.

**4. `--url` still required GitHub — accepted, and this was the worst of them.**
A zip already on the camera, named as `file://`, is how a camera that must not
be given a route to the internet is updated, and it worked before this split
because the installer was on the image. `--url` now takes the installer out of
the named archive, so the archive installs itself and a tree carrying an
installer fix applies it to its own install rather than to the next one.
`file://` is copied rather than handed to curl: whether a given build has the
file protocol compiled in is not something to discover on a camera with no
route out. An archive carrying no installer falls back to the kept copy, and
with neither the run stops and says which.

Proven on hardware with a control, because this one deserved it. With
`raw.githubusercontent.com` and `github.com` blackholed in `/etc/hosts` and no
kept copy: a plain `--branch` run fails, naming both recoveries — so the
blackhole is real — and the same run with `--url=file://` against an archive on
the camera succeeds and plans the install.

**5. A dry run kept the installer — accepted.** That is a write to the camera
by the one mode that promises to touch nothing, and it also made the sentence
`--restore` leans on ("any copy here ran an install on this camera") untrue. A
copy is now kept only for a real run from GitHub: never for a dry run, never
for an archive, which is not ours to keep. Verified on the camera: with the
copy deleted first, a `--dry-run` fetched, ran, printed its plan, and left
neither a copy nor a created state directory.

I did not take the rest of that finding — promoting only after a *successful*
install. `exec` is what hands the run over, so there is no exit status to wait
for without keeping a shell alive across the whole install purely to observe
it, and an installer that passed both content checks is a reasonable thing to
keep. The dry-run half was the sharp edge and it is closed.

**6. `--restore` contacts GitHub with no copy — half accepted.** I am keeping
the fetch: refusing to undo an install that went wrong, on a camera that can
reach the network, is worse than the documented promise being approximate. What
was genuinely wrong is that the promise then read as absolute, so the
installer's own `--restore` text now says it downloads nothing *itself* and
that the stub prefers the kept copy, so a camera that has installed before
needs no network there. Verified: after a real install, blackholed again,
`--restore` ran from the kept copy with no network, emptied the manifest and
left the firmware's own installer showing through from `/rom`.

**7. Offline `--help` — accepted.** "Cannot reach GitHub" is no answer to "what
does this command do", and a camera with no route is exactly where somebody
needs the options. It falls back to a built-in summary naming them and the
`curl` line to fetch the installer by hand.

Each fix has a control showing the test can see it: stage inside the state
directory again, stop opening the archive, let a dry run keep a copy, or remove
the built-in help, and the matching assertions go red. `npm test` exit 0, 0
FAILs.

PR #522, opened 2026-09-17
live: the best picture the camera has of the part you are looking at

widgetii:
Zooming in on the live page magnifies the stream, and the stream is the camera's compromise: a bitrate spread over the whole frame, a quantiser deciding what to throw away, and at 250% one stream pixel stretched over six screen ones. None of that is recoverable from the video, however hard the encoder is pushed.

What *is* available is the frame the encoder was handed before any of it happened. `/image.yuv420` serves it uncompressed, and it now takes `?crop=`, so it will serve just the rectangle on screen — a 982×518 region comes back in about **0.2 s and 760 kB**, against twelve megabytes for the whole frame.

**Detail** freezes the picture and puts that in its place. Esc, a click, or any pan or zoom returns to live.

It is a deliberate act and it costs motion, so it is a control rather than something that happens to you — and the note says what it is and keeps counting its age for as long as it is up. A frozen picture that reads as a live one is the one genuinely dangerous thing this can do.

## Where it pays, measured

Against the uncompressed frame itself, using that camera's own encoder settings:

| | luma PSNR |
|---|---|
| its QP floor — a static, well-lit main-stream view | 42.8 dB |
| the coarse end of its range | 35.3 dB |
| a viewer on a 640×360 sub stream | **28.5 dB** |

A resolution loss costs far more than a quantiser ever does, and the crop always comes from the main channel. So this is worth most exactly where the live picture is worst, which is the right way round — and on a well-lit main-stream view at the QP floor it wins almost nothing, which is worth knowing before anyone expects magic.

## Capability probe

The control is offered only after the camera answers a **sixteen pixel** probe. Asking for a whole frame to find out whether the endpoint exists would move twelve megabytes on every page load and take a capture slot while it did. A 16×16 crop is 384 bytes and answers the better question — not "is there an `/image.yuv420`" but "does this build understand `?crop=` and say what it sent". An older camera serves the endpoint and ignores the parameter, so it comes back with a whole frame and no geometry headers; that is a no, and the control stays hidden.

## Three things this gets right, each of which was wrong first

Every one was found on a camera rather than reasoned about.

**The chroma pair order is read, never assumed.** The reply states it and it is not the same on every part. A decoder that guesses swaps red and blue on half the cameras in the field, which looks like a broken camera rather than a broken client.

**The crop is snapped to the chroma grid before it is asked for.** The camera snaps it anyway — a 4:2:0 frame carries one chroma sample per 2×2 block of luma, and an odd edge has no chroma of its own — so a request of `677x367x643x339` comes back as `676x366x644x340`: up to a pixel wider on each side, and shifted. Drawing that as though it were the rectangle requested put the still one to two pixels off the video it replaced, plainly visible at 250%. Asking for a rectangle already on the grid makes the answer predictable, and the same rectangle then decides where the picture is drawn.

**Chroma is interpolated, and sited the way 4:2:0 actually sites it** — horizontally on the even luma column, vertically halfway between the two rows it covers. Repeating the nearest sample instead drags every colour edge half a sample sideways, which measured as a whole luma pixel horizontally, and gives colour edges a staircase besides.

## Two small changes outside the new file

`preview-zoom.js` grows `pic` and `scale` on `view()`. `visible` says *what* is on screen but not *where*, and assuming it fills the stage is wrong under Fit and wrong again the moment a letterbox appears. The still is now placed with the same arithmetic the video is placed with, so it lands on the same pixels rather than near them. What is left is under one stream pixel, and is the browser scaling a `<video>` and an `` through different resampling paths.

The module also **owns its own geometry rather than borrowing it from the stylesheet**. A camera running an older `bootstrap.override.css` drops the rule, the image falls back to static flow behind the absolutely-positioned video, and what the operator sees is the note counting up beside a picture that never stopped moving — reported from a real camera during testing. Inline styles cannot go stale; the CSS rule stays as the same thing said twice. And a picture that does not decode now says so: an `` whose source fails reports `complete` with a natural size of zero and paints nothing, so without waiting for the load event the page would announce a still and count its age over live video.

## Testing

`npm test` passes untouched — which is the point of the file split: `auto-source` and `staging` run `preview-page.js` in a bare `vm` whose stub elements have no canvas, no fetch and no layout, and this module is in neither `SRCS` list. `tools/lint-templates.sh` clean. No new Bootstrap class needs a CSS regeneration (`small` is already in the purged subset).

Driven on a real camera in a browser: control appears only when supported, crop lands on exactly the region on screen (verified independently by pulling the same rectangle and drawing it on the full frame), 503 while another capture is running is reported as busy rather than as an error, and the age keeps counting. Alignment was measured by cross-correlating the live and still captures — vertical exact, horizontal residual under one stream pixel.


widgetii:
All eight addressed in `6d1f6c7`. Findings 1, 3 and 5 were one fault and the most serious of them:

**Unknown/older/fallback streams grabbing the wrong scene.** Correct, and worse than the report suggests — when the module could not work out how the shown stream maps onto the main channel it carried on with a 1:1 guess *and* still offered the control. A crop of the wrong place looks exactly like a crop of the right one, so nothing on screen would have said so. Geometry is now either known or `null`, and a null one refuses with "the camera did not say how its streams line up". No coercion of an absent stream reading to channel zero, no reading a failed `/api/v1/osd` as "the two channels frame the same scene" (rule 3), and watching the main stream is recognised as the one case with nothing to map.

**Channel switches and silent fallbacks (4, 5).** The map is now tagged with the stream it was learnt for and rechecked at the moment of the grab, which covers the served channel changing without an event. The `mj-stream-changed` handler drops the map *before* starting the refresh, so a grab in that window finds nothing rather than the old transform.

**apiFetch (2).** Right, and it matters more than a convention: with the native fetch an expired session made the probe 401, the control quietly never appeared, and the redirect that exists for exactly this never fired.

**Stuck control (6).** The grab now carries an AbortController and a 20 s deadline, and cancelling aborts it rather than only marking it stale.

**Unknown pixel formats (7).** Only the three layouts this page can decode are accepted, at the probe and again at the grab. Anything else was going through the NV12 path and would have shown confidently wrong colours.

**Blob leak (8).** One `dropSrc()` that revokes and removes; both decode-failure paths were only removing the attribute.

Re-verified on a camera on both channels — main, and a 640×360 sub stream where a 2160×1184 rectangle of the main channel comes back, which is the case this feature is worth most in.

widgetii:
Finding 5 addressed in `98bb012` — and checking it turned up a worse one I had just introduced in the previous commit.

The map from the main channel to the shown stream is built against the size the camera **declares** that stream to be, while the zoom module measures what the decoder actually **produced**. WebRTC sends a smaller picture than the channel is configured for whenever the link asks it to, so the two are not the same thing and every crop is out by that ratio. For a sub stream the error hid inside the map; for the main stream `6d1f6c7` had just replaced it with a bare 1:1 — correct only if nothing had been scaled underneath us.

The transform is now built in two steps — main to the stream as declared, then on to the stream as it is arriving — with the second measured every time a crop is planned rather than learnt once. That also covers the original report: a served channel that changes without an event is caught by the grab-time recheck added in `6d1f6c7`, and a frame that changes size without the channel changing is caught by this one.

Worth noting the finding text still quotes the `now | 0` line from the first commit, which no longer exists — but the concern behind it was real and is now covered from both directions.

Re-verified on a camera on both channels. The main-stream seam is inside the measurement noise now: cross-correlating live against still, the best offset is worth 2.5% over no offset at all, where it was 13% before.

PR #523, opened 2026-09-17
night: read the camera's own count of its day/night switches

widgetii:
The Dashboard already warned about a camera hunting between day and night. On a camera anyone actually needs it for, it could not fire.

`tracker()` counts flips **in this browser**, from the moment the page loaded. At a 45 s cycle that is two minutes of sitting on the page before three flips accumulate; at one flip every ten minutes — still hundreds of coil pulses a week, still fatal to the filter — it is half an hour. Nobody watches a camera on a pole in another town for half an hour, which is how a board flaps for hours without anyone noticing.

## What changes

majestic now publishes `night_flips_1h`: switches in the last hour, counted by the daemon, which is the only observer that is always there. **One reading answers the question**, so it is already true when the page opens. `diagnose()` prefers it and keeps the tracker as the fallback for a camera whose daemon does not publish it — which is every camera until that ships. It rises to `danger` on the camera's own count, because a latching solenoid driven through its whole travel several times an hour is wear, not a nuisance.

Two config findings go with it, and they are the cheapest thing here: **they need no sample, no clock and no history at all.**

- **A night gain threshold of 1x cannot ever be false** — a sensor never reports less than 1x gain — so the camera returns to night seconds after every day, in any light, for ever. That is not a scene problem, and the existing advice (a security lamp, headlights, raise the delays) is exactly wrong for it: a longer delay only makes the cycle slower. So where this finding has named the cause, the hunting banner drops the environmental advice and says what the flapping is costing instead — the same rule the Dashboard already applies to its no-video banner.
- **A value the setting would not accept** — outside its range, or not a whole number. The form refuses those and the API rejects them, so a camera carrying one was edited by hand. Reported, not corrected: the value is the owner's.

## Which mechanism is live is asked of the camera

`night_mode_source` decides whether these two findings apply, not a reconstruction from config keys. That reconstruction cannot be complete — the ADC source has no control on this page at all — so a camera running on its light sensor pad would be read as running on gain and accused of a fault that is deciding nothing. The config test is kept only for a camera that has not answered yet: a daemon too old to publish the gauge, or the couple of seconds before the first heartbeat. It is the weaker test and errs silent.

Where another mechanism shadows the gains, the range finding still reports the bad value but says it decides nothing now and would be used unchanged if automatic day/night took over — rather than claiming the camera is running on it, which would contradict the note the settings page puts under the same field.

## Reproducing it

Set automatic day/night with a five-second delay each way and `autoNightGain: 1` on any camera in any light. The night condition is then true at every gain the sensor can report, so the camera re-enters night one delay after each day, indefinitely. The cycle settles at `autoDayDelay` x8 — the anti-flap penalty at its ceiling — plus `autoNightDelay`.

Illustrative of what the page then shows (synthetic, not a device capture):

```
>> BANNER [DANGER] Night is set to trigger at any brightness
   "Gain multiple that means night" is 1x, and 2x is the lowest setting that
   can ever be false... This is not a tuning problem: raising the delays only
   makes the cycle slower.
   also [DANGER] The camera keeps switching between day and night
   N switches in the last hour, counted by the camera itself. The IR-cut
   filter is a latching solenoid and this is wearing it out.
```

Before this change the same camera produced no banner at all until somebody had watched the page for minutes.

## Tests

38 cases in `tests/ircut-check.test.js`: the config findings with no sample at all, the source guards for all four non-automatic sources, the fallback when the camera has not answered, 2x as the lowest workable threshold, a fractional value inside the bounds, the shadowed-value wording, the camera's count beating a browser tracker at zero, the fallback for a daemon that publishes nothing, the advice fork, and the ordering. Full suite green.


widgetii:
All four findings were real and are fixed in 02a4e1e.

**1 — Lab camera output remains public (rule violation).** Correct on the rule, and the description was the substantive part: rule 6's failure criteria names "pasted console/log output … captured from a lab device" in the diff, title **or description**, and the before/after blocks were exactly that. The description is rewritten around a reproduction recipe — set `autoNightGain: 1` with automatic day/night on any camera — with an explicitly synthetic illustration of the output. The code comment is rewritten the same way: it now states the *configuration* the fault happens under, which is what makes it checkable anywhere, rather than what one device read on one day.

**2 — Sensor-pad cameras get false alerts.** Right, and right on the principle too — `mj-settings.js` already states it for the notes under these same fields: *"Which one is live is not inferred from the config here: the camera publishes the answer and its answer is the one that matters."* `night_mode_source` now decides. The config reconstruction is kept only for a camera that has not answered (a daemon too old to publish the gauge, or the seconds before the first heartbeat) — that fallback is the whole value of a finding that needs nothing observed, so it stays as the weaker test, erring silent.

**3 — Dormant gain settings are called active.** Correct, and it contradicted this page's own note under the same field. The range finding now says the camera is running on the value only when the automatic monitor is what decides, and otherwise that it decides nothing now but would be used unchanged if automatic day/night took over. Still reported either way — the value is wrong either way and bites the moment the other mechanism is cleared.

**4 — Decimal gain values escape validation.** Correct. Both keys are declared as whole numbers, `2.5` sits inside the bounds, and that is precisely the shape a hand edit takes. The check now tests `Number.isInteger` as well and names the fraction rather than claiming it is out of range.

Tests went 21 → 38 in this group, covering each: all four non-automatic sources leaving the tautology alone, the no-sample fallback still answering, the fractional value, and the shadowed wording asserted on the text rather than just the finding's presence. Full suite green.

PR #525, opened 2026-09-17
live: the other camera's view of the part you drew

widgetii:
A camera calibrated against another one that sees the same scene can say where a rectangle of its own picture lands in the other's: `GET /api/v1/calibration/map?peer=<name>&rect=XxYxWxH` answers with the rectangle in the peer's pixels and the address of the peer's snapshot cropped to it. This is the control that asks.

### What the operator gets

- A **Peer** toggle on the live bar, beside Area. It arms the same rubber band; one drag on the picture becomes a question, and the answer is a note: where the rectangle lands in the other camera's picture, with a link that opens that camera's crop of it in a new tab.
- The link opens under the other camera's own login. A cookie for this camera never travels to another one (the reason `cameras-switch.js` gives), so the peer's picture is fetched by nobody but the viewer, on the peer's own origin. Nothing is embedded and nothing is proxied.
- Shown only on a camera that has both the endpoint and a calibrated peer, learnt from `calibration.peers` in its configuration; a picker appears only with two or more cameras to choose between. A refusal is a sentence, never a link, and only an `http(s)` address the camera handed back is ever one.

### How the rectangle is converted

The camera calibrates in the main channel's pixels. A drag is stage pixels; stage to shown-stream pixels is preview-zoom's placement run backwards; shown to main is the camera's own per-stream map from `/api/v1/osd` (the one `mj-region.js` builds), scaled by what the decoder actually delivered against what the channel declares. The same two things `preview-still.js` learns, for the same reason: a rectangle drawn on the sub stream converted as though it were the main one lands somewhere else in the scene while still looking like an answer.

### Playing nicely with the other rubber band

While this control is armed its listeners run first, in the capture phase, and stop the event there, so the same drag does not also pan or zoom. Arming it puts zoom-to-area away. Same floor as Area (a slip is not a question); disarms after one drag; Esc cancels, and clears the note.

### Size

`www/a/preview-peer.js` is 5.8 KB minified; the markup adds one toggle, one `<select>` and one note element to the player, and one CSS rule.

### Checked

- `npm test` (the new `tests/preview-peer.test.js` pins the two conversions with numbers, the appearance rules, the refusals, and the disarming of Area), `tools/lint-templates.sh`, `npm run build:dist`.
- On a camera with a calibrated peer, in headless Chromium through the sign-in the layout tool uses: the toggle appeared, a drag on the sub stream at 206% became a question in main-channel pixels, the note showed where it lands, and the link fetched the peer's crop after signing in there.


widgetii:
Review findings, one commit (`live: four review findings, all taken`):

1. **Stale map after a silent channel switch** — taken. `toMain()` now checks the learnt stream against the one on screen before converting; a mismatch refuses the drag ("the stream on screen changed; draw again") and learns the map again. Test added.
2. **Overlapping geometry answers** — taken. A generation counter keeps only the newest `/api/v1/osd` answer, whichever order they land in. Test added with two held answers landing newer-first.
3. **Armed mode swallowing presses on the bar** — taken. The capture-phase handler now excludes the same chrome zoom-to-area excludes (bar, pad, stats, toasts), so a press there reaches its control and the drag stays armed. Test added.
4. **Link labelled after the picker changed** — taken. `say()` is handed the camera that was asked and labels the link with it. Test added with the picker changed while the answer is held.

Issue #199, opened 2026-08-29
Scroll to the selected section in `Majestic → Settings` on mobile


usa-:
In the mobile view of **Majestic → Settings**, when a settings section is selected from the navigation accordion, the page should automatically scroll down so that the beginning of the selected section is positioned near the top of the screen.

This would make as much of the section's settings as possible visible immediately, instead of leaving the user somewhere in the middle of the page after selecting a section.

The goal is to make navigating to a section on mobile feel direct and predictable: select a section in the accordion, and its first settings are immediately visible.

openipc-ai:
Fix is up in #200.

Selecting a section now scrolls the form column to just below the top of the screen.
Measured on an hi3516av300 at 390×844: tapping **Watchdog** used to leave the card
starting 727px down an 844px screen — the whole viewport was still the tree. It now
starts at 16px, and all 22 sections land in the same place.

Two things came out of building it that you might care about:

* A short section could not reach the top at all, because the rail above it is most of
  a screen and there was nothing left to scroll with — the browser just stopped you
  part of the way down. The form column now reserves a screen below `md`, so the
  landing is the same whether the section has six fields or sixty. The cost is some
  empty space under a short card.
* Clearing the search box used to close the category you were reading. It doesn't any
  more, since scrolling down to a section makes scrolling back up to pick the next one
  a normal move.

One question, since it's your call rather than mine: **should the same jump happen on
a desktop browser?** I've kept it to phone widths, because from `md` up the tree is
pinned to the top of the window and never leaves the screen, so the problem you
reported can't happen there. The only desktop case it would improve is switching
sections while you're scrolled deep into a long one — today you keep your scroll
position and land somewhere in the middle of the new section. Say the word and I'll
lift the restriction; it's one line.

openipc-ai:
Merged as 03efff9 — this closed automatically with the merge, so here is where it landed.

To try it: `updatewebui` on the camera (or wait for the next nightly), then open
**Majestic → Settings** on a phone and pick a section. The page should bring the
section's card to just under the top of the screen, and the rest of the section is
below it rather than below the fold.

Three things worth knowing about how it behaves:

* Picking the section you are **already** on scrolls too. It is a deliberate tap, and
  a control that does nothing reads as broken.
* If you have unsaved changes and answer **Cancel** to the "discard and switch?"
  prompt, nothing moves at all — you stay exactly where you were, on the section you
  were editing.
[...]

PR #526, opened 2026-09-17
settings: a rail taller than the screen takes the form with it

openipc-ai:
Reported from the settings page: scroll the tree down to a section near the bottom — Watchdog, say — click it, and the right-hand side is empty. You have to scroll back up to read the setting you just asked for.

## What it is

`position: sticky` pins the **top** of a box and does nothing else. The rail is six categories and twenty-three sections — **1083px** against a 900px window — so its bottom is off screen at every scroll offset, and reaching it means scrolling the *page*, which carries the form column off the top with it.

The click then makes it worse. Measured on an hi3516av300 at 1280x900:

| | before | after |
|---|---|---|
| page scroll needed to reach Watchdog | 687 | **0** |
| where the card landed | top at **−282px** | top at 110px |
| of the 253px card, on screen | **0px** | 253px |
| document height | 2078 → 1292 | 999 |

The −282 is the browser clamping: Watchdog's document is 1292px where the previous section's was 2078, so the scroll position you were at no longer exists and it lands wherever it can. At 1920x1080 the rail is 3px over the window and the same fault shows in miniature — heading off screen, 150 of 253px visible. At 2560x1440 the rail fits and there is no fault at all, which is why it survived every check made on a big monitor.

## What this does

The tree takes what is left of the window and scrolls inside itself, while the title and the search box stay put above it — the box that filters the list should not be the first thing to leave when the list is scrolled. Browsing sections then moves the page not at all.

The cap is the window **less the rail's own distance from the top of the document**, not a bare `100dvh`: at scroll 0 the rail starts below the navbar, so a full-window cap hangs the tree's own last rows below the fold — the same fault in smaller clothes. That distance is not a constant (110px from 992 up, 130px at 768, more with a banner above the row), so `mj-settings.js` measures it into `--mj-rail-top`, reading the **column** rather than the rail: the rail is the sticky one, and once it has stuck its box is no longer where the document put it.

Two things the inner scroller then owes:

- The active section is scrolled into view **inside the rail**, since a search rebuilds the list and resets its `scrollTop` to zero. By `scrollTop` arithmetic, not `scrollIntoView({ block: 'nearest' })` — that walks every scrollable ancestor and would move the page too, which is the one thing the rail's own scroller exists to stop.
- The reveal a pick already did below md (#199) is now asked of the geometry instead of the breakpoint. Read a long section to the bottom, pick a short one, and the clamp still leaves the card above the fold — so the column comes back when its top edge is off the top of the window, and never jumps a section already in front of you.

`flex-wrap: nowrap` is load-bearing and was not obvious. Bootstrap's `.nav` is `flex-wrap: wrap`, which does nothing to a column of links until the column is given a height — and then wraps them into a **second column** rather than overflowing downwards. The moment the cap went in, the tree laid itself out 365px wide in a 281px pane, reported *no vertical overflow at all*, and put the last six sections off the right-hand edge.

## The rule, so it cannot come back quietly

`tools/check-layout.mjs` grows a check for the class: any `position: sticky` box taller than the window with nothing inside it that scrolls. An inner scroller is the fix, so a box that has one is not reported.

It fires on the page as it was and is silent on the page as it is:

```
  laptop 1280x900
    tall:    div.sticky-md-top: 1083px of sticky in a 900px window, and nothing inside it scrolls
```

and the self-test plants one, so the rule cannot go dead while the run stays green.

## Checks

- `npm test` — all passed.
- `tools/lint-templates.sh` — clean (against an `origin/master` control, since this environment's haserl reports one pre-existing exception on every branch).
- `tools/check-layout.mjs --self-test` — 5 of 5 rules fire.
- `tools/check-layout.mjs` at 390 / 1280 / 2560 on Watchdog, Day / Night, Main stream, Motion detection, OSD and Record — clean on every one.
- Walked by hand in Chromium on an hi3516av300: wheel over the tree (page does not move), click Watchdog (page does not move, card at the top, whole of it on screen); deep link `?tab=watchdog` (rail scrolls itself to the active row); search and clear (active row comes back into the rail's view); resize across the md breakpoint both ways; long section scrolled — the rail sticks without reaching the footer. Below md nothing changes: no cap, no inner scroller, and the pick still travels.


openipc-ai:
All three addressed in `8537b45`.

**New banners hide the rail's last rows** — real, and reproduced before fixing it. Injecting the card warning `storage-check.js` raises on a heartbeat moved the row down 74px while `--mj-rail-top` stayed at 110, so the rail's scrollport ended **74px past the bottom of the window** — this PR's own fault, one layer down. It is a `ResizeObserver` rather than a list of moments, because both notice slots are filled from something that *arrives* and emptied again when the camera recovers; and it observes the body rather than the two slots, because a list of the things that can move the row is a second copy of the page's structure and would be wrong the first time a page grew a third banner. The offset is written only when it has actually changed, so firing for everything else costs a rect and no write. Measured after: banner in → `110px → 184px`, two banners → `258px`, banner out → back to `110px`, with the rail's bottom staying exactly on the fold throughout.

**Rail placement can regress silently** — taken. The two decisions are now `www/a/mj-rail.js`, pure and asked by `tests/rail.test.js`, following the same reasoning `mj-fps.js` and `mj-place.js` are files: both fail silently and neither can be reproduced without a camera carrying enough sections to overflow a short window. The test pins the boundaries, which is where both go wrong — a row flush against an edge is *inside* and is not jumped at, a row taller than the pane shows its beginning rather than its end, a pane with nothing to scroll is left alone (which is what keeps this off the stacked rail below md), and the answer is clamped to the pane it is for. Four behavioural mutations were run against it — branch order swapped, the clamp dropped, the already-visible guard removed, the geometry branch removed — and all four fail the suite, so it is a test that can fail.

**Tall sticky defects can pass checks** — correct, and the exemption was wrong in principle rather than merely in scope. An inner scroller is the fix *only because it is what lets the box itself fit*; a scrollport inside a box that is still too tall has its own last rows below the fold, so a scrolling descendant is evidence of nothing. The rendered box is now the whole measurement, with nothing exempting it. The self-test's plant carries exactly the case you describe — a tall sticky holding an unrelated overflowing `<pre>` — and must still be reported. Run against the Live, Dashboard, Network, Recordings and SD card pages, the stricter rule finds nothing, so it costs no false positives.

PR #527, opened 2026-09-17
live: the other camera, live, in the rectangle you drew

widgetii:
The peer control answered a drag with a link that opened the other camera's crop in a new tab, under that camera's own login, and gave no hint of where that camera looked. Redone on the page itself.

**Where the other camera looks.** Switch Peer on and its outline is drawn on this picture, from `GET /api/v1/calibration/coverage`, laid out through the same stream map the rectangle goes through (so it is right on the sub stream too), asked again every few seconds so it follows the lens, and labelled with the camera's name.

**The loupe.** Draw a rectangle inside the outline and it becomes the other camera's live video of the same place, in the box you drew: the embeddable player is mounted inside the box and pointed at the other camera's origin with the media-scoped session this camera brokers for it (`GET /api/v1/calibration/peer`), sized and offset so the part you drew fills the box. Drag to move, drag the corner to resize; both re-ask the camera where the rectangle lands. The toolbar shows the whole of the other picture, swaps the two (the other camera fills the stage, this one becomes a corner inset painted from its own live picture with the outline and the rectangle on it), goes full screen, pops the video out with the browser's own picture-in-picture, or closes. Double-click swaps; Esc backs out one step. Without a player, or when none can play the stream, the other camera's snapshots are polled instead, cropped, and the tag says "snapshots" rather than "LIVE".

**Pairing, once, in the note.** When this camera is not yet paired with the other one, the note asks for the other camera's password once and sends it to this camera (`POST /api/v1/calibration/pair`), which keeps only the token it is handed; then the loupe opens. A refusal is said in the form.

**Under it.** Every video socket builder — WebRTC and data-channel signalling, MSE, the software rung — now goes through `MajesticSignal.endpoint()`, which takes an optional origin and session (each may be a function, read at every open, so a refreshed session is the one a reconnect carries) and otherwise answers the page's own host exactly as before. The embeddable player passes them through as two mount options. Nothing else changes for the page's own camera.

**Tests.** `tests/preview-endpoint.test.js` pins the URL building; `tests/preview-peer.test.js` is rewritten around a small DOM and a stand-in player: the control's gating, the outline's placement (main and sub stream), the drag, the crop arithmetic, whole, swap and the inset, pairing (wrong then right password), the snapshot fallback, Esc, and a stream change. Whole suite green. Driven in a headless browser against a camera pair running the matching daemon.


widgetii:
Review findings, one commit (`review: seven of eight taken`):

1. **Private addresses in the endpoint test** — taken; documentation addresses throughout.
2. **`expires | 0`** — taken. An expiry the camera did not give stays `null`; the refresh schedule falls back to the documented quarter of an hour only then.
3. **Switching peers with a loupe open** — taken. Picking another peer closes the loupe, whose session and map were the first one's.
4. **A late session answer** — taken. Acquiring a session now returns it rather than writing it into the loupe; the caller keeps it only if its operation is still the current one, closing supersedes everything in flight, and Esc cancels a loupe still being opened. Test added.
5. **A failed move keeps the old crop** — taken. The box goes back to the last place the camera answered, with the reason said. Test added.
6. **The software rung's feed** — already taken in the previous commit, as noted.
7. **ICE settings** — taken. This camera's configuration is kept from the one read at start and its `webrtc` section reaches the peer player. Test added.
8. **A previous peer's codec** — taken. Forgotten with the loupe it was learnt from. Test added.

PR #528, opened 2026-09-17
live: Peer says what to do next, and a click is a question too

widgetii:
Switched on, the peer control drew the other camera's outline and then waited for a rectangle nobody had been told to draw. Two small things:

- **The note says what to do** while the control is armed — "draw a rectangle inside the outline, or click in it" — and goes with the arming, or with the first answer.
- **A click is a question too.** The commonest one is "show me this", so a click inside the outline (a drag below the rubber band's floor) opens a loupe of a set size centred on the point, kept on the picture; the rectangle stays for when the size matters.

Both pinned in `tests/preview-peer.test.js`; whole suite green.


widgetii:
Both taken (`review: both taken`):

1. **A stream change cleared the hint** — restored when the control is still armed. Test added.
2. **Disarming could take the pairing form** — the form is an answer, not the hint, and is marked as one when it takes the note. Test added.

widgetii:
Superseded by the overlay: the loupe is gone altogether, see the next PR.

PR #529, opened 2026-09-17
live: the other camera's picture laid over this one where it belongs

widgetii:
The loupe (#527, refined in #528) was the wrong shape, and driving the page as an operator showed it: the box opened where the outline was and at the same size, so it magnified nothing; it covered the very region it showed; it was black for ten seconds and then half hidden under its own toolbar; and it wanted a rectangle nobody had been told to draw.

**Now.** Switch Peer on and the other camera's outline is drawn on the picture with everything outside it dimmed a little; the tag says what a click will do. Click inside and the other camera's live picture is warped onto exactly that quadrilateral — the embedded player's whole stage, put in place by one CSS perspective transform from the picture's corners to the outline's (Heckbert's square-to-quad, at full precision) — so it sits on this picture aligned with it on the plane the calibration was measured on. Its snapshot shows straight away, warped the same way, while the video connects; snapshots alone, polled, when nothing can play, and the tag says which. Click again, or Esc, and it is gone; Esc again disarms.

**The zoom controls work as they do with Peer off.** A press inside the outline stays the page's: a drag pans under Fill and 1:1, draws the zoom rectangle under Fit or with the area tool armed, and the overlay stays up through it; a release within a few pixels of the press is the click that toggles the overlay. The cursors are the page's own — grab, grabbing, crosshair — plus a hand inside the outline while the other camera's picture is not yet up. A double-click inside is the one thing kept from the page, since two clicks there already toggled the overlay twice. And the zoom goes as far as the *inner* picture's pixels allow: while the overlay is up, `preview-zoom.js` takes a ceiling hook and the peer module answers 3x of the inner picture's width, so the same truck fills the window here as it does on the other camera's own page; the tag states the inner picture's magnification the way the chip states the page's.

**Which of the other camera's streams.** The camera lists the peer's streams with the brokered session; the page watches the H.264 sub stream while the picture on screen is narrower than it is, and the main once the zoom makes it wider (hysteresis either side, three placements before a switch, the snapshot bridging the reconnect). The camera can answer the ask for the main with the sub — a browser whose WebRTC has no H.265 — and the player follows that answer; it is taken as final for the life of the player and the tag says "main stream not served here", rather than a reconnect once a second for ever. The player's own served-channel toast is hidden inside the overlay, where it would be warped with the picture and its dismissing click would hide the overlay.

**Sessions.** One per overlay, kept across clicks while it is good, refreshed before it runs out, let go when Peer is switched off or another peer is picked. When the player runs out of transports the page asks this camera for a fresh session once (`fresh=1`, which the camera answers by dropping its cache) and mounts the player again on it — the other camera has restarted and forgotten every session it minted, and nothing else can tell; a second loss is a real one and falls to the snapshots. `preview-zoom.js` no longer takes the peer player's frame for the page's own.

**Review findings, all three taken.** A click while the session is on its way is neither another ask nor a cancel; the pairing form's success comes back only while Peer is still armed and the peer still the one the form was offered for, and switching Peer off or picking another peer takes the form away; the Peer cursor rules that overrode the area tool's crosshair are gone.

**Gone.** The rectangle, the loupe, the toolbar, swap, pop-out, the hint toast, the fixed click box.

**Tests** (`tests/preview-peer.test.js`, rewritten and grown): the control's gating; the outline, its dimming and its placement on the main and a cropped sub stream; the click inside and outside, a drag inside left to the page, a double-click inside kept from it; the transform landing every corner of the picture on its corner of the outline, and still doing so when the decoder reports another frame size; the snapshot before the video; the zoom ceiling raised to 3x of the peer's picture and withdrawn with the overlay; the stream choice by size on screen with hysteresis and dwell, a refused ask remembered until a fresh player; the fresh session on a loss, once; everything that belongs to one overlay staying with it; pairing, and the form's lifetime; the snapshot fallback; a stream change. Whole suite green.

**Driven on a camera pair** at 1920×1080 in a headless browser: outline, click, live picture within three seconds; a drag inside pans, draws and zooms as with Peer off; eleven cursor readings as intended; the inner picture at 300% of its pixels at the ceiling and the page pulled back to 3x when the overlay goes; the main stream taken within a second of the picture passing 1280 px and the sub back on zooming out; the other camera's majestic killed and restarted under a playing overlay, the video back on a new session thirty seconds later without a reload. Alignment checked by blending the peer's picture over this camera's at half opacity: the cars on the calibrated plane single-edged, trees off the plane showing the expected parallax.

Needs a camera whose firmware answers the coverage and peer endpoints; on an older one the control stays hidden.


widgetii:
Two more commits from driving it on a camera pair at 1920×1080:

- **The snapshot stays until the video is a picture.** The software rung paints what it decodes from the first packet, which is a grey field until a keyframe arrives; the snapshot now stays on top for a short grace after "playing", and a keyframe is asked for at that moment.
- **The overlay watches the peer's native stream.** Measured in a browser without hardware HEVC: the peer's 2592×1944 H.265 main stream through the software decoder ran at 8–11 fps against 25 sent, dropped most frames and stayed grey for ten seconds. With the camera now listing the peer's streams beside the session, the overlay prefers an H.264 sub stream, which every browser decodes natively (measured: 24 fps, nothing dropped, picture within a second) and is more than an overlay a few hundred pixels wide can show; the main stream when there is nothing else.

widgetii:
Two more, both found by driving the page after the maintainer's report:

- **The outline moved on the click, and the two pictures did not align.** The zoom module learns the picture's frame from `loadedmetadata` and `resize` events caught on the way down through the stage, off whatever video fired them. The peer's video fired them too, so the moment the overlay came on the page took the peer's 1280×720 for its own 640×360, recomputed scale and placement, and everything on the picture moved — the outline and the overlay first. Only the elements the zoom module places, the ones carrying `.mj-stage-media`, define the frame now. Measured across the click: frame, scale and corners identical before and after, with the video playing.
- **Switching Peer off left the outline on screen.** The outline is an SVG element and `hidden` is HTMLElement's property; shown and hidden by style now, and the tests ask the style. The page also keeps its brokered session across clicks while it is good, so a click on, off and on again is one session on the peer rather than three.

widgetii:
One more commit: the overlay's stream now follows its size on screen.

The other camera's picture came from its H.264 sub stream at every zoom. Now its width on screen, in device pixels, is held against the sub stream's width: wider and the main stream is asked for; back under four fifths of it and the sub returns. Three placements (750 ms) have to agree before a switch, so a wheel gesture on its way does not flip it back and forth, and the snapshot bridges the reconnect as it does at switch-on. The tag names the size of the picture it shows.

The ask is a preference, and the camera can answer it with the other channel in the signalling — a browser whose WebRTC offers no H.265 is served the sub stream for an H.265 main. Left alone that was a reconnect once a second for ever: the player followed the answer, the module saw the sub while zoomed in and asked again. A channel the player was moved off after being asked for it is now that answer for as long as the player is mounted, and the tag says "main stream not served here". The player's own served-channel toast is hidden inside the overlay, where it would be warped with the picture and its dismissing click would hide the overlay.

Tests: zoomed past the threshold the main is asked for on the third placement, the snapshot placed at its size until the decoder says; a gesture that turns back asks nothing; zoomed out the sub returns; served the sub instead, it is asked for once and the refusal is remembered until a fresh player. Driven on a camera pair at 1920×1080 both ways: with an H.265 main the browser cannot decode, one ask, the tag saying so, no reconnects; with an H.264 main, the sub at 325 px on screen, the main at 2592×1944 within a second of the picture passing 1280 px, the sub back 1.5 s after zooming out.

widgetii:
One more: a press inside the outline stays the page's.

With Peer on, a press inside the outline was stopped at the stage so the click would not also pan, and arming Peer disarmed the area tool. That made the outline the one place where the zoom controls did not work: the page refused to draw a zoom rectangle over the picture the operator wanted to look at closely, and would not pan there either.

Now the press is only watched: a drag inside the outline pans, draws the zoom rectangle, or is the area tool's rectangle when that is armed, exactly as with Peer off, and the overlay stays up through it. A release within a few pixels of the press is still the click that toggles the overlay. A double-click inside is the one thing kept from the page, since two clicks there already toggled the overlay twice and the Fill-to-Fit jump must not move the view under them.

Driven on a camera pair at 1920×1080, main stream, overlay up: a drag inside pans under Fill; under Fit it draws the rectangle and zooms to it; with AREA armed it zooms to the rectangle; a click still toggles; a double-click inside leaves the view alone while one outside still toggles Fill and Fit. Nine checks, all green.

widgetii:
Follow-up to the press pass-through: the cursors now say what a press will do.

They still described the old rule, a stop sign outside the outline and a crosshair inside it, over a picture that in fact pans under Fill and 1:1 and draws under Fit or with the area tool armed. The Peer cursor rules are gone, so the page's own cursors show through with Peer on exactly as with it off. The one addition is a hand inside the outline while the other camera's picture is not yet up, for the click that brings it; once it is up the page's cursor returns, and the area tool's crosshair and a pan in progress win either way. The stage carries a class while the overlay is up, which is all the rule reads.

Read on the camera pair as the browser computes them: grab outside and inside with the overlay up, grabbing mid-pan, the hand inside with the overlay off under Fill and Fit, crosshair everywhere with AREA armed and under Fit outside, grab under 1:1. Eleven readings, all as intended.

widgetii:
One more: with the other camera's picture up, the zoom goes as far as its pixels allow.

The page's ceiling is 3x of its own stream, and the other camera's picture occupies a few hundred of those pixels, so at the ceiling it was shown below its native size: no magnification at all, where the same picture on that camera's own page reaches 3x.

`preview-zoom.js` takes a ceiling hook, a function answering a scale in the frame's own terms, read at every zoom, `null` to withdraw it; a free zoom already past a ceiling that leaves is pulled back to it about the middle of the stage. The peer module supplies it while the overlay is up: 3x of the peer frame's width over the outline's top edge measured in the served stream's pixels, so it is right on Main and Sub alike and a switch from the peer's sub to its main raises it on the spot. The tag states the inner picture's magnification the way the chip states the page's.

Tests: the ceiling is 3x of the peer's declared picture over the outline once the overlay is up, follows the decoded frame, and is withdrawn when the overlay goes; the tag carries the percentage. On the pair: the inner picture reaches 300% of its pixels at a page scale of 9.4x on the peer's sub stream (this headless browser has no H.265, so its main is refused), the ceiling holds under the wheel both ways, and a click that hides the overlay pulls the page back to 3x.

widgetii:
One more: a player refused with its session asks this camera for a fresh one, once.

The other camera can restart and forget every session it minted while this camera still answers the old one from its cache and the page still holds it, both with minutes left. The player ran its transports down and the overlay fell to snapshots riding the same dead session: "connecting…", then nothing, until the cache ran out.

Now when the player runs out of transports the page asks this camera for a fresh session once per overlay (`fresh=1`, which the camera answers by dropping its cache) and mounts the player again on it, the snapshot bridging the switch as at switch-on. A second loss is a real one and falls to the snapshots as before. Tests: the loss asks with `fresh=1`, the player is remounted on the new session with its snapshot, a second loss is the snapshots and no third player, the next overlay gets its own ask. On the pair: the other camera's majestic killed and restarted under a playing overlay, the video back on a new session thirty seconds later without a reload.

PR #534, opened 2026-09-19
live: say when the encoder is not meeting the rate it was set

widgetii:
## Why

A camera can run at several times its configured bitrate indefinitely and nothing on the page says so. The **Encoder out** tile shows the number, the chart draws it against the set rate, and an operator is left to notice the gap and guess why.

Cameras whose majestic publishes `venc*_rc_state` on `/metrics/venc` now answer that themselves — a verdict reached over a thirty-second window with its own hysteresis, the same one that goes into the camera log. This renders it.

Nothing here re-derives a judgement from the byte counter. A second opinion in the browser would smooth an already-smoothed number and disagree with the camera log the moment the two drifted, leaving two answers to one question.

## What

`www/a/rc-check.js` is the decision — DOM-free and tested, the way `video-check.js` and `ircut-check.js` are. Three surfaces read it:

- a line under the encoder chart on the Dashboard (`#st-enc-note`, beside `#st-luma-note` rather than inside the KPI tile — a second line in `.st-tile` lifts every tile in the grid with it);
- a shaded band above the set rate on that chart;
- a panel on the Video sections of Settings, beside the fields it names, mounted by a new `rcPanel(sec)` alongside `ircutPanel` / `ipeyePanel`.

The band is `cfg.bands` and not a second `cfg.ref`: `ref` is singular and sits outside the auto-scale pass, so a second level above the data would silently vanish.

## What it says, and why each branch is worded that way

**Over its rate.** A constant-bitrate channel that sustains far past its target has run out of quantiser; there is no other way it happens. So `maxQp` is named as the lever whenever there is one left to raise, and the mean quantiser is reported as *evidence* rather than used as a test.

The first draft did use it as a test — when the mean sat a few steps under the ceiling it announced there was compression in reserve and advised a higher bitrate. Pointing it at a real camera killed that: a gk7205v200 running 10 Mbit/s against a configured 128 kbit published a **mean of 27 against a ceiling of 30**, while the encoder's own rate-control counters had the working quantiser **pinned at 30 on every frame**. A mean averages the easy parts of a picture in, so it sits below a ceiling that is nonetheless binding — and the sentence sent somebody to the wrong setting with complete confidence, which is the failure this file exists to avoid.

At a ceiling of 51 there is nothing left to raise, so that branch names the bitrate, the frame size and the frame rate instead. That is a fact about the setting, not a threshold picked by eye.

**Starved.** Far fewer frames than `videoN.fps` asks for, which is a sensor on a long exposure and not a rate-control fault at all. It points at Image settings rather than Video — telling somebody whose encoder is being handed a third of a frame per second to change their bitrate is worse than saying nothing.

## What it does not say

A camera whose majestic does not publish the verdict renders **nothing**. Absent is never rounded to 0, because 0 here means *"within its configured rate"* — the one claim that would be a lie about a camera that never spoke. That silent path is the load-bearing one and it is tested both ways.

## Verification

`npm test` green (34 new checks in `tests/rc-check.test.js`), `tools/lint-templates.sh` clean, all three edited JS files pass `node --check`.

`tools/rc-live-check.mjs` points the module at a running camera and prints what each surface would show, so a rename on either side reads as silence there rather than as a green suite. Verified against both kinds of camera:

- one publishing the verdict → `rc_state 2`, and the sentence names `maxQp`;
- one whose majestic predates it → every gauge absent, both surfaces render nothing.


widgetii:
On finding 2, "Dashboard keeps a stale rate warning" — declining, with reasons.

The two-failure gate is this page's deliberate, documented pattern for every condition it displays, not an oversight in the new code. `onSample`'s failure branch clears `setLumaNote`, `renderNoVideo` and `renderIrcut` at `s.fails >= 2` and at no earlier point, and the comment above it says why: *"What is DISPLAYED still waits for the second failure, so a single dropped poll does not flap a banner off and on again."* `setEncNote` sits in that same block and behaves identically to its neighbours.

On Rule 3 specifically — *an absent reading is not a zero, and a failed fetch is not a fact*: nothing here presents the failed fetch as data. A failed sample carries no `m`, so it can never be read as a verdict; the note simply continues to show the last thing the camera actually said, for at most one poll interval (~2s), after which it is either refreshed by a good sample or cleared by the second failure. Saying "this is no longer current" is what `#st-alert-stale` is for, and it fires at the same threshold for every condition on the page.

Clearing at the first failure would reintroduce exactly the flapping the gate exists to prevent: one dropped poll would blink the warning off and back on two seconds later. That is the same defect class this PR's daemon-side counterpart spent a round of review eliminating, so introducing it here would be an odd trade.

Finding 1 was real and is fixed — `meanQp` and `maxQp` are gated together now, with coverage for the partial shape in both `diagnose()` and `note()`, including the mirror case of a mean without a ceiling.

PR #541, opened 2026-09-19
sdcard: redraw only what moved, so a poll stops taking the page away

widgetii:
The SD card page polls every five seconds and rewrote its whole body each time, whether or not anything had changed.

Nothing about that shows in a screenshot — the markup is right and the numbers are right. The cost is paid only by somebody actually *using* the page: the focus ring leaves the button you were about to press, the mountpoint you were half way through selecting goes away, the `title` tooltip you were reading closes, and a running scan's progress bar restarts its transition. Every five seconds, for ever.

It is also what has kept anything stateful off this page. A tab, a chart or a `<details>` cannot live inside a container thrown away on a timer — which is exactly why the two dialogs are declared outside `#sd`, and `sdcard.cgi` says so in a comment.

## What changed

`render()` now asks whether the page actually moved.

- **`build(d, L)`** is the old `render()` body made pure: it takes the card and a map of live chunks, returns markup, and touches nothing. So it can be run twice.
- It is run once with those chunks **blanked** — that string is the page's *shape*.
- Shape unchanged → only the chunks are written, into the `[data-mj]` blocks that hold them.
- Shape changed (a card went away, a button became available, a verdict turned) → the page is rebuilt, which is honest, because something really is different.

The three live chunks are the ones that move on their own while the shape does not: the storage figures, the recorder's counters, and a running scan's bar. `hasLive` rides along with them because it is *presence* rather than value — blanking the chunks must not take the note that belongs to them with it.

The heartbeat drives rendering now, every two seconds rather than every five. That was not safe before (twice the rebuilds, twice the lost selections) and it is what the counters deserve, since they are precisely the half of the page that can move without its shape moving.

## Rendered output is unchanged

This is the whole review question, so it was measured rather than asserted. The previous revision and this one were each run against four card states — healthy, read-only, no card, mid-scan — and the markup compared:

| | result |
|---|---|
| visible text (every word, number, sentence) | **identical** in all four |
| tag-level additions | exactly `<div data-mj="storage">`, `<div data-mj="scan">`, their two closers, and `data-mj="perfrows"` on the existing `<dl>` |
| byte delta | +74 in every state that draws a card; 0 for "no card" |

No new Bootstrap classes, so `bootstrap.min.css` regenerates unchanged.

## Tests

`tests/sdcard-view.test.js` is new. The fault it pins is silent in production and cannot be reproduced on demand — it needs a camera answering two endpoints on two clocks, and only shows if you happen to be touching the page when a poll lands. So what is pinned is the rule:

- a settled page takes three identical heartbeats without rewriting anything
- a write rate that moved is painted into the counters block, and the page around it is not touched
- a build that stops reporting its recorder **does** rebuild — refusing to redraw when something really changed is the worse failure
- a camera that has recorded nothing still shows no counters after a paint, not a row of reassuring noughts

The existing suite passes untouched, including all eight `sdcard-*` / `storage-verdict` files.

This is the first of two: it is deliberately invisible, so that the layout work that follows can be reviewed on its own.


widgetii:
Fixed in 974cec3 — the finding was real and the reasoning was exactly right.

`load()`'s catch wrote the failure notice straight into the page, bypassing `render()`. That left `lastShape` describing a page nobody could see, so the next poll to **succeed** found the shape unchanged, took the paint path, looked for live blocks the notice had replaced, found none, and wrote nothing. The camera was back and the page went on saying it was unreachable for as long as anyone left it open.

Rather than invalidating the cache in the catch, the failure is now a state `render()` owns: `loadErr` is set, `build()` draws the notice from it (without the page's own head, exactly as before), and **nothing writes to the page except `render()`**. That also fixes the half of it the cache reset would not have: a heartbeat arriving during the outage now re-derives the same notice, matches, and leaves it alone, instead of repainting stale figures over it.

`tests/sdcard-view.test.js` covers the round trip — draw, lose the camera, get it back. Against the previous commit it fails with `never saw "Performance"`, which is the bug as described.

Re-checked after the fix: the rendered-output comparison across all four card states is still text-identical to master.

PR #542, opened 2026-09-19
sdcard: lead with the verdict, and stop stranding the readouts

widgetii:
Follows #541, which made this possible: the page no longer rewrites itself on a timer, so a tab and a chart can live on it.

## The reported fault

`dl.list` is a grid of `auto 1fr` with the value right-aligned. That is right in the ~500px card the rule was written for. In a full-width one it put **`Longest pause` at one end of the screen and `305 ms` at the other**, with a metre of nothing between — up to ten rows of it in the Performance card. The same two words also appeared **twice, 180px apart**, meaning the recorder's worst flush in one place and the speed test's worst write chunk in the other.

`#page-sdcard` had no CSS at all, which is the sanctioned hook for exactly this. `dl.list` is narrowed there rather than globally — `network.cgi`, `update.cgi` and `backup.cgi` all want the original.

## Lead with the verdict

#536 gave the page a health verdict, a span probe and a read-back walk, then buried all three in the third of four cards of unrelated shape. The sentence worth reading first — *"This card has a problem."* — was an 11px grey `<span>` beside a heading.

The hero is now the mark, that sentence, what the card **is** on one line, the storage bar, and the three lines of evidence — each with the action that settles it on its own head line rather than below a paragraph that on the counterfeit verdict runs to four lines. The five-row identity `dl` is that one line; the ratings are chips.

**Only the sentence and the three lines need `sdcard-health.js`.** The identity, badge, storage bar and actions do not, and no longer go with them — a build that has not shipped the module is not a camera whose card has no model number. Same reasoning as the pin map's hidden fields.

## Tabs that cost nothing

Which tab is open is deliberately **not** part of the page's shape. If it were, pressing a tab would change the shape and the heartbeat two seconds later would rebuild the page, taking the focus off the tab just pressed. All three panels are in the markup; the tab is applied to the DOM. Switching shows one and hides two, loses nothing, and leaves every word findable with the browser's own find.

## Charts, and what is honest to plot

Neither trace is the longest pause. `records_fsync_us_max` is a high-water mark since boot, so plotting it over time draws a staircase that only goes up and says nothing about now — it stays a figure in the rows. What moves is the **write rate** and the **number of clips waiting**, and the queue is the one warning that arrives while there is still time to act on it. Both share the Dashboard's chart primitives; both are absent on a build with no recorder counters rather than drawn as flat lines at zero.

## The deletion line, and when not to draw it

The storage bar gains the line the recorder deletes at — the thing that explains a card sitting at 94% for ever without filling up. It is a promise about future writes, so it is withheld where there are none: a read-only card, or one the span probe found **wrapped**, reports free space that is fiction. Verified in both states.

The "about 52 min of footage left" figure is **not** here: `recordings.js` derives it from the day's clip list, which this page does not have. The only rate available here is the instantaneous heartbeat, which would swing every two seconds. It stays where it is computed properly.

## Dialogs

Both were forms.

**Format** now says what it does in the order it does it — the capacity check runs *first*, and that is the step that catches a counterfeit card, which is what the health panel means when it points at Format. Its `confirm()` box is gone: the acknowledgement is in the dialog it was launched from, where it can name the recordings being erased, and it is a real `<label><input type=checkbox>` because a confirm box was at least reachable from the keyboard. A failed format re-arms it.

**Change the card** is `mj-modal-lg` with the six steps down the dialog instead of wrapped across it into three ragged lines, each ticked as it passes and carrying its own live line — where somebody holding a card is actually looking.

## Tests

`tests/sdcard-view.test.js` gains two groups, both pinning things that fail silently:

- the page still says what the card is on a build with no health module, and claims no verdict it cannot support
- all three panels are in the markup, pressing a tab rewrites nothing, **and neither does the heartbeat after it** — which is the property that keeps focus on the tab you just pressed

Every existing `sdcard-*` and `storage-verdict` test passes untouched, including the ~60 string assertions in `sdcard-perf` — every user-visible string and every formatter is unchanged.

`npm test`, `tools/lint-templates.sh` and `tools/regen-bootstrap-css.sh` all clean.

Design canvas the layout was drawn against, including a faithful recreation of the current page for comparison: https://claude.ai/code/artifact/107043b3-d996-40fa-ada8-3f8a22278369


widgetii:
Two of the three fixed in fd23925; the third is declined with evidence.

**1. Unused cards promise automatic deletion — real, fixed.** The bar drew the line from the global `records.maxUsage` whenever the card was healthy, without asking whether the recorder is enabled or pointed at *this* card. A spare card in the slot was told its oldest clips go at 95% when nothing will ever sweep it.

There is one notion of "the recorder uses this card" now — `configuredHere()` — and three things ask it: whether to say "recording to this card", whether to offer the button that makes it so, and whether the bar may draw that line. It also corrects the Recording panel, which compared `records.path` to the mountpoint *exactly*, so a camera recording into a subdirectory of the card was offered a button to start doing what it was already doing.

**3. Older cameras show empty chart panels — real, fixed.** Both charts were gated on `records_state` existing, but the traces come from two *other* counters. Each chart is now gated on its own, and a panel with neither is not drawn at all. This is the page's own rule — an absent reading is not a zero — and drawing a labelled box that never gets a line is that mistake in a bigger box.

Both covered by new cases in `tests/sdcard-view.test.js`.

**2. Private tracker details become public — declined.** Four things, any one of which settles it:

- `OpenIPC/firmware` is **public** — 2201 stars, and `repos/OpenIPC/firmware` reports `private=false, visibility=public`.
- Issue 1747 is reachable and titled *"SSC338Q - Slow sdcard"* — exactly what the comment cites it for.
- The line is **already on master**, at `www/a/sdcard.js:227`. This branch only moved it within the file, from `ratingRow`'s comment to `ratingPills`', because it is the same reasoning about the same thing. Nothing is exposed that was not already public in this file.
- `CLAUDE.md` prescribes this form: *"Cite the issue that reported the fault (`(#273)`, `OpenIPC/majestic#311`) — a document that stays readable and says what the code cannot."* Two other files on master carry the same style of citation.

widgetii:
On finding 2 — I've dropped it, though not for quite the reason given.

The letter of the rule does not seem to cover it: `OpenIPC/firmware` is verifiably **public** (`private=false, visibility=public`, 2201 stars), issue 1747 is reachable and titled *"SSC338Q - Slow sdcard"* — exactly what the comment cites it for. The checklist's failure criterion names *"a link or #number reference into a **private** tracker"*, and its success criteria explicitly permit *"an issue in the public OpenIPC/majestic tracker"*. The private repository the rule is about is the majestic daemon's source, which is a different thing from the public firmware build tree. The line was also already on master at `www/a/sdcard.js:227` — this branch only moved it between two comments.

But it is a compliance-labelled rule on a public repository, the sentence loses nothing without the pointer, and arguing the letter of a disclosure rule in order to keep a citation is a poor trade. `CLAUDE.md` offers the alternative outright: *"cite the issue … or cite nothing and let the sentence stand on its own."* So it stands on its own now.

Scope note: the same citation appears **three times on master** (`www/a/sdcard.js`, `www/a/storage-check.js`, `www/cgi-bin/j/sdcard.cgi`). I have changed only the line this branch already touched. Sweeping files this PR has no other reason to open is not its business, and whether the other two should go is a call for whoever owns that rule — happy to do it in a follow-up if you want them gone.

Issue #316, opened 2026-09-04
Image settings: mirror/flip options are in different places and the description is outdated


usa-:
It feels a bit illogical that the mirror options are located under **Image → Live adjustments**, while the rotation options are under **Image → Image**.

Also, the **Image → Image** page says **“Brightness, contrast, saturation, hue, mirror and flip”**, but there are no mirror or flip settings there.

openipc-ai:
@usa- thank you — you were right on both counts, and they turned out to be one mistake.

**What was wrong.** The settings page decided *where* to draw a setting from a flag in the camera's schema (`x-live`) that only says a change can be applied without restarting the video pipeline. Mirror and flip carry that flag, so they were lifted onto Live adjustments beside the picture. A quarter turn (90° / 270°) does not — it swaps the stream's width and height, so it needs a pipeline reload — and it stayed behind on a two-row Image page. The sentence you quoted was a note that page composed from the names of the six settings that had been lifted away from it; it led with the list, so it read as a description of the page rather than as a pointer away from it.

**What changed** — merged in #346, in `master` now:

- **Image → Live adjustments is the whole Image section.** Under *Image* the rail now reads *Live adjustments / ISP / Day-Night*; the separate Image page is gone, and an old `?tab=image` bookmark lands on Live adjustments.
- **Orientation is one column:** the pad (Normal / Mirror / Flip / 180°) plus a **Quarter turn** row (None / 90° / 270°) right under it. Mirror and flip still apply the moment you press them. A quarter turn cannot be previewed — the camera has to rebuild the stream for it — so it is staged like any other setting and takes effect on Save, with the reload Save then asks for. The row says so, rather than leaving you to wonder why the picture did not move.
- **Automatic tuning**, the one remaining Image setting, sits in a small Image card under the deck.
- **Stock** in the tone strip now resets only the four tone knobs. It used to put the picture the right way up as well, which nobody pressing "Stock" on Brightness was asking for.

**Something you did not report, fixed by the same change.** The same rule silently dropped the **Bitrate** row from *Main stream* and *Sub stream* on any camera running a majestic build from 2026-09-02 onwards, because majestic began flagging the bitrate the same way. It is back, staged like before. There is now a test that checks every setting in the schema lands on exactly one page, so that class of disappearance fails the build instead of the camera.

**To try it:** run `updatewebui` on the camera (it installs `master`), or take the next nightly firmware. No newer majestic is needed; on an older one nothing here behaves differently.

[...]

PR #543, opened 2026-09-19
docs: CLAUDE.md carries the rules, the subsystems carry their own

openipc-ai:
`CLAUDE.md` had reached **151,025 chars**, past the 150k limit Claude Code truncates at — so the tail of a file whose only job is to be read was at risk of not being read. It had grown 115k → 151k in six days, because every feature PR appends its own essay, so a trim that only got under the limit would have bought about four days.

## What the noise was, measured

- **The per-subsystem essays were a second telling of the source comments.** 26k chars of the file's long sentences shared >25% of their 7-word shingles with comments in the tree. Nine of nine narrated subsystems sampled — `sbin/sdscan`, `ircut-check.js`, `preview-zoom.js`, `mj-tree.js`, `mjcrypto.js`, `sdcard-health.js`, `motion-notify.sh`, `sbin/updatewebui`, `p/pages.cgi` — already carried the same reasoning in their own header, where it is read at the moment it matters. `p/pages.cgi` repeated its passage almost word for word.
- **Coverage was inverted — an anthology, not a map.** 44 of the 83 modules in `www/a/` were never named at all (`network.js`, `logs.js`, `files.js`, `mj-region.js`, `pin-sweep.js`, …), while the IR-cut bullet alone ran 12.3k chars.
- **Stale.** "What this is" still described the suite as two test files. There are 86.
- **One principle restated up to eight times** in eight subsystem voices — the three-answer rule (*known* / *known-absent* / *could not ask*) appeared in the heartbeat, `notify.js`, `mbr_ok`, `ircut-check`, `gpio.cgi`, `mj_ptz`, `sdcard-health` and the conventions list.
- **263 bold runs across 88 bullets**, so no emphasis stood out.

## The rule that decided every cut

> `CLAUDE.md` carries what an agent needs **before** it opens a file: where things are, the cross-cutting rules, the contracts between parts, and traps invisible at the call site. The per-subsystem *why* stays at the point of use.

Six subsystems span too many files for any one header to own the contract; those get a document of their own, read on demand rather than loaded every session.

| | chars |
|---|---|
| `CLAUDE.md` | **151,025 → 26,291** |
| `docs/live-player.md` | 17,917 |
| `docs/settings-page.md` | 13,552 |
| `docs/notifications.md` | 8,434 |
| `docs/storage.md` | 6,307 |
| `docs/day-night.md` | 5,625 |
| `docs/deploying.md` | 4,660 |
| `docs/fpv-and-plates.md` | 2,337 |

`CLAUDE.md` keeps what applies to every session: what this is, the authentication and claim flow with the EULA policy, a repository map that also names the seven docs and says to read the matching one first, one consolidated **Rules** section, and the architecture contracts — request lifecycle, `common.cgi` helpers, persistent state and the majestic API.

## Nothing was cut on the assumption the source said it

Every removed narrative was checked against its destination first. All 110 normative sentences in the old file were then re-checked against the new tree; the sixteen with no lexical overlap are either deliberate history (`get_night`, `--no-self-update`, `header_fpv.cgi`) or live at the point of use, verified there — `ircut-check.js:944` for the single-pin verdict, `sbin/sdscan` for median-relative bucketing, `mj-settings.js:2918` for the Backspace guard. One genuine gap surfaced and was restored: *the haserl never reads the config and never handles a POST*.

Also gone, since the WebUI and majestic ship together: every "older majestic falls back to…" passage and the "Requires a majestic with X" notes beside them. Where one carried a real invariant it is re-founded on the reason rather than the build — `stillSet()` re-reads after a save because a save that silently failed to disconnect a coil would state the opposite of what the feature guarantees, not because an older daemon ignores a `null` leaf.

## Two bad cross-repo references, fixed rather than carried across

Both were pre-existing in `CLAUDE.md` and would have landed in new files otherwise:

- One reference cited for the motion bounding box pointed at a tracker number that does not exist in the public `OpenIPC/majestic` tracker, so it resolved for nobody who could read it. The sentence now says the fix landed firmware-side and names no tracker.
- **`OpenIPC/majestic#311`**, cited twice for the Network form, resolves to a transferred issue about an RTMP memory leak, so it sent a reader somewhere unrelated. That sentence now stands on its own, and the convention that quoted it as an example asks for a reference that actually resolves.

The ten remaining issue numbers (`#171 #189 #202 #227 #273 #288 #302 #316 #367 #478`) were each checked against this repository's tracker and all resolve.

## Verification

- Every backticked repo path in the new tree exists; the only misses are camera-side runtime paths and buildroot's `majestic-webui.mk`.
- `tools/lint-templates.sh`, `.pr_agent.toml` and `pr_compliance_checklist.yaml` all cite conventions still present in `CLAUDE.md`.
- `docs/` never reaches the camera: `tools/build-dist.sh` copies `www` and `LICENSE` by name, so there is no flash cost.


openipc-ai:
All three findings addressed in 98f79dd — both code findings verified against the source first.

**2. Player guidance skips a fallback — correct, and my regression.** The old `CLAUDE.md` said `WebRTC → MSE → MJPEG → note` in one place while saying elsewhere that MSE falls to the software decoder on a refused codec; compressing the two kept the wrong half. `preview-chain.js` states the real walk at the top of the file: `WebRTC → MSE → [software decode] → the caller's floor`, with MJPEG being the **Live page's floor** rather than a rung — the settings preview shows an alert at the same point. Restored along with the two properties that explain why the rung gets no radio: it is the same `/ws/video` bytes decoded in WebAssembly rather than a transport, and which rungs below MSE are open is read through `preview-transport.js` rather than off the decoder module, because the tests drive those gates from a stub.

**3. FPV guidance points to unused locale data — correct.** `wfb.cgi` does go through `p/common.cgi` and `p/header.cgi`, but neither of those reads `j/locale.cgi` and neither does `wfb.cgi`; it builds its labels from its own `wfb_*` variables. The only readers are `camera.cgi`, `p/player.cgi` and `www/a/preview-page.js`. The claim is inherited from the old file and would have sent somebody editing an FPV label to a file that cannot affect it.

**1. Private tracker number exposed — already fixed before the review ran**, which is why `docs/notifications.md` no longer carries the citation. The description and the commit message both contain no such number now. What cannot be undone is the description's edit history, which GitHub keeps visible; recreating the pull request would not purge it either, since a closed one stays readable. Noted rather than silently closed.

openipc-ai:
On finding 1, for the record before merge.

Every surface I control is clean: the PR title, the PR description, my comments and the branch's commit messages all carry no such identifier. The single occurrence in the diff is the **removal** line — this PR deletes it.

Three residuals are outside what I can act on, and none is a reason to hold the merge:

1. GitHub keeps the description's edit history visible. There is no API to purge it, and recreating the PR would not help, since a closed one stays readable.
2. The review comment above quotes the identifier itself. Deleting a review record to hide a finding would be worse than the finding.
3. **It is already on public `master` and has been since it was written.** The gate reads the diff and sees the string; what the diff is doing is taking it *off* the default branch. Blocking here would leave it published.

So merging is strictly the lower-exposure option. Findings 2 and 3 were real errors in the compression and are fixed in 98f79dd.

Issue #548, opened 2026-09-20
Move IR-CUT FILTER and CAMERA LIGHT settings to SWITCHING


usa-:
On **Camera → Settings → Image → Day / Night**, the **IR-CUT FILTER** and **CAMERA LIGHT** sections currently contain only one setting each, with the same name as their respective section.

These settings could be moved to the **SWITCHING** section, directly below **AUTOMATIC DAY/NIGHT**.

This would make the layout more logical, since these two settings define whether the IR-cut filter and camera light follow the **Automatic Day/Night** switching.

openipc-ai:
@usa- — you were right, and there was a little more wrong in that section than you could see from the page. Fixed in #550, now on master.

**Both headings are gone.** *IR-cut filter* and *Camera light* now sit together under one heading, **DRIVEN BY DAY/NIGHT** — the question they both answer: does day and night move this, do you, or is it left alone. Neither control is introduced any more by a heading repeating its own name.

**Where they went is not quite what you asked for, and I would rather say why than let you find out.** You suggested putting them inside SWITCHING, directly under *Automatic day/night*. I built exactly that and rendered it on a camera before deciding anything. The section is laid out in two columns and the page prefers to break between headings; with those two controls added, SWITCHING becomes tall enough that no break between headings will serve, so the break falls *inside* it and the second column opens on *"Gain multiple that means day again"* with no heading over it. That is a fault we fixed once already (#325) and I did not want to put it back. A heading of their own, immediately after SWITCHING, keeps every column starting on a heading at phone, laptop and monitor widths.

**The same look turned up something nobody had reported.** *"Pause inside a day/night switch"* was being drawn under CAMERA LIGHT, where it has no business — it is the gap between the picture going grey and the IR-cut filter swinging, nothing to do with the lamp. It sat there because it is newer than the list that decides the headings, and anything missing from that list falls in under whichever heading came last. It is under SWITCHING now, and the test that was supposed to catch that had gone blind, so that is mended too.

How the section reads now:

```
 SWITCHING                               DRIVEN BY DAY/NIGHT
   Automatic day/night                     IR-cut filter
   Legacy settings                         Camera light
   Gain multiple that means night        NIGHT PICTURE
   Gain multiple that means day again      Colorless night mode
   Seconds of darkness before night        DRC value in night mode
   Seconds of brightness before day
   Seconds between light checks
   Pause inside a day/night switch (ms)
```

[...]

PR #550, opened 2026-09-20
Day / Night: one heading for the two things day and night moves

openipc-ai:
On **Camera → Settings → Image → Day / Night**, two headings each introduce a
single row carrying the same words as the heading above it — **IR-CUT FILTER**
over *IR-cut filter*, **CAMERA LIGHT** over *Camera light*. Reported in #548,
and real. The second is worse than the reporter can see: on a build whose daemon
carries the PWM lamp keys that group has five rows, so only the filter's half of
it is universal.

The filter and the lamp each answer one question, and it is the same question:
does day and night move this, do you, or is it left alone. That shared question
is the heading now, and the two rows sit under it together.

### What was asked for was measured rather than argued

The report asks for both rows **inside SWITCHING, under Automatic day/night**. I
rendered that on a camera before writing it off. The group then grows too tall
for any cut to fall between groups — `layoutCols()` abandons the between-group
preference once the best one still leaves the taller column past 75% of the
section — and the second column opens on *"Gain multiple that means day again"*
with no heading over it. That is the fault #325 was about.

Measured at 1280 px (same outcome at 2560):

| arrangement | col0 | col1 | second column starts with |
|---|---|---|---|
| today | 1098 | 598 | `## NIGHT PICTURE` |
| both rows inside Switching, as asked | 780 | 829 | **a row, no heading** |
| both rows inside Switching + a heading for the automatic parameters | 527 | 1133 | `## WHEN IT SWITCHES` |
| **this PR** | 1017 | 643 | `## DRIVEN BY DAY/NIGHT` |

@usa- — the heading wording is mine, not yours, and so is the decision not to put
the rows inside SWITCHING. If *Follows day/night* reads better than *Driven by
day/night*, say so and I will change it.

### And a row was already under the wrong heading

*"Pause inside a day/night switch"* is the gap between the picture going grey and
the IR-cut filter swinging, and it was drawn under **CAMERA LIGHT**. A key named
in no group renders after the last group with no heading of its own, so a setting
the daemon adds inherits whatever heading came last. It joins Switching, which is
what it is about. Left alone it would have moved under *Night picture* with this
change, so it is not separable from it.

`tests/tree.test.js` carries an orphan check written for exactly this, but reads
a schema fixture that predates the key — so the key goes into the fixture too, or
the guard stays inert for the next one as well.

### After

```
 ## SWITCHING                            ## DRIVEN BY DAY/NIGHT
    Automatic day/night                     IR-cut filter    [auto v]
    Legacy settings                         Camera light     [auto v]
    Gain multiple that means night       ## NIGHT PICTURE
    Gain multiple that means day again      Colorless night mode
    Seconds of darkness before night        DRC value in night mode
    Seconds of brightness before day
    Seconds between light checks
    Pause inside a day/night switch (ms)
```

`CAMERA LIGHT` keeps its heading for the PWM lamp settings and is dropped rather
than drawn empty on a build without them, which is what it already did for those
same keys.

### Verified on an hi3516av300

- the orphan check fails on the fixture change alone and passes after the
  regrouping — the guard was seen to fire before it was trusted
- every column starts with a heading at 390, 1280 and 2560
- `tools/check-layout.mjs --open "Day / Night"` clean at all three widths, after
  its `--self-test` fired every rule
- the Legacy switch still sits directly above the rows it toggles, still shows
  one mechanism at a time, and still empties both thresholds
- the filter mode raises the save bar, saves, and reads back from the camera
- the panel above the form renders identically before and after in all three
  filter modes

Refs #548


openipc-ai:
Both review findings read, one of them fixed.

**1. A comment preserves review debate — fixed** in 3cb548a. The sentence did
record the exchange rather than the constraint: "what was asked for" and
"measured instead of argued" point at a discussion that is not in a checkout,
and give the reader nothing to check. It now states the mechanism instead —
`layoutCols()` drops its between-group preference once the best such cut still
leaves the taller column past 75% of the section, and the second column then
opens on a row with no heading over it.

**2. Two settings miss the required group — deliberate, and reconciled here
rather than restored.** #548 asks for both rows inside SWITCHING under Automatic
day/night. Rendering that on a camera is what ruled it out: the group grows past
the threshold above, and the second column opens on *"Gain multiple that means
day again"* with no heading — the #325 regression, measured at 1280 and 2560 and
tabled in the PR description.

What the issue actually reports is a heading that introduces a single row
repeating its own name, twice over. That is fixed: neither heading exists any
more, and the two rows share one heading that says what they have in common. The
placement was the reporter's proposed remedy, not the complaint, and the
remaining difference is where the shared heading sits rather than whether the
rows are still stranded under their own names.

The reporter is asked directly in the PR description to say if the wording or
the placement should be different.

PR #553, opened 2026-09-20
Day / Night: the filter hunt is back beside the wiring it sets

### LINKED ISSUE #552: Day / Night: finding the IR-cut pins is three presses and another page away
Reported by users against the Day / Night page, and it is a regression I
introduced in #510 (merged 16 Sep). Reproduced and measured on an hi3516av300
running current master.

## What it was

Until #510 the Day / Night panel carried a **primary** button, `Find them for
me`, beside the pad map. Pressing it opened the sweep's confirm card in place,
at the top of the same panel; the second press was `Start`. Two presses, no
navigation, both buttons on screen when it was their turn.

## What it is now

The button is `Find them on the Pins page`, `btn-outline-secondary` — the same
weight as `Test the filter` sitting next to it, so nothing says one of these
starts a search and the other tests wiring. Pressing it loads
`camera.cgi?tab=pins` at the top and stops there. From there the reader has to
find a section called **Find a pin**, choose between two doors, and press
`Which pins move the day/night filter?`, which draws a confirm card whose
`Start` button is below the fold.

Three presses, a page change, and measured at three viewports:

| viewport | lands at | filter-hunt door | `Start` card |
|---|---|---|---|
| 1280×900 | top | y=816, just in view | y=1008, **off screen** |
| 1280×720 | top | y=816, **off screen** | y=1008, **off screen** |
| 390×844 | scrollY=649, mid chip drawing | y=2299, **off screen** | y=2582, **off screen** |

Neither the arrival nor the door scrolls anything into view, so on a short
laptop and on a phone the reader presses a button and sees nothing happen —
twice. On the phone the previous page's scroll position is carried over, so
they land in the middle of the chip drawing rather than at the top of it.

## Why the move does not hold for this hunt

#510 split the two hunts on the argument that driving pads needs the chip's
own context — the pads that must not be driven, the ones already carrying
something, the record that survives a reboot a pad caused — and that this
context is the Pins page's subject.

That is right for the **general** hunt, whose detector is the device
inventory and which walks the whole chip. It is not right for the **filter**
hunt:

- Its detector is the picture. The only thing it can find is an IR-cut filter.
  It is not a general GPIO tool and cannot be used as one.
- The context the argument rests on **is already on the Day / Night page**.
  `ircut-map.js` draws the avoid list, the pads the kernel holds, the pads
  assigned to another role, and it says which of those it could not check.
- Its result belongs to Day / Night. #510 had to build a `sessionStorage`
  hand-over so a find on the Pins page could be staged into the Day / Night
  fields — a hand-over that exists only because the hunt was moved away from
  the form its answer goes into.
- The person who wants it is on Day / Night, looking at four roles reading
  *not set* and, often, a magenta picture.

## Proposed fix

`pin-hunt.js` already takes the host page's handle as an injected `pins`
object (`sweep`, `select`, `refresh`, `changed`) and never reaches into that
page's DOM, which is exactly the shape needed to mount it twice.

- Mount the **filter** hunt back on Day / Night, in place, driving the pad map
  that is already there, with the find staged straight into the hidden pin
  fields the way it was before — nothing written behind anyone's back.
- The Pins page keeps both doors, so somebody who arrived there looking for a
  filter is still offered it.
- One implementation, two mounts; the pure half stays in `ircut-scan.js`.

If the split is kept instead, the minimum is that the button must land the
reader **on the hunt with the filter door already open and in view**, not at
the top of a page about soldering.

## One dead leftover found on the way

`www/a/mj-settings.js:8073` renders `<div id="mj-ircut-find-why" hidden>`.
Nothing in the tree writes to it or unhides it — it is what is left of the
sweep's "why this is refused" line, which went with the sweep. It should go
too.

openipc-ai:
Closes #552.

The IR-cut pins cannot be found in one press on Day / Night any more, because
#510 moved the hunt to the Pins page and left behind a button that only
navigates there. Measured on an hi3516av300, that route is three presses and a
page change, and two of those presses visibly do nothing.

| viewport | lands at | filter door | `Start` card |
|---|---|---|---|
| 1280×900 | top | y=816, just in view | y=1008, **off screen** |
| 1280×720 | top | y=816, **off screen** | y=1008, **off screen** |
| 390×844 | scrollY=649, mid chip drawing | y=2299, **off screen** | y=2582, **off screen** |

Nothing scrolls on arrival or on the door, so on a short window and on a phone
the reader presses a button and sees no change, twice in a row.

## The split was by page where it should have been by detector

#510's argument is that driving pads needs the chip's context — what must not be
driven, what is already carrying something, the journal that survives the reboot
a pad caused. That is right for the **general** hunt, which holds one pin and
diffs the camera's device inventory: it can find anything, so it belongs with
the chip.

It is not right for this one. This hunt pulses **pairs** and reads the
**picture**, so an IR-cut filter is the only thing it can ever find — and every
piece of that context is already on the Day / Night page, where `ircut-map.js`
draws the avoid list, the pads the kernel holds, the pads another role has, and
says what it could not check.

The strongest argument against the move is the one the move itself had to build:
a `sessionStorage` hand-over, so a find made on the Pins page could be staged
into the wiring fields on this one. That exists only because the hunt was moved
away from the form its answer goes into.

## What this does

`pin-hunt.js` takes `only: 'filter'` and is mounted twice. The Pins page still
offers both doors for somebody who arrives there. What differs between the two
mounts is the door and where a find lands — the sweep, the confirm card, the
journal and the resume are one copy, and the pure half stays in `ircut-scan.js`.

- **A find is staged, not carried.** The host supplies `pins.use`; this one
  pushes the pair through `pushAssign()`/`map.set()`, the path the map's own
  edits take, so the save bar comes up and nothing reaches majestic until Save.
  That is what lets the filter test go on refusing to judge wiring the camera was
  never given. Without `pins.use` the old hand-over runs, which is the Pins route.
- **No resting card.** The door is the panel's own button, so an idle panel costs
  the page no height and Cancel closes the card rather than replacing it with a
  standing offer to do the same thing again.
- **The button says why it cannot.** `cannotHunt()` reaches the panel through
  `mount()`'s `blocked()`, disabling the button and printing the reason under it
  — reviving `#mj-ircut-find-why`, markup written and read by nothing since the
  hunt left. Where the kernel will not say who holds which pad, that sentence now
  points at the map above rather than at another page.
- **The two buttons stop matching.** One starts a search for wiring nobody knows,
  the other measures wiring already set; while both were the same quiet outline,
  their words were the only thing telling them apart.

## A hazard closed on the way past

Pre-existing, and worse with two doors. `stopped` is one module-global flag, and
teardown-then-mount happens in the same task when a reader leaves one section for
another — so a sweep told to stop found the flag false again on its next pad and
carried on driving hardware with its Stop button already thrown away. A run now
holds the generation it started under and stops when that is superseded, which no
later mount can undo. Checked that this cannot stop a sweep spuriously: the Pins
page's live pin levels go through `applyLevels()`/`applyHunt()` and never
`paint()`, so nothing repaints that section on a timer.

## Measured

On an hi3516av300 with a filter on pads 5 and 6:

- One press, and the confirm card with its `Start` button is in view at 390×844,
  1280×720, 1280×900 and 2560×1440.
- The sweep finds it on the first pair — *"Pins 5 and 6 drive the filter — 6 is
  the one that closes it"* — with both pads lit on the map while it runs. Driving
  each pair by hand through `/api/v1/gpio` agrees: 6 against 5 gives a colour
  frame, 5 against 6 gives the IR look.
- **Use these** fills both fields, marks two rows dirty and raises Save, and
  `config.json` still carries neither key. Pressing Save posts exactly
  `{"nightMode":{"irCutPin1":"5","irCutPin2":"6"}}`.
- The filter test refuses until that Save: *"The wiring on the pin map has not
  been saved yet."*
- The Pins route is unchanged end to end — find there, hand over, land on
  Day / Night with both fields staged and the proposal consumed.
- An unfinished scan is answered on Day / Night too.
- With `ownersUnknown` injected, the button is disabled and the line reads *"Set
  the coils by hand on the map above instead."*
- `tools/check-layout.mjs`: self-test fired all five rules, then clean at 390,
  1280 and 2560 on **both** pages with the confirm card open.

No stylesheet regeneration — the classes are already in the purged sheet.


openipc-ai:
Both findings read, one taken and one declined.

## Taken — the module header (9fa8bb2)

Right, and worth more than the finding claims. `www/a/mj-settings.js` had **no**
header at all: it opened straight into its IIFE, alone among the modules in this
directory and the longest one there by some margin. This PR adds hunt lifetime
and result staging to it, so that was the moment to fix it.

It now says what the module is — majestic's schema rendered as a form, knowing
nothing about what any individual setting means — and three refusals that matter
to anyone editing it: nothing reaches majestic outside a Save, configuration is
only ever read from the daemon, and a panel that put a listener on `document` or
left hardware moving is torn down when the section goes rather than when a
replacement happens to mount. It also states the panel contract, since that is
the thing a reader is most likely to break, and points at `docs/settings-page.md`
and `docs/day-night.md` for the reasoning that spans more files than this one.

## Declined — the `#552` citations

This is the tree's documented convention, not a lapse from it. `CLAUDE.md`, under
*Repo etiquette*:

> **A comment explains the code, not the review that produced it.** Cite the
> **issue** that reported the fault — `(#273)` in this tree, or an issue number in
> a public OpenIPC tracker that actually resolves — or cite nothing and let the
> sentence stand. Never a PR number, never a review tool, never a comment
> addressed to a reviewer.

`#552` is an issue in this repository that resolves, and it is the issue that
reported this fault. The prohibition is on PR numbers and review tooling, neither
of which appears here. `www/a/*.js` already carries 122 citations of exactly this
form — `(#317)` 27 times, `(#325)` 14, `(#273)` 11 — and they are load-bearing:
`(#325)` is why `SECTION_GROUPS` exists at all, and a reader who deletes the
grouping without reading that issue reinvents the nineteen-control deal it was
written to stop.

The finding's own reasoning is the case *for* keeping them. It says the
surrounding prose already preserves the rationale — it does, which is why the tag
is a pointer to the measurement and the reporter behind the sentence rather than a
substitute for it. Removing three of 122 instances would leave the convention
inconsistent and lose the only link from the code to the report that produced it.

Also noting this was filed as a 🐞 Bug. It changes no behaviour.

Issue #551, opened 2026-09-20
Day / Night: two banners send you to a switch that no longer exists


openipc-ai:
Found while checking the wiki against a current camera, not by a user report.

`www/a/ircut-check.js` has three sentences that name a control called **"Drive
the IR-cut filter"** or **"Drive the camera light"** and tell the reader to go
and turn it back on:

- `www/a/ircut-check.js:264` — *"nothing moves the filter until \"Drive the IR-cut filter\" is turned back on"*
- `www/a/ircut-check.js:300` — *"If this camera has no IR-cut filter fitted, turn off \"Drive the IR-cut filter\" and nothing here will ask again."*
- `www/a/ircut-check.js:543` — *"night mode leaves it dark until \"Drive the camera light\" is turned back on."*

Neither switch exists any more. Each actuator carries one three-state mode now —
the rows labelled **IR-cut filter** and **Camera light**, reading *off*, *manual*
or *auto* — which is what #512 followed the daemon into. Checked against a
current camera's schema: no setting carries a title beginning "Drive".

So the reader is told to find a control that is not on the page, in the one
place where the whole point of the sentence is to hand them the fix. The third
one is the worst of the three: it is the Dashboard's answer to "this camera has
no IR-cut filter fitted", and it is the sentence that exists so nobody has to
dismiss the banner forever.

The fix is the words, not the logic — each sentence should name the row and the
value, e.g. set **IR-cut filter** to **off**. While in there, `www/cgi-bin/
dashboard.cgi:59` and `www/a/mj-tree.js:234` quote the old label too; both are
comments, so they are explanation rather than instruction, but they read as
though the switch is still there.

PR #558, opened 2026-09-20
tests: a sentence may not name a control the camera does not have

### LINKED ISSUE #556: Nothing checks that a control name quoted in prose still exists
Follow-up to #551, which was one instance of this.

## The defect class

User-facing prose quotes schema **titles** — `"IR-cut filter"`, `"Minimum
sensor threshold for day"` — because CLAUDE.md requires it: *"Never name a
config key in text a person reads. Use the words the page puts on screen."*
That is the right rule. But a title is the daemon's to change, and nothing in
this repository checks that a title a sentence quotes still exists.

When #512 followed the daemon's three-state actuator modes, two titles went
away. Three sentences went on naming them for four days, on the page whose
entire job is to hand the reader the fix — and one of them reached the
Dashboard banner beside a button that opened the very page where the control
no longer was.

**The sharpest part is that a test made it worse, not better.** The only
assertion covering any of those sentences pinned the *old* wording, in its
body and in its own name (`'and that finding names the switch that answers
it'`). So the suite did not fail to notice the rot — it defended it. The two
sentences nobody had pinned are the two that were free to rot quietly, and the
one that was pinned had to be rewritten before the fix could land.

## How big it is

Measured over `www/a/*.js` on `master`, excluding comments:

- **34** distinct capitalised strings appear quoted inside a sentence.
- **7** of them match a title in `tests/fixtures/schema-hisi.json`. These are
  the ones that can rot: `"IR-cut filter"`, `"Camera light"`, `"Single IRcut
  is inverted"`, `"Minimum sensor threshold for day"`, `"Maximum sensor
  threshold for night"`, `"Gain multiple that means night"`, `"Seconds of
  brightness before day"` — in `ircut-check.js` and `mj-settings.js`.
- **27** do not, and nearly all are trivially not labels: SVG path data
  (`"M10 4.5v11M4.5 10h11"`), and short button/ARIA strings (`"Close"`,
  `"Dismiss"`, `"Next day"`). A path-data pattern and a length floor exclude
  them.

So the guard is small: a handful of real labels, and an exclusion rule that is
mechanical rather than a hand-kept list of exceptions.

## The obstacle, which has to be solved first

`tests/fixtures/schema-hisi.json` carries **203** titles. A current camera's
`/api/v1/config.schema.json` carries **326**. The fixture is about 62% of a
real camera, and the gap is not exotic keys:

> `rc-check.js` quotes `"Most compression allowed (QP)"`. It is **correct** —
> `video0.maxQp` carries exactly that title on a camera — but the fixture has
> no QP key at all, so a guard keyed to the fixture would fail a good label.

That is this tree's most-repeated rule pointed at a test: *an absent reading is
not a zero, and a failed fetch is not a fact*. "Not in the fixture" means **could
not check**, not **wrong**, and a guard that conflates them produces a confident
failure about a label nobody has looked at. But degrading that to silence gives
back exactly the case that caused #551, since a removed title also matches
nothing.

**So the fixture has to become complete before the guard can be strict.**
Regenerate it from a current camera as its own step — it is worth doing anyway,
since `tree.test.js` and `boot-blob.test.js` read it too and are equally blind
to the 123 keys it does not carry.

## Proposed shape

1. Regenerate `tests/fixtures/schema-hisi.json` from a current camera, and note
   in it which build it came from.
2. Add a test that walks the user-facing modules, extracts every quoted
   capitalised string that is not SVG path data, and asserts each one is a title
   in the fixture.
3. Keep the exclusion mechanical. If it ever needs a per-string allow-list, that
   is a sign the sentence should not be quoting that string.

It would have failed on `"Drive the IR-cut filter"` the day #512 landed.

openipc-ai:
Closes #556.

The page explains a problem by naming the row that fixes it, because CLAUDE.md
requires exactly that — *"Never name a config key in text a person reads. Use the
words the page puts on screen."* So the prose quotes schema **titles**, and a
title belongs to the daemon, which is free to rename or remove one. Nothing
checked that it had not.

When it did, three sentences went on naming switches that were no longer on the
page for four days (#551) — one on the Dashboard, beside a button opening the
very page the control had left. **And the suite did not merely miss it: the one
test that touched those sentences pinned the dead wording**, in its assertion and
in its own name, so it had to be rewritten before the fix could land. A test can
defend a defect as easily as catch one, and the two sentences nobody had pinned
are the two that rotted quietly.

## The rule is mechanical, which is the only reason this is worth having

The house convention is a title in straight double quotes inside a sentence.
Everything else carrying double quotes in these files is concatenated markup, and
markup is distinguishable without judgement: **an attribute value is preceded by
`=`**. That one test drops every `aria-label`, `title=`, `class=` and SVG `d=` in
the tree — 40 occurrences against the 15 that are prose. The rest are as blunt:
capitalised, contains a space, not an ALL-CAPS nodeName. Measured over `www/a/`,
that leaves exactly the 8 real labels and nothing else.

**No per-string allow-list, deliberately.** A list of strings to ignore is a place
for a rotted label to hide, and needing one would mean a sentence is quoting
something that is not a control — which is its own defect.

## A fixture of its own, and not the regeneration the report proposed

I went to regenerate `schema-hisi.json` as #556 suggested and found it is
**curated**: it holds one specific bitrate flag for `tree.test.js`, and a hint in
it was reworded so it names no build option. Rewriting it to widen coverage here
would put both at risk for a reason that has nothing to do with the tree.
`schema-titles.json` is titles alone.

It is the **union** of a current HiSilicon camera's titles and the curated
fixture's, because neither is a superset — the camera carries 66 the fixture
lacks, the fixture 32 the camera lacks, being other SoCs and other daemons. A
title in neither is one no schema on file has ever had. That union is what makes
the QP ceiling `rc-check.js` quotes pass: it is correct on hardware and absent
from the curated fixture, so a check keyed to that fixture alone would have failed
a good label.

**The limit is said out loud in the file.** A label real on hardware nobody here
has would fail this. That is the known-absent case, and the answer is to add the
title — one reviewable line recording which camera said so — rather than to teach
the check to look away from a string.

## Mutation-checked both ways

A new guard that passes immediately has proved nothing.

| mutation | result |
|---|---|
| put `"Drive the IR-cut filter"` back | **FAIL**, naming `ircut-check.js:265` and the string — the #551 defect, caught |
| break the extractor so it matches nothing | **FAIL** on the no-op guard (`found 0`) and on the extractor's own unit checks |

The extractor is a function rather than a regex inline in the assertion, so the
rule itself is tested: six checks cover a label in a sentence, an attribute value,
SVG path data, an ALL-CAPS nodeName, a single word and a comment line.


openipc-ai:
All three taken, in `d8c4ff6`. Two of them were holes in the check itself rather
than in its wording, which is the worst kind of finding to get right.

## Split control names can rot unnoticed — the important one

Correct, and worse than the finding states. This tree wraps prose at 72 columns
everywhere, so a quoted label straddles the wrap often. Scanning physical lines
did not report a wrong answer; it **quietly declined to look**, at exactly the
longest sentences — the ones most likely to be naming something.

Measured: **3 of 15 occurrences hidden, and one label outright** —
`"Seconds of darkness before night"`, which no version of this check had ever
seen. The literals are joined back before scanning now, keeping the line the
sentence *starts* on so a failure still points somewhere.

Mutation-checked on the case only the new version can reach — rotting the half of
that label which lives on the second line now fails with
`ircut-check.js:738 "Seconds of dusk before night"`.

## Long control names escape validation

Correct. One shipped title runs to 109 characters, so the 70-character bound
excluded it outright. The bound is gone: the other filters — attribute value,
capitalised, contains a space, not ALL-CAPS — carry the whole rule, and none of
them is a size. A cap is just a second way to decline to look.

## A test header cites an AI agent

Correct on this repository's own terms. `pr_compliance_checklist.yaml` forbids a
comment naming an AI agent and exempts exactly the three files that *state* that
rule; a test is not one of them, whatever the filename being pointed at.

The sentence never needed the citation. The fact is that the page names controls
in the words on screen rather than by the dotted key underneath, and a reader of
this file can check that without being told where it is written down.

## Two unit checks added

A fix to an extractor that is not itself tested is just a different untested
extractor. So: a label split across a wrapped concatenation is returned whole and
reported against the line its sentence starts on, and a 90-character label is
found. The no-op floor rises from 7 to 15, which is what the extractor now
actually sees.

## One thing noted, not changed

Two pre-existing comments name the convention file the same way — in
`www/a/pin-hunt.js` and twice in `tools/lint-templates.sh`. They are outside this
diff so I have left them; happy to file that separately if it is worth a pass.

PR #559, opened 2026-09-20
Network: the values an operator types are arguments, not shell

### LINKED ISSUE #547: Network settings are not shell escaped
At https://github.com/OpenIPC/majestic-webui/blob/master/www/cgi-bin/network.cgi#L494, the network configuration parameters are passed to setnetwork via the shell without escaping. This prevents saving the network configuration when any part of it (for example, the wifi password) contains special characters interpreted by the shell.

I recommend looking through the source code and making sure this doesn't happen anywhere else, too. I don't think it's a security issue for your threat model (since if you're logged in, the web ui will already give you a root shell), but it will definitely cause errors or certain characters to be lost if the user input contains special characters.

openipc-ai:
Fixes #547.

@danielrparks is right, and the audit he asked for turned up three more places
with the same fault — one of which the reported bug was hiding.

## The reported bug

`network.cgi` pasted every field into one string and `eval`'d it, so the shell
parsed what the operator typed. `setnetwork` was never at fault: it reads
`getopts`, so it has always taken its values as separate arguments. It just had
no way to receive them separately while the caller was building one string.

Measured on an hi3516av300, posting to the hostname field (it is in the command
on every save):

| typed | before | after |
|---|---|---|
| `my camera` | hostname became `my`, page said **"Network settings saved."** | saved whole |
| `cam"quote` | refused, with a raw `sh: syntax error: unterminated quoted string` on screen | saved whole |
| `cam'apos` | same | saved whole |
| `cam;touch /tmp/x` | the file was created | saved as a literal name |

The threat-model note in the issue is right — `j/run.cgi` already gives an
authenticated user a root console — so this is correctness, not privilege
escalation.

## Why the rest had to come with it

`update_caminfo` wrote the sysinfo cache as `name='value'` with nothing
escaped, and `common.cgi` sources that file before it can render anything. A
camera named with an apostrophe made it unparseable and **every page answered
500** — dashboard, live, network, camera and logs all measured — recoverable
only over SSH.

Today the network page refuses such a name *by accident*, its own quoting being
broken in the way above. Fixing #547 alone would have converted "a save that
fails" into "a network page that can take the WebUI down", so the two are in
one commit.

## The other two families

The five extension pages wrote their config with `eval echo $value` inside
double quotes. The read globbed — a caption of `snap *` was stored as a listing
of `/var/www/cgi-bin` — and the double quotes meant the file, which is sourced
by the page, by `clip_hook_wanted`, by `record.sh` and by the senders, had a
caption of `Motion at the "front door"` truncate and leave `door` to be run as
a command, while a `$(...)` in one ran on every page load.

The three senders built curl command lines the same way. `telegram` and
`ntfy.sh` escaped the snapshot path and message by hand and left the token,
chat, thread and proxy credentials bare — `-u ${ntfy_user}:${ntfy_pass}` among
them. `openwall` escaped nothing.

Separately, `field_text`, `field_password` and `field_hidden` put their value
into `value="…"` with no `attr_escape`, though the comment above `field_switch`
explains exactly why that matters: a camera named `cam"q onerror=x` rendered an
input whose attribute closed early and gained an `onerror`.

## The fix

Everything that built a command string now accumulates argv with `set --` and
runs `"$@"` — which is what `sbin/max` has always done, and it is the model.
Everything written into a file that is later sourced goes through a new `shq()`
in `common.cgi`. The three field helpers use `attr_escape`.

`shq` uses `sed` rather than `${s//x/y}` because that is not POSIX: dash parses
it and fails at run time, so `sh -n` cannot tell the difference, and the one
helper whose whole job is quoting is the last that should depend on it.

Also: `/tmp/webui.log` no longer records the WiFi passphrase, which it did on
every save. Both notification senders had already dropped their command-line
echo for the same reason.

## Verified, on an hi3516av300 + imx415

- All four reproductions above now save whole; the apostrophe case leaves every
  page at 200.
- Captions containing quotes, globs, apostrophes and `$(...)` round-trip byte
  for byte, and sourcing the config executes nothing.
- All three senders build exactly one argument per value, checked with the
  execution stubbed so nothing was uploaded anywhere.
- The camera was returned to the state it was found in.

## Keeping it out

`tools/lint-templates.sh` gains a section banning all three shapes; the
legitimate count of each is now zero. It does **not** ban `eval` — an
assignment does not split its right-hand side, a lookup by a fixed name is a
lookup, and `ex()` and `j/run.cgi` exist to run a command. Each pattern is
anchored past comments, because the first draft of that section was failed by
its own explanation of itself.

`tests/shell-quoting.test.js` drives `shq` and `t_value` out of the shipped
`common.cgi` with a real `sh` and round-trips seventeen awkward values through
an actual sourced file. Both halves were mutation-checked: dropping the
apostrophe escaping and restoring `eval echo` each fail it.


openipc-ai:
Both findings addressed in 5d61c5d.

**2. Backslash values fail every build** — correct, and CI caught it at the
same time. `echo` under dash acts on the backslashes in its argument, so `\c`
ended the output and the file was written as `caption='C:`. It is `printf`
everywhere now. Worth recording that the camera's busybox `echo` leaves
backslashes alone, so this was latent on a camera rather than live — but a
writer that depends on which `echo` it got is a writer waiting to lose a
value.

**1. Config values lose trailing newlines** — correct, on two of the three
paths named. The capture *inside* `shq` stripped them, and so did
`$(t_value ...)` in each config writer. `shq` now streams its opening quote,
the escaped value and its closing quote, and a new `conf_write` composes the
whole assignment line with no capture anywhere on the path; the six writers
take their value with a plain assignment rather than `$(t_value ...)`, since
an assignment's right-hand side is neither word-split nor newline-trimmed.

One correction to the finding: capturing `shq` *at the call site* —
`"name=$(shq "$v")"` — was never lossy, because `shq`'s output ends in a
single quote, so there are no trailing newlines for the substitution to
strip. Confirmed by mutation: restoring that spelling passes the new
trailing-newline cases, while restoring the capture inside `shq` fails them.
So capturing `shq` is still fine where a string is being assembled rather
than a file written, which is what the network page's log line does.

`update_caminfo` also went through `conf_write`. Its values are already
`$(...)` captures of command output by the time `shq` sees them, so there was
no trailing newline left to lose there — but it is the same writer and it
should not be the one place with its own spelling.

The test now writes its line with the shipped `conf_write` rather than a
spelling invented in the test, since both faults were in the writer and not
in `shq`, and it covers a trailing newline and a run of them. Its structural
"am I testing anything" assertion on `shq`'s body is gone: it fired on every
rewrite, including the two mutants that would have named these faults, before
the behavioural cases could run. Five mutants now fail it — `shq` capturing
its own output (2 cases), `shq` without the apostrophe escape (1), `shq`
without quoting (17), `t_value` back to `eval echo` (6), and `conf_write`
using `echo` (1).

Re-verified on the hi3516av300 + imx415 that produced the original
reproductions.

PR #568, opened 2026-09-22
SD card: say how much this camera has written to this card

widgetii:
The sentence the SD card page needs once the recorder publishes a durable
per-card write total. The daemon side is tracked separately; this is the page.

A fourth line on the Card health block, under the three findings:

> **Written by this camera** — This camera has written 3.4 TB to this card since
> 2 November.

It reads `records_card_bytes_written_total` and
`records_card_start_time_seconds` — a total the recorder keeps in a file on the
card itself, so it survives a restart and travels with the card. That matters
because the cards these cameras run have no SMART, and this is the nearest thing
to Total-LBAs-Written such a card can offer.

## Not a finding, and deliberately not in `head()`

The same rule `provenance` already follows. A card that has taken 40 TB is not
thereby a card with a problem, and there is no rated endurance in any register
on an SD card to turn the number into a percentage or a remaining life. The
tests assert the absence of all of that:

```js
check('no percentage, no remaining life, no "healthy"',
    !/%|remaining|life|wear|healthy/i.test(tb.text), tb.text);
check('and the headline is still decided by the three findings alone',
    v.head.level === 'ok', …);
```

## Two limits, in the wording rather than beside it

A reader who does not know them will draw the wrong conclusion from the number.
"**This camera** has written", because a card arriving with a year of somebody
else's footage starts from zero here. And the figure is host bytes — a lower
bound on the wear the card's flash actually took, since the translation layer
writes more than it is handed.

## Three absences that are not zeroes

- A build whose recorder does not publish the total gets **no line at all**
  rather than an empty one. It is an extra fact, and its absence says nothing
  about the card the reader needs to hear.
- A start time of `0` is the camera saying it has never had a clock it could
  believe — which most of these boards have not at boot, having no
  battery-backed RTC. The total is still shown and the date is simply left off:
  "since 1 January 1970" would be the page inventing one.
- **A counter that is not a reading is not a reading.** `num()` returns whatever
  the scrape parsed, so `Infinity`, `NaN` and negatives reach here. `bytes()`
  answers `''` for exactly those, and `''` interpolated into the sentence reads
  as "This camera has written  to this card" — a confident claim with the number
  missing out of the middle of it. A nonfinite start time renders the words
  "Invalid Date" the same way. Both are now refused at the top of `written()`,
  with the total still said when only the date is unusable.

## `bytes()` grows a TB step

It needed one anyway. A camera writing 40 GB a day passes a terabyte in under a
month, and `14600 GB` is not a number anybody reads.

## Tested

`npm test` green, including 22 checks in `tests/sdcard-health.test.js` covering
the wording, the three absences and the malformed-input cases. `npm run
build:dist`, `./tools/lint-templates.sh` and `./tools/regen-bootstrap-css.sh`
all clean — the last one matters here because purgecss extracts class names from
comments in this tree and this change adds several; the regenerated
`bootstrap.min.css` is byte-identical to the committed one.

Also exercised against a camera's live `/metrics`: an hi3516av300 recording
continuously to a 32 GB FAT32 card, reading back the two counters and rendering
the line from them.

Safe to merge in either order — on a camera whose recorder does not publish the
metrics the block renders exactly as it does today — but it says nothing useful
until a build carrying them exists.


widgetii:
All three were right, and two of them were mine in the description rather than
in the code.

**1. Private tracker details become public — fixed.** The description named the
sibling repository and a tracker number in it. Rewritten to refer to the daemon
side without naming either. This one cannot be undone after merge, so thank you
for catching it before.

**2. Lab camera output becomes public — fixed.** The pasted rendered output was
captured from an identified lab camera. Removed; the verification is now
described by hardware class and configuration, which is what the gate asks for.

**3. Malformed metrics render empty claims — fixed (425345f).** Correct, and it
is the exact failure this module exists to prevent. `num()` hands back whatever
the scrape parsed, so `Infinity`, `NaN` and negatives reach `written()`;
`bytes()` answers `''` for those, and `''` interpolated into the sentence reads
"This camera has written  to this card" — a confident claim with the number
missing out of the middle of it. A nonfinite start time puts "Invalid Date"
there the same way.

Both are refused at the top of `written()` now, and the two cases are kept
apart deliberately: an unusable total means no line at all, while an unusable
date costs only the date — the total is still worth saying, and dropping it
would lose a good reading over a bad one beside it.

Ten new checks cover it: `Infinity`, `-Infinity`, `NaN` and a negative total
each produce no line, and a nonfinite or negative start time leaves the
sentence with its total and no date and no "Invalid".

